PR #10, opened 2019-07-09
Adjusted example application

sergii-vostrikov:
The example now uses test catalog and downloads the first metadata field
and first data from it

Resolves: OLPEDGE-298


examples/dataservice-read/example.cpp
```diff
@@ -32,29 +32,101 @@
...
+const std::string kKeyId("");      // your here.access.key.id
+const std::string kKeySecret("");  // your here.access.key.secret
+const std::string kCatalogHRN("hrn:here:data:::edge-example-catalog");
+constexpr size_t kMaxLayers(5);
+constexpr size_t kMaxPartitions(5);
+constexpr auto kLogTag = "read-example";
+
+// handles catalog response and returns a partition ID from it to be processed
+// in later requests
+std::string handleCatalogResponse(
```
andescu: Method names should be CamelCase not camelCase
See https://google.github.io/styleguide/cppguide.html#Function_Names
Apply please for all below.

PR #15, opened 2019-07-10
Travis CI setup without docker

olrudenk:
Travis CI setup without docker
Using xenial dist
Resolves: OLPEDGE-420

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,18 @@ language: cpp
 matrix:
   include:
   - os: linux
-    services: docker
+    env:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,10 +2,34 @@ language: cpp
...
+        - libsodium-dev
+        - libsqlite3-dev
+        - libssl-dev
+        - libsystemd-dev
+        - pkg-config
+        - python3-dev
+        - python3-openssl
+        - python3-venv
+        - sqlite3
+        - valgrind
```
andescu: why do we need Valgrind

PR #25, opened 2019-07-16
Add build verification for Windows on Azure

olrudenk:
Create inital Azure config file
Used only pull request for Azure pipelines trigger
Some syntax simplified in windows build script
Resolves: OLPEDGE-531


olrudenk:
> Should the yml file not be hidden, means .azure-pipelines.yml?

No, it creates by default azure-pipelines.yml file

PR #28, opened 2019-07-17
ApiCacheRepository now logs all its major steps.

sergii-vostrikov:
Resolves: OLPEDGE-528

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/ApiCacheRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/ApiCacheRepository.cpp
@@ -20,8 +20,11 @@
 #include "ApiCacheRepository.h"
 
 #include <olp/core/cache/KeyValueCache.h>
+#include <olp/core/logging/Log.h>
 
 namespace {
+constexpr auto kLogTag = "ApiCacheRepository";
+
 std::string CreateKey(const std::string& hrn, const std::string& service,
                       const std::string& serviceVersion) {
   return hrn + "::" + service + "::" + serviceVersion + "::api";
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/ApiCacheRepository.cpp
```diff
@@ -20,8 +20,11 @@
 #include "ApiCacheRepository.h"
 
 #include <olp/core/cache/KeyValueCache.h>
+#include <olp/core/logging/Log.h>
 
 namespace {
+constexpr auto kLogtag = "ApiCacheRepository";
```
andescu:
```suggestion
constexpr auto kLogTag = "ApiCacheRepository";
```

olp-cpp-sdk-dataservice-read/src/repositories/ApiCacheRepository.cpp
```diff
@@ -41,15 +44,18 @@ void ApiCacheRepository::Put(const std::string& service,
                              const std::string& serviceVersion,
                              const std::string& serviceUrl) {
   std::string hrn(hrn_.ToCatalogHRNString());
+  auto key = CreateKey(hrn, service, serviceVersion);
+  LOG_TRACE_F(kLogtag, "Put '%s'", key.c_str());
```
andescu: Why trace and not info?
Same below.

sergii-vostrikov: as per decision to put methods entry at trace level

andescu: Methods entry yes but you are printing parameters.
Method entry trace log should be without parameters print:
E.g.
```
LOG_TRACE_F(kLogTag, "%s", __PRETTY_FUNCTION__);
```

PR #37, opened 2019-07-18
Applies pimpl idiom with shared_ptr to CancellationContext.

andescu:
Currently to be able to share the CancellationContext accross multiple contexts
you need to wrap it inside a shared_ptr. As this is designed to be used accross
multiple contexts only, applying a pimpl idiom with shared_ptr automatically
takes care of sharing the CancellationContext.

Relates-to: OLPEDGE-402

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
@@ -20,6 +20,7 @@
 #pragma once
 
 #include <functional>
+#include <memory>
 #include <mutex>
 #include "CancellationToken.h"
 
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
```diff
@@ -29,58 +30,62 @@ namespace olp {
...
-  /**
-  * @brief Execute a given cancellable code block if the operation has not yet
-  been cancelled. Otherwise, execute a custom cancellation function.
-  * @param execute_fn A function to execute if this operation is not yet
-  cancelled. This function should return a CancellationToken which
-  CancellationContext will propogate a cancel request to.
-  * @param cancel_fn A function which will be called if this operation is
-  already cancelled.
-  */
-  inline void ExecuteOrCancelled(
```
leonid-ziskel: Maybe ExecuteIfNotCancelled ?

olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
```diff
@@ -29,58 +30,62 @@ namespace olp {
 namespace client {
 
 /**
- * @brief Wrapper which manages cancellation state for multiple network
+ * @brief Wrapper which manages cancellation state for any asynchronous
  * operations in a thread safe way.
```
borzun: >  thread safe way.

Because _CancellationContext_ is movable, this means that if thread#1 moves instance of _CancellationContext_ while other thread#2 executing some method, like _ExecuteOrCancelled_, this will lead to the data race (the _impl__ will be null, thus potential crash in thread#2). 

Thus, I guess, it is better instead of accessing _impl__:
```
if (!impl) {
    return;
}
```
use copy of this pointer:
```
auto impl = impl_;
if (!impl) {
    return;
}
```
Correct me If I'm wrong...

olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
```diff
@@ -29,58 +30,62 @@ namespace olp {
...
+   * @param cancel_fn A function which will be called if this operation is
+   * already cancelled.
+   * @deprecated Will be removed once TaskScheduler will be used.
    */
-  inline void CancelOperation() {
-    std::lock_guard<std::recursive_mutex> lock(mutex_);
-    if (is_cancelled_) {
-      return;
-    }
+  void ExecuteOrCancelled(const ExecuteFuncType& execute_fn,
```
borzun: Add also basic tests for _ExecuteOrCancelled_ - to cover those cases:

1. When _ExecuteOrCancelled_ is called, to check that _execute_fn_ is is also invoked;
2. When _CancellationContext_ was cancelled and _ExecuteOrCancelled_ is called, check that _cancel_fn_ is called and _execute_fn_ is not.

Or we do not need to cover @deprecated methods in tests?

olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
```diff
@@ -29,58 +30,62 @@ namespace olp {
...
+   * @return `true` on context cancelled, `false` otherwise.
+   */
+  bool IsCancelled() const;
 
  private:
-  std::recursive_mutex mutex_;
-  olp::client::CancellationToken sub_operation_cancel_token_{};
-  bool is_cancelled_{false};
+  /**
+   * @brief Implementation to be apple to shared a instance of
```
borzun:
```suggestion
   * @brief Implementation to be applied to a shared instance of
```
But not sure about article :)

olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
andescu: This is a shared pimpl, using shared_ptr. Means both thread1 and thread2 should only own a shared_ptr copy of the object and not a reference or anything. If thread1 owns a copy and moves it to the next level it only moves the shared_ptr copy so the impl_ should still be valid for all. If you move to thread1 and thread2 the same instance then there is no safeguard anyway for this and you should change your code to copy capture the context.

olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
andescu: As discussed, I just marked the method as deprecated and it will be removed after TaskScheduler is in use.

olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
andescu: Well, I hope I can remove it soon.
I don't think we need to explicitly add tests for deprecated methods.

olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
andescu: What I mean was `able`.
Thanks for the hint.

olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
borzun: > Means both thread1 and thread2 should only own a shared_ptr copy of the object and not a reference or anything.

Yes, I meant the case where there is a global _CancellationContext_ instance, which can be manipulated by 2 threads simultaneously (and one of them is performing move operation). Ofc, this case is most likely will not happen in the real scenario, but, if we state that public API is thread-safe we should comply with this statement.  

And, currently, this class has thread-safe copy-ctor and copy-assignment operator. Then why not to have thread-safe move-ctor & move-assignment operator?

andescu: But this means I need to protect the shared_ptr itself.
This is not a likely use case at all as the user only gets a copy as a return. What he does with his copy is his deal. Until now user gets this std::shared_ptr<CancellationContext> and can copy/move/crash just as easy as before.

borzun: It doesn't matter which way user will use CancellationContext - shared_ptr, raw pointer or reference. We just need to make sure that thread-safety is guaranteed regarding its contents.

PR #47, opened 2019-07-24
PrefetchTilesProvider logging enhancement

sergii-vostrikov:

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/MultiRequestContext.h
+++ b/olp-cpp-sdk-dataservice-read/src/MultiRequestContext.h
@@ -38,7 +38,7 @@ namespace dataservice {
 namespace read {
 
 namespace {
-constexpr auto kLogTag = "MultiRequestContext";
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-read/src/MultiRequestContext.h
```diff
@@ -38,7 +38,7 @@ namespace dataservice {
 namespace read {
 
 namespace {
-constexpr auto kLogTag = "MultiRequestContext";
+constexpr auto kMultiRequestContextLogTag = "MultiRequestContext";
```
serhiikobalt: in some places you rename kLogTag but not in the other (e.g. in PrefetchTilesProvider)

sergii-vostrikov: implementation files work perfectly in  separate files while MultiRequestContext is an in-header implementation which conflicts with other declarations of kLogTag

PR #48, opened 2019-07-25
Adds generic thread safe queue for both FIFO and LIFO usage.

andescu:
SyncQueue is a generic template thread safe queue that is able to
serve in both FIFO and LIFO order elements to concurrent threads.
Add unit tests for SyncQueue and runs them in the PSV test pipeline.

Relates-to: OLPEDGE-450


olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.h
```diff
@@ -0,0 +1,106 @@
...
+ * @tparam Container The type of the underlying container that should mimic the
+ * API of std::queue providing the following functions:
+ * - empty()
+ * - push(T&&)
+ * - push(const T&)
+ * - front()
+ * - pop()
+ */
+template <typename T, typename Container>
+class CORE_API SyncQueue final {
```
borzun: What about Front() method, which just retrieves top element without pulling it?  I guess this method will be helpful for the users of this class.

olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.h
```diff
@@ -0,0 +1,106 @@
...
+   * from pulling any elements from the queue.
+   */
+  void Close();
+
+  /**
+   * @brief Pulls one element from the SyncQueue.
+   * @param element The element pulled from the queue.
+   * @param closed \c true if this SyncQueue was closed, \c false otherwise.
+   */
+  void Pull(T& element, bool& closed);
```
borzun:
```suggestion
  bool Pull(T& element);
```
Why not return whether Pull method was successful? This will be more convenient, cause user don't have to provide additional variable (bool closed);

olp-cpp-sdk-core/tests/thread/unittest/SyncQueueTest.cpp
```diff
@@ -0,0 +1,207 @@
...
+  sync_queue.Pull(element, closed);
+  ASSERT_TRUE(element) << "Push() or Pull() failed.";
+  element();
+
+  EXPECT_TRUE(sync_queue.Empty());
+  EXPECT_FALSE(closed);
+  EXPECT_TRUE(check);
+}
+
+TEST(SyncQueueTest, push) {
```
borzun: Divide into 2 tests - for rvalues and laves

olp-cpp-sdk-core/tests/thread/unittest/SyncQueueTest.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    auto string = std::make_shared<std::string>("lvalue");
+    sync_queue.Push(string);
+
+    EXPECT_FALSE(sync_queue.Empty());
+    EXPECT_TRUE(string && !string->empty());
+  }
+}
+
+TEST(SyncQueueTest, pull) {
```
borzun: Divide into 2 test cases - for open queue and closed queue

olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.h
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * @brief Forwards the passed element into the SyncQueue.
+   * @param element The rvalue reference to the element.
+   */
+  void Push(T&& element);
```
borzun: Shouldn't pushing to queue new element fail when queue is closed?

olp-cpp-sdk-core/tests/thread/unittest/SyncQueueTest.cpp
andescu: No need. With SCOPED_TRACE you can easily see what failed and you don't need to carry around so many test cases. This is very helpfull in case you have unit/integration tests with complex and cost-intensive setup/teardown.

olp-cpp-sdk-core/tests/thread/unittest/SyncQueueTest.cpp
andescu: See above.

olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.h
andescu: Means two or more concurrent threads can access the same element at the same time? This is super tricky and dangerous. I would not provide this possibility.

olp-cpp-sdk-core/tests/thread/unittest/SyncQueueTest.cpp
borzun: Disagree:

- this test has 2 cases, which do not depend on each other; So, it is logical to divide them into separate test cases;
- what if I want to run only one case (consider example when one case fails and other not) - in your approach, still 2 cases will run;
- currently in SynQueueTest there are no complex setup / teardown;
- we should strive to minimize assertion per test - ideally it should be 1 assertion per test.

olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.h
andescu: But what does return true/false mean. You call Pull() and it returns false on success and true on a closed queue. This will be confusing in my opinion. That's why the variable is more self-explanatory and user-friendly.

olp-cpp-sdk-core/tests/thread/unittest/SyncQueueTest.cpp
andescu: - This tests one method, as this is a method oriented test setup so it is more logical to merge them together as then you have an overview of what was tested and whatnot and you can easily merge test scenarios in one.
- What is the problem on running 2 cases in one? In most scenarios you can see based on previous scenarios, in the same test, if the test approach was wrong or you changed something you should have not.
- But consistency is key in shared projects. Imagine every test has a different paradigm based on developer, type of class, API, etc. You need a lot of time to adapt.
- Why? I didn't see any rule for this and this will be very hard to accomplish, resp we will have endless execution time on CI if we test each member of the class for each scenario in one test case. This will explode.

olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.h
borzun: Ok, let's stick to the YAGNI principle.

olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.h
borzun: >  That's why the variable is more self-explanatory and user-friendly.

The current approach is not user-friendly as-well, cause user has to declare addition bool variable, even when he has no intention to use at all.

> But what does return true/false mean. 

Returning _false_ when the queue is closed - this means that the operation wasn't successful. True otherwise.

----------------------------
An alternate approach is to use a pointer instead of reference and default-initialize it to nullptr:
```suggestion
  void Pull(T& element, bool* closed = nullptr);
```
But the approach with returning true/false for me seems better.

olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.inl
```diff
@@ -0,0 +1,94 @@
...
+      return;
+    }
+
+    queue_.push(std::forward<T>(element));
+  }
+  ready_.notify_one();
+}
+
+template <typename T, typename Container>
+inline void SyncQueue<T, Container>::Push(const T& element) {
```
borzun: I guess we should also need to provide possibility to notify user that operation wasn't successful as it is done with `Pull` method, WDYT?

olp-cpp-sdk-core/tests/thread/unittest/SyncQueueTest.cpp
borzun: - if this is method oriented test setup, then we should have 2 tests per methods, cause there are 2 push methods with different parameters;
- this more like structural split of test case; the problem is overhead when you want to run just one test case - 2 cases will be checked instead single one;
- we are relatively new team and should agree on how our tests should looks like; It is ok to discuss them now;
- Many TDD folks uses this approach. But it is not a mandatory rule, but nice to have; 
Detailed discussion with pros and cons on SE: https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test

sergii-vostrikov: There should be different tests against different scenarios. This allows to isolate a particular case for whichever reason. The simplest benefit is test name notification.
If a scenario requires a complex things to run in a queue/parallel this also should be a separate test case.
So, I agree with @borzun, this case should be split.

olp-cpp-sdk-core/tests/thread/unittest/SyncQueueTest.cpp
```diff
@@ -0,0 +1,218 @@
...
+    SyncQueueShared sync_queue;
+
+    auto string = std::make_shared<std::string>("rvalue");
+    sync_queue.Push(std::move(string));
+
+    EXPECT_FALSE(sync_queue.Empty());
+    EXPECT_FALSE(string) << "string should be moved";
+  }
+  {
+    SCOPED_TRACE("Push supports rvalues");
```
serhiikobalt: Typo. "push supports lvalues".

olp-cpp-sdk-core/include/olp/core/thread/SyncQueue.inl
andescu: Currently there is no-one to listen to it if the operation failed.
I would consider it once it is needed.

PR #52, opened 2019-07-30
Add ThreadPoolTaskScheduler implementing TaskScheduler class.

andescu:
ThreadPoolTaskScheduler is an async task handler with user-defined amount of
concurrent threads listening to a SyncQueue that is fed by the user with
tasks to be executed trough the TaskScheduler API. Each thread started gets a distinct name, 
this way it is easier to profile or debug the thread pool in case of an error or in a complex system to track which component uses to much CPU time. Thread name setup is currently only supported for Linux, QNX and macOS.
Two basic unit tests scenarios with single-user push and multi-user push also added.

Relates-to: OLPEDGE-394

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/thread/ThreadPoolTaskScheduler.h
+++ b/olp-cpp-sdk-core/include/olp/core/thread/ThreadPoolTaskScheduler.h
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-core/include/olp/core/thread/ThreadPoolTaskScheduler.h
```diff
@@ -0,0 +1,58 @@
...
+#include <thread>
+#include <vector>
+
+#include <olp/core/thread/SyncQueue.h>
+#include <olp/core/thread/TaskScheduler.h>
+
+namespace olp {
+namespace thread {
+
+class CORE_API ThreadPoolTaskScheduler final : public TaskScheduler {
```
mykhailo-kuchma: What about copy/move operations? We should rather support them or disable them in TaskScheduler class.

andescu: As the TaskScheduler is an "abstract" interface there is not much need for deleting, I would leave to the implementation to decide if he supports copy & move operations.
For the ThreadPoolTaskScheduler it is implicitly deleted, both copy & move operations, as I have a user-declared destructor and SyncQueue is not copiable. But to be cleared i will add them.

PR #54, opened 2019-07-30
Configure Travis CI for code coverage

olrudenk:
This will setup Travis CI to run code coverage on gcc compiler job.
Add badges to README.md file.
Remove windows build from Travis CI because of Azure pipelines.
Resolves: OLPEDGE-238


README.md
```diff
@@ -1,5 +1,17 @@
-[![Build Status](https://travis-ci.com/heremaps/here-olp-edge-sdk-cpp.svg?branch=master)](https://travis-ci.com/heremaps/here-olp-edge-sdk-cpp)
-[![Build Status](https://dev.azure.com/heremaps/github/_apis/build/status/heremaps.here-olp-edge-sdk-cpp?branchName=master)](https://dev.azure.com/heremaps/github/_build/latest?definitionId=1&branchName=master)
+&nbsp;
+
+| Linux                          | Windows                         |
```
andescu: I would not put the table before the h1 name anymore.
I would move it if you need a table after "The HERE OLP Edge SDK is a work in progress ... " sentence.
Or remove the table and only leave badges as before until we figure out how to do it. I think we need an bigger table with all supported and verified platforms and then we can add badges there. Leo is preparing something like this.

PR #56, opened 2019-07-30
Logging prefix set to EDGE_SDK_

sergii-vostrikov:
All logging macros now include EDGE_SDK as a prefix

Resolves: OLPEDGE-407


sergii-vostrikov:
> This is a huge change. Please split into multiple small commits.
> 
> 1. Change Log.h macros, add #defines to keep old macros alive until all classes are adapted.
> 2. Adapt module based macros.
> 3. Remove old macros.

Your proposition would generate unnecessary many new pull-requests. The ticket stands for adding log prefix for all macros, without a 
The changes in the review if quite repetitive and easy to follow (despite many files affected).
I doubt if there any benefits of spending resources on generation additional 58 self-dependent pull-requests.

sergii-vostrikov:
I think we need to make a conversation on necessity of such splits.

sergii-vostrikov:
Discarded as per discussion above

PR #57, opened 2019-07-31
Internal Log.h macros renaming

sergii-vostrikov:
Internally-used macros inside Log.h file now have EDGE_SDK_ prefix

Relates-to: OLPEDGE-407

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/logging/Log.h
+++ b/olp-cpp-sdk-core/include/olp/core/logging/Log.h
@@ -54,11 +54,11 @@
  * compiler warnings.
  */
 #ifdef _MSC_VER
-#define CORE_LOOP_ONCE()                                                  \
+#define EDGE_SDK_CORE_LOOP_ONCE()                                         \
   __pragma(warning(push)) __pragma(warning(disable : 4127)) while (false) \
       __pragma(warning(pop))
 #else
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/include/olp/core/logging/Log.h
```diff
@@ -223,26 +225,26 @@
...
+    std::abort();                           \
+  }                                         \
+  EDGE_SDK_CORE_LOOP_ONCE()
 
 #ifdef LOGGING_DISABLED
-#define LOG(level, tag, message) \
-  do {                           \
-  }                              \
-  CORE_LOOP_ONCE()
+#define EDGE_SDK_LOG(level, tag, message) \
```
andescu: I wonder why we don't need to mark the parameters as unused same as for:
```cpp
#ifdef LOGGING_DISABLE_DEBUG_LEVEL
#define LOG_TRACE(tag, message)           \
  do {                                    \
    ::olp::logging::NullLogStream __strm; \
    __strm << message;                    \
    CORE_UNUSED(tag, __strm);             \
  }                                       \
  CORE_LOOP_ONCE()
```

sergii-vostrikov: Not a scope of the changes, but seems CORE_UNUSED is redundant in macro definition.

PR #61, opened 2019-08-01
Log.h macros migration preparation

sergii-vostrikov:
All macros in Log.h now have EDGE_SDK_ prefix.
To ease code review, all currently used macros from Log.h are mocked
with old names.
This allows further step-by-step migration to 'prefixed' logging.

Relates-to: OLPEDGE-407

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/logging/Log.h
+++ b/olp-cpp-sdk-core/include/olp/core/logging/Log.h
@@ -33,6 +33,18 @@
 #include <sstream>
 #include <string>
 
+#define LOG_FUNCTION_SIGNATURE EDGE_SDK_LOG_FUNCTION_SIGNATURE  /// @deprecated
+#define LOG_FATAL EDGE_SDK_LOG_FATAL                            /// @deprecated
+#define LOG_TRACE EDGE_SDK_LOG_TRACE                            /// @deprecated
+#define LOG_TRACE_F EDGE_SDK_LOG_TRACE_F                        /// @deprecated
+#define LOG_DEBUG EDGE_SDK_LOG_DEBUG                            /// @deprecated
+#define LOG_DEBUG_F EDGE_SDK_LOG_DEBUG_F                        /// @deprecated
+#define LOG_INFO EDGE_SDK_LOG_INFO                              /// @deprecated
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/include/olp/core/logging/Log.h
```diff
@@ -33,6 +33,21 @@
 #include <sstream>
 #include <string>
 
+/**
+ * @brief Temporary macros to ease EDGE_SDK_ prefix transition review.
+ */
+#define LOG_FUNCTION_SIGNATURE EDGE_SDK_LOG_FUNCTION_SIGNATURE
```
mykhailo-kuchma: Shouldn't this be at the end of file after EDGE_SDK_* are defined?

sergii-vostrikov: Does it matter? These lines to removed soon any way and at the beginning of the file it catches an eye better.

andescu: you need to move these down as last elements in the file else the macros will not expend correctly.

sergii-vostrikov: https://gcc.gnu.org/onlinedocs/cpp/Argument-Prescan.html#Argument-Prescan
> Macro arguments are completely macro-expanded before they are substituted into a macro body, unless they are stringized or pasted with other tokens. After substitution, the entire macro body, including the substituted arguments, is scanned again for macros to be expanded. The result is that the arguments are scanned twice to expand macro calls in them.

In short, the macros will be expanded correctly. I'd prefer to keep them at the top of the file to catch an eye better.

olp-cpp-sdk-core/include/olp/core/logging/Log.h
```diff
@@ -33,6 +33,21 @@
 #include <sstream>
 #include <string>
 
+/**
+ * @brief Temporary macros to ease EDGE_SDK_ prefix transition review.
```
andescu: "to ease ... review"?
Remove please, mark them as deprecated.

sergii-vostrikov: done

PR #63, opened 2019-08-01
Logging prefix for 'core' subproject

sergii-vostrikov:
All instances of loggging macros usage in olp-cpp-sdk-core now utilize
ones with EDGE_SDK prefix.

Relates-to: OLPEDGE-407


olp-cpp-sdk-core/src/network/Network.cpp
```diff
@@ -399,8 +399,8 @@ Network::RequestId Network::Send(NetworkRequest request,
                       char time_buffer[50];
                       std::strftime(time_buffer, sizeof(time_buffer), "%c %Z",
                                     request_time);
-                      //                LOG_ERROR( LOGTAG, "Request: " << url <<
```
andescu: Remove dead code

sergii-vostrikov: done

PR #74, opened 2019-08-05
Change return type of Send method in Network from olp::network2

serhiikobalt:
This is required to avoid calling callback in the same thread
where Send method was called

Related-To: OLPEDGE-369

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/network2/NetworkTypes.h
+++ b/olp-cpp-sdk-core/include/olp/core/network2/NetworkTypes.h
@@ -44,5 +44,58 @@ enum class RequestIdConstants : std::uint64_t {
   RequestIdMax = std::numeric_limits<RequestId>::max(),
 };
 
+/**
+ * @brief Common Network error codes.
+ */
+enum class ErrorCode {
+  NO_ERROR = 0,
+  IO_ERROR = -1,
+  AUTHORIZATION_ERROR = -2,
+  INVALID_URL_ERROR = -3,
+  OFFLINE_ERROR = -4,
+  CANCELLED_ERROR = -5,
+  AUTHENTICATION_ERROR = -6,
+  TIMEOUT_ERROR = -7,
+  UNKNOWN_ERROR = -8,
+};
+
+/**
+ * @brief Helper class representing the outcome of making a network request.
+ * It will contain either a valid RequestId or the error code in case request
+ * trigger failed.  The caller must check whether the outcome of the request was
+ * a success before attempting to access the result or the error.
+ */
+class NetworkStatus {
+ public:
+  NetworkStatus(RequestId id) : request_id_(id) {}
+  NetworkStatus(ErrorCode code) : error_code_(code) {}
+
+  /**
+   * @brief Check if network request push was successfull.
+   */
+  bool IsSuccessfull() const {
+    return error_code_ == ErrorCode::NO_ERROR &&
+           request_id_ !=
+               static_cast<RequestId>(RequestIdConstants::RequestIdInvalid);
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/include/olp/core/network2/NetworkTypes.h
```diff
@@ -44,5 +44,29 @@ enum class RequestIdConstants : std::uint64_t {
   RequestIdMax = std::numeric_limits<RequestId>::max(),
 };
 
+/**
+ * @brief Common Network error codes.
+ */
+enum class ErrorCode {
```
andescu: You need to give them negative numbers, else we are not able to merge them with the NetworkResponse::status_ which also includes the HTTP response codes.

Also i think the Network.h ErrorCodes are more appropriate and more user friendly.
If needed we can extend with additional error from NetworkProtocol.h ErrorCode.

olp-cpp-sdk-core/include/olp/core/network2/NetworkTypes.h
```diff
@@ -44,5 +44,48 @@ enum class RequestIdConstants : std::uint64_t {
...
+ */
+class NetworkStatus {
+ public:
+  NetworkStatus(RequestId id, ErrorCode code)
+      : request_id_(id), error_code_(code) {}
+
+  /**
+   * @brief Check if network request push was successfull.
+   */
+  bool IsSuccessfull() const { return error_code_ == ErrorCode::NO_ERROR; }
```
andescu:
```suggestion
  bool IsSuccessfull() const { return error_code_ == ErrorCode::NO_ERROR && request_id_ != static_cast<RequestId>(RequestIdInvalid); }
```

olp-cpp-sdk-core/include/olp/core/network2/NetworkTypes.h
```diff
@@ -44,5 +44,48 @@ enum class RequestIdConstants : std::uint64_t {
...
+  INVALID_URL_ERROR = -3,
+  OFFLINE_ERROR = -4,
+  CANCELLED_ERROR = -5,
+  AUTHENTICATION_ERROR = -6,
+  TIMEOUT_ERROR = -7,
+  UNKNOWN_ERROR = -8,
+};
+
+/**
+ * @brief Return value of Send method.
```
andescu:
```suggestion
 * @brief Helper class representing the outcome of making a network request.
 * It will contain either a valid RequestId or the error code in case request trigger failed.  The caller must check
 * whether the outcome of the request was a success before attempting to access the result or the error.
```

olp-cpp-sdk-core/include/olp/core/network2/NetworkTypes.h
```diff
@@ -44,5 +44,58 @@ enum class RequestIdConstants : std::uint64_t {
...
+  UNKNOWN_ERROR = -8,
+};
+
+/**
+ * @brief Helper class representing the outcome of making a network request.
+ * It will contain either a valid RequestId or the error code in case request
+ * trigger failed.  The caller must check whether the outcome of the request was
+ * a success before attempting to access the result or the error.
+ */
+class NetworkStatus {
```
borzun:
```suggestion
class SendRequestOutcome {
```
I guess NetworkStatus is a bit misleading name. For me it sounds like the status of network connection - for example, _offline_, _online_. I see that AWS using `Outcome` class as a analogue of our `ApiResponse` class:  https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/error-handling.html

So, I suggest similar name strategy - like "`NetworkRequestOutcome`" or "`SendRequestOutcome`", WDYT?

PR #82, opened 2019-08-12
Remove std::thread raw usage replacing it with TaskScheduler.

andescu:
Relates-to: OLPEDGE-456


andescu:
Not needed anymore as task handled by #84.
Closing PR.

PR #91, opened 2019-08-19
Adapt PendingRequest class for sync execution flow

mykhailo-kuchma:
Adapt PendingRequest class for a scenario when the key is removed before it is added.
This happens when the client is configured to use sync. task scheduler.

Resolves: OLPEDGE-638


olp-cpp-sdk-dataservice-read/test/unit/src/PendingRequestsTest.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+TEST(PendingRequests, CancellAllPendingRequest) {
+  olp::dataservice::read::PendingRequests pending_request;
+  auto key = pending_request.GenerateRequestPlaceholder();
+  bool cancelled = false;
+  auto token = olp::client::CancellationToken([&]() { cancelled = true; });
+  EXPECT_TRUE(pending_request.Insert(token, key));
+  EXPECT_TRUE(pending_request.CancelPendingRequests());
+  EXPECT_TRUE(cancelled);
+}
```
serhiikobalt: \n

PR #92, opened 2019-08-19
Add factory for default network implementation

serhiikobalt:
Provide a free function that returns unique_ptr to default network
implementation.

Related-To: OLPEDGE-641

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
@@ -26,6 +26,7 @@
 #include <boost/optional.hpp>
 
 #include "olp/core/client/CancellationToken.h"
+#include "olp/core/http/Network.h"
 #include "olp/core/network/NetworkProxy.h"
 
 namespace olp {
@@ -39,10 +40,6 @@ class HttpResponse;
 namespace thread {
 class TaskScheduler;
 }  // namespace thread
-
-namespace http {
-class Network;
-}  // namespace http
 
 namespace client {
 
--- a/olp-cpp-sdk-core/include/olp/core/http/Network.h
+++ b/olp-cpp-sdk-core/include/olp/core/http/Network.h
@@ -83,5 +83,10 @@ class CORE_API Network {
   virtual void Cancel(RequestId id) = 0;
 };
 
+/**
+ * @brief Create default Network implementation.
+ */
+CORE_API std::unique_ptr<Network> CreateDefaultNetwork(size_t max_requests_count);
+
 }  // namespace http
 }  // namespace olp
--- a/olp-cpp-sdk-core/src/http/Network.cpp
+++ b/olp-cpp-sdk-core/src/http/Network.cpp
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include "olp/core/http/Network.h"
+
+#include "olp/core/porting/make_unique.h"
+#include "olp/core/utils/WarningWorkarounds.h"
+
+#ifdef NETWORK_HAS_CURL
+#include "curl/NetworkCurl.h"
+#elif NETWORK_HAS_ANDROID
+#include "android/NetworkAndroid.h"
+#elif NETWORK_HAS_IOS
+#include "ios/OLPNetworkIOS.h"
+#elif NETWORK_HAS_WINHTTP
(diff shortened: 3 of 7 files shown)


olp-cpp-sdk-core/include/olp/core/http/Network.h
```diff
@@ -83,5 +83,10 @@ class CORE_API Network {
   virtual void Cancel(RequestId id) = 0;
 };
 
+/**
+ * @brief Get default Network implementation.
+ */
+CORE_API std::unique_ptr<Network> GetDefaultNetwork();
```
borzun:
```suggestion
CORE_API std::unique_ptr<Network> CreateDefaultNetwork();
```
I think the "CreateDefaultNetwork" better describe what this method does, i.e. creating a new instance of the network class instead retrieving some pre-created object.

olp-cpp-sdk-core/include/olp/core/http/Network.h
```diff
@@ -83,5 +83,10 @@ class CORE_API Network {
   virtual void Cancel(RequestId id) = 0;
 };
 
+/**
+ * @brief Create default Network implementation.
+ */
+CORE_API std::unique_ptr<Network> CreateDefaultNetwork();
```
andescu: I presume task description is old. This should be part of the OlpClientSettingsFactory class.

olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
```diff
@@ -186,7 +183,8 @@ struct OlpClientSettings {
    * @brief The network instance to be used to internally operate with OLP
    * services.
    */
-  std::shared_ptr<http::Network> network_request_handler = nullptr;
+  std::shared_ptr<http::Network> network_request_handler =
+      http::CreateDefaultNetwork();
```
andescu: This is not good. Means everybody creates a new Network instance before it can share and already existing one.
Means:

```cpp
// Create settings to pass to clients -> 1. Network instance created 
auto settings1 = std::make_shared<olp::client::OlpClientSettings>();
// Setup client settings
...
// Create second settings to pass to other client, BAD - 2. Network instance created
auto settings2 = std::make_shared<olp::client::OlpClientSettings>();
settings2->network_request_handler = settings1->network_request_handler;
```

Move back to nullptr initialization.

olp-cpp-sdk-core/include/olp/core/http/Network.h
andescu: What is the point of this still?
You can do it same as it is done now in NetworkProtocol GetProtocol():

```cpp
#ifdef NETWORK_HAS_CURL
  return std::make_shared<NetworkProtocolCurl>();
#endif
#ifdef NETWORK_HAS_ANDROID
  return std::make_shared<NetworkProtocolAndroid>();
#endif
#ifdef NETWORK_HAS_IOS
  return std::make_shared<NetworkProtocolIos>();
#endif
#ifdef NETWORK_HAS_WINHTTP
  return std::make_shared<NetworkProtocolWinHttp>();
#endif
  return nullptr;
```

serhiikobalt: @andescu 

If every network implementation has its own CreteDefaultNetwork() you can plug & unplug network implementations solely in CMakeLists and let the linker do the job for you. There will be less subjects to change and make mistake in if you ever need to touch network implementations.

andescu: Looks very confusing to me, having multiple declarations of the same method.
This also will make life very hard for those that want to implement own Network handler as they will also need to implement CreateDefaultNetwork() which will be very hard to achieve if you build olp sdk c++ as shared lib before.

olp-cpp-sdk-core/src/http/Network.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
```
andescu: Remove additional newline

olp-cpp-sdk-core/src/http/Network.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "curl/NetworkCurl.h"
+#elif NETWORK_HAS_ANDROID
+#include "android/NetworkAndroid.h"
+#elif NETWORK_HAS_IOS
+#include "ios/NetworkProtocolIos.h"
+#elif NETWORK_HAS_WINHTTP
+#include "winhttp/NetworkWinHttp.h"
+#endif
+
+
```
andescu: remove additional newline

olp-cpp-sdk-core/src/http/Network.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+
+#include "olp/core/http/Network.h"
+#include "olp/core/porting/make_unique.h"
```
andescu: newline between own header and project headers

olp-cpp-sdk-core/src/http/Network.cpp
```diff
@@ -0,0 +1,54 @@
...
+#elif NETWORK_HAS_WINHTTP
+#include "winhttp/NetworkWinHttp.h"
+#endif
+
+
+namespace olp {
+namespace http {
+
+CORE_API std::unique_ptr<Network> CreateDefaultNetwork() {
+
```
andescu: remove newline

PR #97, opened 2019-08-21
Add china URLs to read and write lookup API.

andescu:
To support china use cases two new URLs will be used for
here-cn and here-cn-dev partitions based on the provided HRN.
This also fixes a small bug in the logging formatter.

Relates-to: OLPEDGE-558


olp-cpp-sdk-core/src/logging/MessageFormatter.cpp
```diff
@@ -89,7 +89,7 @@ MessageFormatter::Element::Element(ElementType type_) : type(type_), limit(0) {
 
 const MessageFormatter::LevelNameMap& MessageFormatter::defaultLevelNameMap() {
   static MessageFormatter::LevelNameMap defaultLevelNameMap = {
-      {"[TRACE]", "[DEBUG]", "[INFO ]", "[WARN ]", "[ERROR]", "[FATAL]"}};
```
r0busta: Why it was fixed that way? @mykhailo-kuchma, maybe you could recall this?

PR #100, opened 2019-08-21
Added user-agent string in http network request

borzun:
Added OLP-CPP-SDK user agent string for all requests from iOS and Android network.

Relates to: OLPEDGE-552

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,9 @@ if (MSVC AND EDGE_SDK_MSVC_PARALLEL_BUILD_ENABLE)
     message(STATUS "MSVC - Using parallel compilation (/MP)")
 endif (MSVC AND EDGE_SDK_MSVC_PARALLEL_BUILD_ENABLE)
 
+# Add preprocessor definitions
+add_compile_definitions(EDGE_SDK_VERSION_STRING="${olp-cpp-sdk_VERSION}")
+
 # Add cmake dir to search list
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
--- a/olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
+++ b/olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
@@ -35,7 +35,10 @@
 
 namespace {
 constexpr auto kLogTag = "NetworkAndroid";
-}
+constexpr auto kUserAgentHeader = "User-Agent";
+constexpr auto kUserAgentAndroid =
+    "OLP-CPP-SDK/" EDGE_SDK_VERSION_STRING " (Android)";
+}  // namespace
 namespace olp {
 namespace http {
 
@@ -603,10 +606,30 @@ void NetworkAndroid::ResetRequest(JNIEnv* env, RequestId request_id) {
   req->second->Reinitialize();
 }
 
-jobjectArray NetworkAndroid::CreateExtraHeaders(
+bool NetworkAndroid::SetStringArrayElement(JNIEnv* env, jobjectArray array,
+                                           size_t index,
+                                           const std::string& str) {
+  jstring jstr = env->NewStringUTF(str.c_str());
+  if (!jstr || env->ExceptionOccurred()) {
+    EDGE_SDK_LOG_ERROR(kLogTag, "Failed to create string at index="
+                                    << index << "; with string=" << str);
+    return false;
+  }
+  utils::JNIScopedLocalReference str_ref(env, jstr);
+
+  env->SetObjectArrayElement(array, index, jstr);
+  if (env->ExceptionOccurred()) {
+    EDGE_SDK_LOG_ERROR(kLogTag, "Failed to set string at index="
+                                    << index << "; with string=" << str);
+    return true;
+  }
+  return true;
+}
+
+jobjectArray NetworkAndroid::CreateHeaders(
     JNIEnv* env,
-    const std::vector<std::pair<std::string, std::string> >& extra_headers) {
-  if ((extra_headers.empty())) {
+    const std::vector<std::pair<std::string, std::string> >& headers) {
+  if ((headers.empty())) {
     return 0;
   }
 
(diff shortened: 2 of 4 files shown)


olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
```diff
@@ -35,7 +35,10 @@
 
 namespace {
 constexpr auto kLogTag = "NetworkAndroid";
-}
+constexpr auto kUserAgentHeader = "User-Agent";
```
andescu: We need to define this headers in some common place like NetworkUtils.h or NetworkConstants.h, e.g. https://github.com/microsoft/cpprestsdk/blob/master/Release/include/cpprest/details/http_constants.dat

CMakeLists.txt
```diff
@@ -42,6 +42,9 @@ if (MSVC AND EDGE_SDK_MSVC_PARALLEL_BUILD_ENABLE)
     message(STATUS "MSVC - Using parallel compilation (/MP)")
 endif (MSVC AND EDGE_SDK_MSVC_PARALLEL_BUILD_ENABLE)
 
+# Add preprocessor definitions
+add_compile_definitions(EDGE_SDK_VERSION_STRING="${olp-cpp-sdk_VERSION}")
```
andescu: I would prefer to add individual numbers as defines.
```
add_compile_definitions(EDGE_SDK_VERSION_STRING="${olp-cpp-sdk_VERSION}")
add_compile_definitions(EDGE_SDK_VERSION_MAJOR="${olp-cpp-sdk_VERSION_MAJOR}")
add_compile_definitions(EDGE_SDK_VERSION_MINOR="${olp-cpp-sdk_VERSION_MINOR}")
add_compile_definitions(EDGE_SDK_VERSION_PATCH="${olp-cpp-sdk_VERSION_PATCH}")
```

olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
```diff
@@ -617,52 +640,48 @@ jobjectArray NetworkAndroid::CreateExtraHeaders(
   }
 
   // Merge vector of pair strings into one JNI array
-  const size_t header_count = extra_headers.size() * 2;
-  jobjectArray headers = (jobjectArray)env->NewObjectArray(
+  const size_t extra_headers_count = 1;
```
andescu: do you need and extra variable for this? Why not:
```
const size_t header_count = (headers.size() + 1) * 2;
```

olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
```diff
@@ -617,52 +640,48 @@ jobjectArray NetworkAndroid::CreateExtraHeaders(
...
-    }
+  // Setup extra headers:
+  if (!SetStringArrayElement(env, jheaders, headers.size() * 2,
+                             kUserAgentHeader)) {
+    EDGE_SDK_LOG_ERROR(kLogTag, "Failed to create extra header string: index="
+                                    << headers.size() * 2
+                                    << "; name=" << kUserAgentHeader);
+    return 0;
+  }
+  if (!SetStringArrayElement(env, jheaders, headers.size() * 2 + 1,
```
andescu: you need to check somehow if the user does not already has provided a similar header.
If this is the case then you probably need to merge them instead of providing duplicates. As far as I know not all HTTP headers allow duplicate entries.

olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
```diff
@@ -35,7 +35,10 @@
 
 namespace {
 constexpr auto kLogTag = "NetworkAndroid";
-}
+constexpr auto kUserAgentHeader = "User-Agent";
+constexpr auto kUserAgentAndroid =
```
andescu: can also be moved to common place like NetworkUtils.h or NetworkConstants.h

olp-cpp-sdk-core/src/http/ios/OLPNetworkIOS.mm
```diff
@@ -39,6 +39,8 @@
     static_cast<RequestId>(RequestIdConstants::RequestIdInvalid);
 
 constexpr auto kLogTag = "OLPNetworkIOS";
+constexpr auto kUserAgentHeader = "User-Agent";
```
andescu: see above comments

olp-cpp-sdk-core/src/http/ios/OLPNetworkIOS.mm
```diff
@@ -50,6 +52,17 @@
...
+
+  // Add OLP user-agent header
+  NSString* custom_user_agent = headers[@(kUserAgentHeader)];
+  NSMutableString* user_agent = [[NSMutableString alloc] init];
+  if (custom_user_agent.length > 0) {
+    [user_agent appendString:custom_user_agent];
+    [user_agent appendString:@"; "];
+  }
+  [user_agent appendString:@(kUserAgentiOS)];
+  headers[@(kUserAgentHeader)] = user_agent;
```
andescu: You need to check if this is already set and merge

olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
borzun: I checked that case on Android - this header will be merged automatically in the format:
User-Agent: <old_user_agent_value>; <new_user_agent_value>

olp-cpp-sdk-core/src/http/ios/OLPNetworkIOS.mm
borzun: I did this already - check lines 57 - 62

olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
andescu: Please add a comment then so that we know that the headers will be merged

olp-cpp-sdk-core/src/http/ios/OLPNetworkIOS.mm
andescu: I see. Thanks

borzun:
This pull request is closed because needed major rework.

CMakeLists.txt
borzun: Done, but used actual integers instead strings for Major / Minor / Patch versions (in order to perform versions check)

Maybe we should add the version integer in the way it boost does:
```
//  BOOST_VERSION % 100 is the patch level
//  BOOST_VERSION / 100 % 1000 is the minor version
//  BOOST_VERSION / 100000 is the major version

#define BOOST_VERSION 104500
```

olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
borzun: Not valid anymore

olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
borzun: Not valid anymore

olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
borzun: Done - see https://github.com/heremaps/here-olp-edge-sdk-cpp/pull/102

PR #102, opened 2019-08-22
Added the User-Agent header to http network request

borzun:
Added OLP-CPP-SDK User-Agent header for all request in olp-cpp-sdk-core with next format:
	User-Agent: OLP-SDK-CPP/<Version> (<PlatformName>)

Resolves: OLPEDGE-552

### DIFF
--- a/olp-cpp-sdk-core/src/client/OlpClient.cpp
+++ b/olp-cpp-sdk-core/src/client/OlpClient.cpp
@@ -19,6 +19,7 @@
 
 #include "olp/core/client/OlpClient.h"
 
+#include <cctype>
 #include <chrono>
 #include <future>
 #include <sstream>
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -104,14 +105,27 @@ std::shared_ptr<network::NetworkRequest> OlpClient::CreateRequest(
...
     networkRequest->AddHeader(header.first, header.second);
   }
+
+  std::string custom_user_agent;
   for (auto header : headerParams) {
-    networkRequest->AddHeader(header.first, header.second);
+    // Merge all User-Agent headers into one header.
+    // This is required for (at least) iOS network implementation,
+    // which uses headers dictionary without any duplicates.
+    if (header.first == http::kUserAgentHeader) {
```
andescu: you may need to compare lower case here.

PR #104, opened 2019-08-22
[DRAFT] Added User-Agent string to libCurl implementations

borzun:
+ fixed compilation on macOS with clang

Relates to: OLPEDGE-635


andescu:
Not needed as verified it trough curl verbose mode in scope of #102

PR #112, opened 2019-09-03
Temporary change network factory method to produce shared_ptr.

mykhailo-kuchma:
Changed network factory method to produce a shared_ptr instead of unique_ptr.
Currently network implementation are derived from std::enable_shared_from_this
and it requires to be a shared_ptr.

Relates-To: OLPEDGE-543

### DIFF
--- a/olp-cpp-sdk-core/src/http/Network.cpp
+++ b/olp-cpp-sdk-core/src/http/Network.cpp
@@ -19,7 +19,6 @@
 
 #include "olp/core/http/Network.h"
 
-#include "olp/core/porting/make_unique.h"
 #include "olp/core/utils/WarningWorkarounds.h"
 
 #ifdef NETWORK_HAS_CURL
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-core/src/http/Network.cpp
```diff
@@ -35,18 +34,18 @@
 namespace olp {
 namespace http {
 
-CORE_API std::unique_ptr<Network> CreateDefaultNetwork(size_t max_requests_count) {
+CORE_API std::shared_ptr<Network> CreateDefaultNetwork(size_t max_requests_count) {
 #ifdef NETWORK_HAS_CURL
   CORE_UNUSED(max_requests_count);
```
andescu: Define only once the CORE_UNUSED() please outside of the #ifdefs
You can still use the variable even if you mark it as unused.

mykhailo-kuchma: Done

PR #113, opened 2019-09-03
NetworkCurl body payload handling fix

sergii-vostrikov:
The body's data now stored at CURL's handle structure as a shared
pointer. This allows to be sure the data would not be destroyed during
the send process.

Relates-to: OLPEDGE-692

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
@@ -428,8 +428,8 @@ ErrorCode NetworkCurl::SendImplementation(
 
   const auto& config = request.GetSettings();
 
-  RequestHandle* handle =
-      GetHandle(id, callback, header_callback, data_callback, payload);
+  RequestHandle* handle = GetHandle(id, callback, header_callback,
+                                    data_callback, payload, request.GetBody());
   if (!handle) {
     return ErrorCode::UNKNOWN_ERROR;  // NetworkProtocol::ErrorNotReady;
   }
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -622,7 +623,8 @@ void NetworkCurl::AddEvent(EventInfo::Type type, RequestHandle* handle) {
 NetworkCurl::RequestHandle* NetworkCurl::GetHandle(
     int id, Network::Callback callback, Network::HeaderCallback header_callback,
     Network::DataCallback data_callback,
-    const std::shared_ptr<std::ostream>& payload) {
+    const std::shared_ptr<std::ostream>& payload,
```
andescu: both should be passed by copy with later moves

olp-cpp-sdk-core/src/http/curl/NetworkCurl.h
```diff
@@ -20,6 +20,7 @@
 #pragma once
 
 #include <olp/core/http/Network.h>
+#include <olp/core/http/NetworkRequest.h>
```
andescu: is already included via Network.h

sergii-vostrikov: We should prefer use principle of "if it's used it should be included". This allows to avoid to be dependent on side-effect includes.

andescu: ok

PR #114, opened 2019-09-03
Added a capture for a network body for winhttp implementation

mykhailo-kuchma:
Previously network requests was captured as a copy on upper levels (by a user),
and there was no such problem.

Relates-To: OLPEDGE-692


olp-cpp-sdk-core/src/http/winhttp/NetworkWinHttp.h
```diff
@@ -113,6 +113,7 @@ class NetworkWinHttp : public Network,
 #ifdef NETWORK_HAS_ZLIB
     z_stream strm;
 #endif
+    NetworkRequest::RequestBodyType body;
```
andescu: move up, next to payload

PR #118, opened 2019-09-03
[DRAFT] Fixed memory leak with NetworkAndroid instance

borzun:
Fixed problem with NetworkAndoird instance not being destroyed when the owner of it doesn't hold it anymore.
Thus, it will lead to the Java memory being released after network is not reachable anymore for client.

NOTE that this fix is not yet properly tested and should be verified before merge,
because it changes the way the NetworkAndroid::run_thread_ is being stopped.

Relates to: OLPEDGE-371


borzun:
Closing ticket due to refactoring of Java native methods with peer pattern (see OLPEDGE-735)

PR #119, opened 2019-09-03
Add ccache support to travis-ci builds

serhiikobalt:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,17 @@ addons:
     packages:
     - libboost-all-dev
     - libssl-dev
-    - libcurl4-openssl-dev 
+    - libcurl4-openssl-dev
     - clang
 matrix:
   include:
   - compiler: clang
+    cache: ccache
     env: BUILD_TYPE=RelWithDebInfo WORKSPACE=$TRAVIS_BUILD_DIR
-    name: Linux Build using clang 
+    name: Linux Build using clang
     script: $WORKSPACE/scripts/linux/psv/travis_build_psv.sh
   - compiler: gcc
+    cache: ccache
     env: BUILD_TYPE=COVERAGE WORKSPACE=$TRAVIS_BUILD_DIR
     name: Linux Build using gcc & tests & code coverage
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -2,6 +2,7 @@ language: cpp
 os: linux
 dist: xenial
 env: WORKSPACE=$TRAVIS_BUILD_DIR
+cache: ccache
```
andescu: ccache only makes sense when you can share it between builds. But this is a docker build which is initialized new on each trigger so you will basically always have a cold cache.

scripts/linux/psv/travis_build_psv.sh
```diff
@@ -1,5 +1,11 @@
...
+mkdir -p build
+cd build
+cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
+    -EDGE_SDK_BUILD_EXAMPLES=ON \
+    -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+    ..
+
 make -j8
 cd ..
```
andescu: To prove that this really works please add `ccache -s`so that we see how many cache hits we had.

scripts/macos/psv/travis_build_psv.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -xe
+mkdir -p build
+cd build
+cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
+    -EDGE_SDK_BUILD_EXAMPLES=ON \
+    -DBUILD_SHARED_LIBS=ON \
+    ..
+make -j
+cd ..
```
andescu: same here

scripts/macos/psv/travis_build_psv.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -xe
+mkdir -p build
+cd build
+cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
+    -EDGE_SDK_BUILD_EXAMPLES=ON \
+    -DBUILD_SHARED_LIBS=ON \
+    ..
+make -j
```
sergii-vostrikov: j8 ?

PR #121, opened 2019-09-04
Adapts StreamLayerClient to use TaskScheduler for Flush context.

andescu:
This removed the raw std::thread usage and triggeres instead a task
to be executed in the context of the TaskScheduler. AutoFlushController
is the only class remaining that still uses raw std::thread but this
will need a big refactoring and will be postponed post v0.7.0 release.

Relates-to: OLPEDGE-457

### DIFF
--- a/examples/dataservice-write/example.cpp
+++ b/examples/dataservice-write/example.cpp
@@ -24,6 +24,7 @@
 
 #include <olp/authentication/TokenProvider.h>
 #include <olp/core/client/HRN.h>
+#include <olp/core/client/OlpClientSettings.h>
 #include <olp/core/client/OlpClientSettingsFactory.h>
 #include <olp/core/logging/Log.h>
 
@@ -41,6 +42,7 @@ const std::string kLayer("");            // layer name inside catalog to use
 const std::string kData("hello world");  // data to write
 
 constexpr auto kLogTag = "write-example";
+constexpr size_t kPublishRequestsSize = 5u;
 }  // namespace
 
 int RunExample() {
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-write/src/StreamLayerClientImpl.cpp
```diff
@@ -360,12 +372,11 @@ StreamLayerClientImpl::Flush() {
...
     const StreamLayerClient::FlushCallback& callback) {
-  auto cancel_context = std::make_shared<CancellationContext>();
+  CancellationContext cancel_context;
 
-  // TODO: We should introduce an Executor abstraction to handle thread
-  // spwaning/pooling in a centralized/ configurable way
   auto self = shared_from_this();
-  auto flush_thread = std::thread([self, callback, cancel_context]() {
+
+  auto func = [=]() mutable {
```
borzun: I think it is better to capture `cancel_context` explicitly, cause we had problem with 'capture everything' policy in other places, what do you think? 

It can be handy when debugging/reading the lambdas; it will be easier to track which actual object is used in lambda's body and which is not.

examples/dataservice-write/example.cpp
```diff
@@ -40,6 +40,7 @@ const std::string kLayer("");            // layer name inside catalog to use
 const std::string kData("hello world");  // data to write
 
 constexpr auto kLogTag = "write-example";
+constexpr size_t kSizePublish = 5u;
```
borzun:
```suggestion
constexpr size_t kPublishRequestsSize = 5u;
```
Or, I don't know, maybe `kNumRequestsToPublish` or `kRepeatPublishRequestCount`, cause current name is a bit unclear.

olp-cpp-sdk-dataservice-write/src/StreamLayerClientImpl.cpp
```diff
@@ -375,45 +386,49 @@ olp::client::CancellationToken StreamLayerClientImpl::Flush(
...
-           self->QueueSize() > 0) {
-      auto publish_data_request = self->PopFromQueue();
-      if (publish_data_request == boost::none) {
-        continue;
+    while ((!maximum_events_number || counter < *maximum_events_number) &&
+           (self->QueueSize() > 0) && !cancel_context.IsCancelled()) {
+      auto publish_request = self->PopFromQueue();
+      if (publish_request == boost::none) {
+        // Nothing to flush
+        break;
```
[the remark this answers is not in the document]
andescu: During Queue() invalid PublishRequests should be catched and never instered into the queue.
Looking at PopFromQueue() I see that also when the queue is empty we receive boost::none which together with `continue` will result in an endless loop. But ill change it back to `continue` and hope that `self->QueueSize() > 0` will reduce the harm.

olp-cpp-sdk-dataservice-write/src/StreamLayerClientImpl.cpp
```diff
@@ -48,20 +48,32 @@
 using namespace olp::client;
 using namespace olp::dataservice::write::model;
 
-#define LOG_TAG "olp::dataservice::write::StreamLayerClient"
+namespace olp {
+namespace dataservice {
+namespace write {
 
 namespace {
+constexpr auto kLogTag = "StreamLayerClient";
```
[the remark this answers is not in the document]
andescu: Done.

olp-cpp-sdk-dataservice-write/src/StreamLayerClientImpl.cpp
andescu: As from my past experience when capturing explicit you should also add some warnings to compilation for unused captures else you could face some nasty memory leaks.
With default capture the compiler takes care that only what is used is captured. Usually there is no difference between them except you want to guard against using something that you should not. Might be that we had issues because we didn't use what should have been captured. But this is a design issue and we should not solve this by just capture individually.

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
```diff
@@ -1610,7 +1620,6 @@ TEST_P(StreamLayerClientCacheOnlineTest,
 
   EXPECT_LE(3, default_listener->GetNumFlushEvents());
   EXPECT_LE(3, default_listener->GetNumFlushEventsAttempted());
-  EXPECT_EQ(0, default_listener->GetNumFlushEventsFailed());
```
borzun: Better to not remove this line, but leave it commented with description why this check was removed.

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
```diff
@@ -1581,7 +1589,7 @@ TEST_P(StreamLayerClientCacheOnlineTest,
...
+  // EXPECT_EQ(0, default_listener->GetNumFlushEventsFailed());
```
borzun: Better to add the comment why it was commented (about actual problem with this failed count)

PR #127, opened 2019-09-04
Fixed data race when generating NetworkAndroid's unique id

borzun:
Fixed problem with `StreamLayerClientOnlineTest.ConcurrentPublishDifferentIngestApi` test, which
was deadlocked because of invalid NetworkAndroid uniqueId generating policy:
there was the data race that there are could be created several instances of NetworkAndroid,
which had same unique id, thus, it resulted that some requests were lost.

Relates to: OLPEDGE-697

### DIFF
--- a/olp-cpp-sdk-core/src/http/android/HttpClient.java
+++ b/olp-cpp-sdk-core/src/http/android/HttpClient.java
@@ -37,6 +37,7 @@ import java.net.URLConnection;
 import java.net.UnknownHostException;
 import java.net.SocketTimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/src/http/android/HttpClient.java
```diff
@@ -51,7 +52,8 @@
   public static final int TIMEOUT_ERROR = -7;
   public static final int THREAD_POOL_SIZE = 8;
 
-  private int uniqueId = (int) (System.currentTimeMillis() % 10000);
+  private static AtomicInteger uniqueIdCounter = new AtomicInteger(0);
+  private int uniqueId = -1;
```
mykhailo-kuchma: Currently, requestId 0 is considered invalid on cpp side, is it ok here?

borzun: It is not request id here, it is the id of NetworkAndroid object. 

But I can change that in order to align it with other places in code,

PR #128, opened 2019-09-04
Adapt CatalogClient tests.

mykhailo-kuchma:
Adapt CatalogClient tests to the latest changes.

Relates-To: OLPEDGE-545

### DIFF
--- a/olp-cpp-sdk-dataservice-read/test/unit/src/CatalogClientTest.cpp
+++ b/olp-cpp-sdk-dataservice-read/test/unit/src/CatalogClientTest.cpp
@@ -751,6 +751,7 @@ class CatalogClientMockTest : public CatalogClientTestBase {
   void TearDown() {
     client_.reset();
     settings_.reset();
+    Mock::VerifyAndClearExpectations(network_mock_.get());
     network_mock_.reset();
   }
 
@@ -1388,7 +1389,7 @@ TEST_P(CatalogClientMockTest, GetPartitionsGarbageResponse) {
             partitionsResponse.GetError().GetErrorCode());
 }
 
-TEST_P(CatalogClientMockTest, DISABLED_GetCatalogCancelApiLookup) {
+TEST_P(CatalogClientMockTest, GetCatalogCancelApiLookup) {
   olp::client::HRN hrn(GetTestCatalog());
 
   auto waitForCancel = std::make_shared<std::promise<void>>();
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-dataservice-read/test/unit/src/CatalogClientTest.cpp
```diff
@@ -754,6 +754,7 @@ class CatalogClientMockTest : public CatalogClientTestBase {
   void TearDown() {
     client_.reset();
     settings_.reset();
+    Mock::VerifyAndClearExpectations(network_mock_.get());
```
andescu: If it is really important to check this resp to fail if you have unmet expectations then i would consider using StrictMock around NetworkMock. This way everything fails that is not expected or expected and not met.

olp-cpp-sdk-dataservice-read/test/unit/src/CatalogClientTest.cpp
```diff
@@ -940,6 +941,7 @@ class CatalogClientMockTest : public CatalogClientTestBase {
 
     // Catch any non-interesting network calls that don't need to be verified
     EXPECT_CALL(*network_mock_, Send(_, _, _, _, _)).Times(testing::AtLeast(0));
+    EXPECT_CALL(*network_mock_, Cancel(_)).Times(testing::AtLeast(0));
```
andescu:
```suggestion
    EXPECT_CALL(*network_mock_, Cancel(_)).Times(0);
```
Or as specified before, even Google is recommending this, use StrictMock to guard against unexpected calls:

```
[INFO] StreamLayerClientImpl - Flushed 1 publish requests

GMOCK WARNING:
Uninteresting mock function call - returning directly.
    Function call: Cancel(5)
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#knowing-when-to-expect for details.
```

olp-cpp-sdk-dataservice-read/test/unit/src/CatalogClientTest.cpp
```diff
@@ -2375,16 +2461,15 @@ TEST_P(CatalogClientMockTest, GetCatalogCacheWithUpdate) {
   waitForEnd->get_future().get();
 
   // Request 2 to check there is a cached value.
-  EDGE_SDK_LOG_TRACE_F("CatalogClientMockTest", "Request Catalog, CacheOnly"); 
-  request.WithFetchOption(CacheOnly); future =
-    catalogClient->GetCatalog(request);
+  EDGE_SDK_LOG_TRACE_F("CatalogClientMockTest", "Request Catalog, CacheOnly");
```
andescu: Hmm. You should not use logs in tests. Use SCOPED_TRACE("....") to see it only when test fails.

olp-cpp-sdk-dataservice-read/test/unit/src/CatalogClientTest.cpp
```diff
@@ -2375,16 +2461,15 @@ TEST_P(CatalogClientMockTest, GetCatalogCacheWithUpdate) {
...
+  request.WithFetchOption(CacheOnly);
+  future = catalogClient->GetCatalog(request);
   EDGE_SDK_LOG_TRACE_F("CatalogClientMockTest", "get CatalogResponse2");
```
andescu: Same here.

PR #130, opened 2019-09-05
 OLPEDGE-383. Shared libs ON. Libsegfault lib added.

ystefinko:
Shared libs tuned ON. 
Linux libsegfault library added for segmentation fault crash identification in logs.

Resolves: OLPEDGE-383


ystefinko:
Fixed comment and updated branch with latest master fixes.

PR #131, opened 2019-09-05
Removed WinHttp inheritance from std::enable_shared_from_this

mykhailo-kuchma:
Removed WinHttp inheritance from std::enable_shared_from_this.
Fixes the issue with new network design when network expected to be
destroyed after the last pointer is destroyed.

Relates-To: OLPEDGE-544

### DIFF
--- a/olp-cpp-sdk-core/src/http/winhttp/NetworkWinHttp.cpp
+++ b/olp-cpp-sdk-core/src/http/winhttp/NetworkWinHttp.cpp
@@ -320,7 +320,7 @@ SendOutcome NetworkWinHttp::Send(NetworkRequest request,
                                            url_components.lpszScheme));
     auto& connection = http_connections_[server];
     if (!connection) {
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/src/http/winhttp/NetworkWinHttp.cpp
```diff
@@ -945,8 +949,7 @@ NetworkWinHttp::ResultData::ResultData(RequestId id, Network::Callback callback,
       completed(false),
       cancelled(false) {}
 
-NetworkWinHttp::ConnectionData::ConnectionData(
-    std::shared_ptr<NetworkWinHttp> owner)
+NetworkWinHttp::ConnectionData::ConnectionData(NetworkWinHttp* owner)
     : self(std::move(owner)), http_connection(NULL) {}
```
andescu:
```suggestion
    : self(owner), http_connection(NULL) {}
```

PR #132, opened 2019-09-05
Keep network instance in unit tests alive until inner thread is running

serhiikobalt:
Network is instantiated once for every test suite and reused in all tests of that suite.

Relates-to: OLPEDGE-544

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
@@ -199,11 +199,11 @@ NetworkCurl::~NetworkCurl() {
 }
 
 bool NetworkCurl::Initialize() {
+  std::lock_guard<std::mutex> init_lock(init_mutex_);
   if (state_ != WorkerState::STOPPED) {
     EDGE_SDK_LOG_DEBUG(kLogTag, "Already initialized");
     return true;
   }
-  std::lock_guard<std::mutex> init_lock(init_mutex_);
 
 #ifdef NETWORK_HAS_PIPE2
   if (pipe2(pipe_, O_NONBLOCK)) {
@@ -284,15 +284,10 @@ void NetworkCurl::Deinitialize() {
   }
 
   std::lock_guard<std::mutex> init_lock(init_mutex_);
+  // We should not destroy this thread from itself
   if (thread_.get_id() != std::this_thread::get_id()) {
     event_condition_.notify_all();
     thread_.join();
-  } else {
-    // We are trying to stop the very thread we are in. This is not recommended,
-    // but we try to handle it gracefully. This could happen by calling from one
-    // of the static functions (rxFunction or headerFunction) that was passed to
-    // the cURL as callbacks.
-    thread_.detach();
   }
 }
 
--- a/olp-cpp-sdk-dataservice-write/test/unit/TestIndexLayerClient.cpp
+++ b/olp-cpp-sdk-dataservice-write/test/unit/TestIndexLayerClient.cpp
@@ -136,9 +136,15 @@ class IndexLayerClientTestBase : public ::testing::TestWithParam<bool> {
 
 class IndexLayerClientOnlineTest : public IndexLayerClientTestBase {
  protected:
-  virtual std::shared_ptr<IndexLayerClient> CreateIndexLayerClient() override {
-    auto network = olp::client::OlpClientSettingsFactory::
+  static std::shared_ptr<olp::http::Network> s_network;
+
+  static void SetUpTestSuite() {
+    s_network = olp::client::OlpClientSettingsFactory::
         CreateDefaultNetworkRequestHandler();
+  }
+
+  virtual std::shared_ptr<IndexLayerClient> CreateIndexLayerClient() override {
+    auto network = s_network;
 
     olp::authentication::Settings authentication_settings;
     authentication_settings.token_endpoint_url =
@@ -160,6 +166,11 @@ class IndexLayerClientOnlineTest : public IndexLayerClientTestBase {
         olp::client::HRN{GetTestCatalog()}, settings);
   }
 };
+
+// Static network instance is necessary as it needs to outlive any created
+// clients. This is a known limitation as triggered send requests capture the
+// network instance inside the callbacks.
+std::shared_ptr<olp::http::Network> IndexLayerClientOnlineTest::s_network;
 
 INSTANTIATE_TEST_SUITE_P(TestOnline, IndexLayerClientOnlineTest,
                          ::testing::Values(true));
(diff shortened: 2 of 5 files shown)


olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -284,15 +284,13 @@ void NetworkCurl::Deinitialize() {
   }
 
   std::lock_guard<std::mutex> init_lock(init_mutex_);
+  // We are trying to stop the very thread we are in. This is not recommended,
```
andescu: You cannot join() from the same thread. And based on the `if (thread_.get_id() != std::this_thread::get_id())` guard you also do not stop from the same thread you are in.
The comment should be changed to reflect that we should not stop from the same thread we are in.

olp-cpp-sdk-dataservice-write/test/unit/TestIndexLayerClient.cpp
```diff
@@ -159,6 +158,10 @@ class IndexLayerClientOnlineTest : public IndexLayerClientTestBase {
     return std::make_shared<IndexLayerClient>(
         olp::client::HRN{GetTestCatalog()}, settings);
   }
+
+  std::shared_ptr<olp::http::Network> network_{
```
andescu: Is there any difference between this and the previous initialization? At the end you still create one and destroy one instance on each test case.
More stable would be to have it in the static SetUpTestSuite() and TearDownTestSuite() and only create it once and destroy it once until we fixed the issues and made it more dynamic

olp-cpp-sdk-dataservice-write/test/DataserviceWriteTestMain.cpp
```diff
@@ -20,7 +20,14 @@
 #include <gtest/gtest.h>
 #include "testutils/CustomParameters.hpp"
 
+#include <olp/core/client/OlpClientSettingsFactory.h>
+
+std::shared_ptr<olp::http::Network> g_network;
+
 int main(int argc, char** argv) {
+  g_network = olp::client::OlpClientSettingsFactory::
```
borzun: There are some tests, which test parallel execution (like `StreamLayerClientOnlineTest.ConcurrentPublishDifferentIngestApi`). 

By executing requests in parallel and creating several instances of network, we can check whether we do not have any data race inside network implementations, so, I do not recommend to create shared instance of network per thread.

serhiikobalt: this network instance is created before any thread is spawned, it's shared across all threads.

borzun: Looks like indeed it was the original intention behind this test, but it helped me to find my problem. :) I agree that we might need indeed have such tests for `Network` itself.

olp-cpp-sdk-dataservice-write/test/unit/TestIndexLayerClient.cpp
```diff
@@ -136,9 +136,15 @@ class IndexLayerClientTestBase : public ::testing::TestWithParam<bool> {
 
 class IndexLayerClientOnlineTest : public IndexLayerClientTestBase {
  protected:
-  virtual std::shared_ptr<IndexLayerClient> CreateIndexLayerClient() override {
-    auto network = olp::client::OlpClientSettingsFactory::
+  static std::shared_ptr<olp::http::Network> s_network;
```
andescu:
```suggestion
  // Static network instance is necessary as it needs to outlive any created clients.
  // This is a known limitation as triggered send requests capture the network instance inside the callbacks.
  static std::shared_ptr<olp::http::Network> s_network;
```

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
```diff
@@ -191,11 +191,17 @@ class StreamLayerClientTestBase : public ::testing::TestWithParam<bool> {
 };
 
 class StreamLayerClientOnlineTest : public StreamLayerClientTestBase {
+  static std::shared_ptr<olp::http::Network> s_network;
```
andescu:
```suggestion
  // Static network instance is necessary as it needs to outlive any created clients.
  // This is a known limitation as triggered send requests capture the network instance inside the callbacks.
  static std::shared_ptr<olp::http::Network> s_network;
```

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
```diff
@@ -1250,10 +1258,16 @@ TEST_P(StreamLayerClientMockTest, DISABLED_SDIIConcurrentPublishSameIngestApi) {
 
 class StreamLayerClientCacheOnlineTest : public StreamLayerClientOnlineTest {
  protected:
+  static std::shared_ptr<olp::http::Network> s_network;
```
andescu:
```suggestion
  // Static network instance is necessary as it needs to outlive any created clients.
  // This is a known limitation as triggered send requests capture the network instance inside the callbacks.
  static std::shared_ptr<olp::http::Network> s_network;
```

olp-cpp-sdk-dataservice-write/test/unit/TestVersionedLayerClient.cpp
```diff
@@ -78,16 +78,19 @@ TEST_P(VersionedLayerClientOfflineTest, StartBatchRequestTest) {
 }
 
 class VersionedLayerClientOnlineTest : public VersionedLayerClientTest {
- public:
  protected:
+  static std::shared_ptr<olp::http::Network> s_network;
```
andescu:
```suggestion
  // Static network instance is necessary as it needs to outlive any created clients.
  // This is a known limitation as triggered send requests capture the network instance inside the callbacks.
  static std::shared_ptr<olp::http::Network> s_network;
  
```

Add newline before `SetUpTestSuite()`

olp-cpp-sdk-dataservice-write/test/unit/TestVolatileLayerClient.cpp
```diff
@@ -108,10 +108,15 @@ class VolatileLayerClientTestBase : public ::testing::TestWithParam<bool> {
 
 class VolatileLayerClientOnlineTest : public VolatileLayerClientTestBase {
  protected:
+  static std::shared_ptr<olp::http::Network> s_network;
```
andescu:
```suggestion
  // Static network instance is necessary as it needs to outlive any created clients.
  // This is a known limitation as triggered send requests capture the network instance inside the callbacks.
  static std::shared_ptr<olp::http::Network> s_network;
  
```

Add newline before `SetUpTestSuite()`

PR #143, opened 2019-09-07
Added a local server for test purposes

mykhailo-kuchma:
Test network implementation against local server


olp-cpp-sdk-core/tests/http/unittest/TestNetwork.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+TEST(TestNetwork, get_request) {
+  auto network = olp::http::CreateDefaultNetwork(1);
+
+  auto payload = std::make_shared<std::stringstream>();
+
+  std::mutex m;
+  std::condition_variable cv;
+
+  olp::http::NetworkRequest request("http://localhost:3000/get_request");
```
andescu: We should use proxy settings for this and let the SDK call normal URL so that we can react based on those.

mykhailo-kuchma:
> Looks promising good, but i would extend a little bit the functionality of the server and make it more generic.
> Maybe we can introduce some kind of json test file where we can specify what catalog, layers, partitions, etc. to use and what to send back in case this is received. That way we do not need to change the server itself every time we introduce a new test case but rather extend/adapt the json file.

Of course, this is just a basic setup. To make sure we have the same network behavior across platforms.

olp-cpp-sdk-core/tests/http/unittest/TestNetwork.cpp
mykhailo-kuchma: added also a test for HTTP proxy with basic auth.

andescu: So these are only for testing Network layer or clients?

mykhailo-kuchma: Network at the moment

ystefinko:
I've reviewed .travis.yml. Looks good.

PR #144, opened 2019-09-08
Added codecov yaml

mykhailo-kuchma:
Added codecov yaml to include tests directories from code coverage reports.

Resolves: OLPEDGE-744

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,28 @@
+coverage:
+  range: 60..100
+  round: down
+  precision: 1
+  status:
+    project:
+      core:
+        paths:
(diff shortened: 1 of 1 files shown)


andescu:
> Commit message is too long and it should be "exclude" instead of "include"!

Also, ticket is missing.

codecov.yml
```diff
@@ -0,0 +1,6 @@
+ignore:
```
andescu: Can you also please add threshold 1% so that we do not get all the false negatives on unrelated changes.
See https://docs.codecov.io/docs/commit-status

PR #145, opened 2019-09-08
Fixed flaky tests about waiting for the getBatchResponse success

borzun:
Some dataservice-write tests uses the same getBatchResponse waiting logic to determine whether batch was correctly submitted and succeeded. However, because of some limitations on server, we might not receive the response in time (i.e. after 100 iterations). Thus, such tests appear to be flaky. Because of this, need to temporarily disable failing checks in those tests until proper solution will be found.

Relates to: OLPEDGE-697, OLPEDGE-708

### DIFF
--- a/olp-cpp-sdk-dataservice-write/test/unit/TestVersionedLayerClient.cpp
+++ b/olp-cpp-sdk-dataservice-write/test/unit/TestVersionedLayerClient.cpp
@@ -200,7 +200,13 @@ TEST_P(VersionedLayerClientOnlineTest, StartBatchTest) {
       break;
     }
   }
-  ASSERT_EQ("succeeded", getBatchResponse.GetResult().GetDetails()->GetState());
+  // There are can be a case that GetBatch() is not "succeeded", but in
+  // "submitted" state even after 100 iterations,
+  // which actually means that there are might be a problem on the server side,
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-dataservice-write/test/unit/TestVersionedLayerClient.cpp
```diff
@@ -505,7 +518,13 @@ TEST_P(VersionedLayerClientOnlineTest, PublishToBatchMultiTest) {
...
     }
   }
-  ASSERT_EQ("succeeded", getBatchResponse.GetResult().GetDetails()->GetState());
+  // There are can be a case that GetBatch() is not "succeeded", but in
+  // "submitted" state even after 100 iterations,
+  // which actually means that there are might be a problem on the server side,
+  // (or just long delay). Thus, better to rewrite this test, or do not rely on
+  // the real server, but use mocked server.
+  // ASSERT_EQ("succeeded",
+  // getBatchResponse.GetResult().GetDetails()->GetState());
```
sergii-vostrikov: In the first place, should the test even check for value set by a server?
It is OK to analyze if the field is present since SDK parses it from response JSON. But the values "submitted" or "succeeded" are not in the SDKs responsibility.

I propose to remove the checks for details' field completely from out (online) unit tests projects, for all layers affected.

PR #149, opened 2019-09-09
Enable Volatile Layer Client write tests

serhiikobalt:
* fix disabled tests
 * refactor cancellation tests

Relates-To: OLPEDGE-710

### DIFF
--- a/olp-cpp-sdk-dataservice-write/test/unit/TestVolatileLayerClient.cpp
+++ b/olp-cpp-sdk-dataservice-write/test/unit/TestVolatileLayerClient.cpp
@@ -32,6 +32,8 @@
 
 #include "HttpResponses.h"
 
+#include "TestCommons.h"
+
 using namespace olp::dataservice::write;
 using namespace olp::dataservice::write::model;
 
@@ -541,6 +543,10 @@ class VolatileLayerClientMockTest : public VolatileLayerClientTestBase {
  protected:
   std::shared_ptr<NetworkMock> network_;
 
+  void TearDown() override {
+    testing::Mock::VerifyAndClearExpectations(network_.get());
+  }
+
   virtual std::shared_ptr<VolatileLayerClient> CreateVolatileLayerClient()
       override {
     olp::client::OlpClientSettings client_settings;
@@ -612,7 +618,7 @@ class VolatileLayerClientMockTest : public VolatileLayerClientTestBase {
 INSTANTIATE_TEST_SUITE_P(TestMock, VolatileLayerClientMockTest,
                          ::testing::Values(false));
 
-TEST_P(VolatileLayerClientMockTest, DISABLED_PublishData) {
+TEST_P(VolatileLayerClientMockTest, PublishData) {
   auto new_client = CreateVolatileLayerClient();
   {
     testing::InSequence dummy;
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
```diff
@@ -1923,44 +1972,48 @@ TEST_P(StreamLayerClientCacheMockTest, FlushDataMultiple) {
...
-  auto response = cancel_future.GetFuture().get();
+  auto promise = client_->Flush();
+  waitForCancel->get_future().get();
+  promise.GetCancellationToken().cancel();
+  pauseForCancel->set_value();
+
+  auto response = promise.GetFuture().get();
 
   ASSERT_EQ(1, response.size());
+  testing::Mock::VerifyAndClearExpectations(network_.get());
```
andescu: Adding additionally this is not really necessary as google does it for you https://github.com/google/googletest/blob/master/googlemock/docs/cheat_sheet.md#verifying-and-resetting-a-mock
-> *gMock will verify the expectations on a mock object when it is destructed ...*

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
```diff
@@ -112,6 +114,17 @@ void PublishSdiiSuccessAssertions(
   EXPECT_FALSE(result.GetResult().GetTraceID().GetGeneratedIDs().at(0).empty());
 }
 
+template <typename T>
+void PublishCancelledAssertions(
```
andescu: What is PublishCancelledAssertions?
Do you mean CheckResponse?

serhiikobalt: These assertions are shared across all write tests where publish requests are cancelled.

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
serhiikobalt: sometimes network mock might be never destroyed as reference to is is hold by expectation inside the mock

andescu: Then move it to TearDown() please

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
andescu: The name of the function is wrong. What do you publish? I dont see any publish? What you do in the function is checking if the result is ok. So the function should be called CheckResponse()

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
serhiikobalt: In my opinion TearDown is not the place to verify assertions or expectations on mocks event.

olp-cpp-sdk-dataservice-write/test/unit/TestIndexLayerClient.cpp
```diff
@@ -115,6 +117,14 @@ class IndexLayerClientTestBase : public ::testing::TestWithParam<bool> {
     return index;
   }
 
+  std::string PrintError(const olp::client::ApiError& error) {
```
andescu: This is not used. Remove?

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
serhiikobalt: there are already functions named 
```
PublishDataSuccessAssertions()
PublishFailureAssertions()
```
Also similar names are used in tests for other layer types. I agree the naming is wrong but at least it is consistent.

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
andescu: En contraire mon frere, see https://github.com/google/googletest/blob/master/googletest/docs/faq.md#should-i-use-the-constructordestructor-of-the-test-fixture-or-setupteardown-ctorvssetup

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
andescu: When moving you have the chance to change it.

olp-cpp-sdk-dataservice-write/test/unit/TestStreamLayerClient.cpp
serhiikobalt: I didn't get that. It tells that I might want to prefer TearDown over destructors. Nothing there tells that VerifyAndClearExpectations should be preferably called in TearDown and there are plenty of examples when it called in tests. Placing this call inside TearDown might help code deduplication, but believe having all assertions in one place at the buttom of each test is more readable and less astonishing.

PR #150, opened 2019-09-10
Enabled TestCacheMock tests on PSV

borzun:
Enabled the failing StreamLayer's TestCacheMock tests (i.e. StreamLayerClientCacheMockTest suite) on PSV job.
The reason was that those tests were flagged as "online", however they are actually "offline", thus, the CI job couldn't get the env variables in order to run those tests.

Fixes: OLPEDGE-626


borzun:
Closing pull request - will create a draft request with suggested changes instead

PR #162, opened 2019-09-12
Move different auth tests to separate files

serhiikobalt:
Move different auth tests to separate files

    Split AuthenticationTest in to multiple:
     * AuthenticationOfflineTest
     * AuthenticationOnlineTest
     * ArcGisAuthenticationOnlineTest
     * FacebookAuthenticationOnlineTest
     * GoogleAuthenticationOnlineTest

    Split Here Oauth into online and offline part.

    Follow naming convention for test suits.

    Relates-to: OLPEDGE-594


serhiikobalt:
@andescu  there is a rule that name of file should match name of test suite. I can't enforce that without either renaming tests or splitting them across different files

andescu:
We should not exaggerate and split test fixtures to .h and .cpp files. This is unnecessary and pollutes the global namespace.
The test fixtures should usually not be reused, except they are used as base class for the actual test fixture, e.g. AuthenticationBaseTest.h. Everything else is unnecessarily exported into the global namespace. Please change and move all to .cpp files only. Especially mocks!

Also as already mentioned in #160 you only need to link against gmock_main as it already includes gtest, gmock and a main.cpp, means

```
target_link_libraries(olp-cpp-sdk-authentication-tests
        PRIVATE
            custom-params
            gmock_main
            olp-cpp-sdk-authentication
    )
```

PR #164, opened 2019-09-13
[DRAFT] Fixed the deadlock in StreamLayerClientClientImpl::Flush

borzun:
1. Fixed the deadlock in StreamLayerClientImpl::Flush, which caused by deleting the only instance of ThreadPoolTaskScheduler by its worker thread. Thus, it resulted in deadlock, because worker thread tries to join itself. Fixed by detaching if thread differs.
2. Enabled the StreamLayerClientCacheMockTest tests for StreamLayerClient on PSV job. The reason that they were disabled was that those tests were flagged as "online" in a code, however they are actually "offline", thus, the CI job couldn't get the env variables in order to run those tests.

Relates-to: OLPEDGE-626
Fixes: OLPEDGE-712


borzun:
Closed because of new approach for refactoring of AutoFlushController

PR #165, opened 2019-09-13
Fixed some disabled unit test GetCatalogClient

sergii-vostrikov:
Update-related tests no longer wait for cancel signal.

Relates-to: OLPEDGE-678

### DIFF
--- a/olp-cpp-sdk-dataservice-read/tests/CatalogClientTest.cpp
+++ b/olp-cpp-sdk-dataservice-read/tests/CatalogClientTest.cpp
@@ -702,9 +702,8 @@ generateNetworkMocks(std::shared_ptr<std::promise<void>> pre_signal,
                         callback_placeholder](olp::http::RequestId id) {
     if (!completed->exchange(true)) {
       auto cancel_code = static_cast<int>(ErrorCode::CANCELLED_ERROR);
-      (*callback_placeholder)(NetworkResponse()
-                                  .WithError("Cancelled")
-                                  .WithStatus(cancel_code));
+      (*callback_placeholder)(
+          NetworkResponse().WithError("Cancelled").WithStatus(cancel_code));
     }
   };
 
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-dataservice-read/tests/CatalogClientTest.cpp
```diff
@@ -1634,10 +1634,14 @@ TEST_P(CatalogClientMockTest, DISABLED_GetPartitionsCancelLayerVersions) {
       waitForCancel, pauseForCancel, {200, HTTP_RESPONSE_LAYER_VERSIONS});
```
andescu: Dont you need to change {200, HTTP_RESPONSE_LAYER_VERSIONS} to send a lookup response back instead of HTTP_RESPONSE_LAYER_VERSIONS

sergii-vostrikov: Good point, will add that for consistency. Although that has no effect since the test cancels the request right at the time of response is generated.

olp-cpp-sdk-dataservice-read/tests/CatalogClientTest.cpp
```diff
@@ -1630,13 +1630,16 @@ TEST_P(CatalogClientMockTest, DISABLED_GetPartitionsCancelLayerVersions) {
...
 
-  EXPECT_CALL(*network_mock_,
-              Send(IsGetRequest(URL_LAYER_VERSIONS), _, _, _, _))
+  EXPECT_CALL(*network_mock_, Send(IsGetRequest(URL_LOOKUP_CONFIG), _, _, _, _))
       .Times(1)
       .WillOnce(testing::Invoke(std::move(send_mock)));
 
+  EXPECT_CALL(*network_mock_,
+              Send(IsGetRequest(URL_LAYER_VERSIONS), _, _, _, _))
+      .Times(0);  // TODO: make test to wait/cancel at URL_LAYER_VERSIONS state
```
andescu: As discussed, please have a look if the test case is still testing what it says: GetPartitionsCancelLayerVersions.
Currently it looks like we are cancelling look-up API call instead.

sergii-vostrikov: cancel* tests fixes to enabled be in an upcoming pull-request

PR #168, opened 2019-09-16
Remove cancelled flag from NetworkResponse

mykhailo-kuchma:
Cancellation flag is duplicated by status code.
Method is not used by SDK.

Resolves: OLPEDGE-780

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/http/NetworkResponse.h
+++ b/olp-cpp-sdk-core/include/olp/core/http/NetworkResponse.h
@@ -39,13 +39,6 @@ class CORE_API NetworkResponse final {
   bool IsCancelled() const;
 
   /**
-   * @brief Set cancelled field.
-   * @param[in] cancelled True if associated request was cancelled.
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-core/include/olp/core/http/NetworkResponse.h
```diff
@@ -32,19 +32,6 @@ namespace http {
  */
 class CORE_API NetworkResponse final {
  public:
-  /**
-   * @brief Check if associated request was cancelled.
-   * @return true if associated request was cancelled.
-   */
-  bool IsCancelled() const;
```
andescu: Might be helpfull to keep this and instead of using the bool use the status code, e.g. 

```cpp
bool IsCancelled() const {
  return status_ == ErrorCode::CANCELLED_ERROR;
}
```

mykhailo-kuchma: Added a check

PR #177, opened 2019-09-17
Separate unit & integration core tests

serhiikobalt:
* use different ENV variable names for dataservice-read & -write
 * move 2 online tests from core to tests/integration

Relates-To: OLPEDGE-717

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,9 @@ add_sdks()
 add_subdirectory(docs)
 
 if(EDGE_SDK_ENABLE_TESTING)
+    # Add tests
     add_subdirectory(tests/functional)
+    add_subdirectory(tests/integration)
 endif()
 
 # Add example
@@ -93,4 +95,3 @@ endif()
 
 # Add uninstall script
 add_custom_target(uninstall "${CMAKE_COMMAND}" -P "${CMAKE_MODULE_PATH}/uninstall.cmake")
-
--- a/scripts/linux/fv/olp-cpp-sdk-integration-test.variables
+++ b/scripts/linux/fv/olp-cpp-sdk-integration-test.variables
@@ -1,3 +1,6 @@
+export dataservice_read_test_secret="${dataservice_read_test_secret}" #defined previously as protected value on Gitlab
+export dataservice_read_test_appid="${dataservice_read_test_appid}" #defined previously as protected value on Gitlab
+export dataservice_read_test_catalog="hrn:here:data:::hereos-internal-test-v2"
 export dataservice_write_test_secret="${dataservice_write_test_secret}" #defined previously as protected value on Gitlab
 export dataservice_write_test_appid="${dataservice_write_test_appid}" #defined previously as protected value on Gitlab
 export dataservice_write_test_catalog="hrn:here:data:::olp-cpp-sdk-ingestion-test-catalog"
--- a/scripts/linux/fv/olp-dataservice-read-test.variables
+++ b/scripts/linux/fv/olp-dataservice-read-test.variables
@@ -1,3 +1,3 @@
-export secret="${dataservice_read_test_secret}" #defined previously as protected value on Gitlab
-export appid="${dataservice_read_test_appid}" #defined previously as protected value on Gitlab
-export catalog="hrn:here:data:::hereos-internal-test-v2"
+export dataservice_read_test_secret="${dataservice_read_test_secret}" #defined previously as protected value on Gitlab
+export dataservice_read_test_appid="${dataservice_read_test_appid}" #defined previously as protected value on Gitlab
+export dataservice_read_test_catalog="hrn:here:data:::hereos-internal-test-v2"
--- a/tests/integration/CMakeLists.txt
+++ b/tests/integration/CMakeLists.txt
@@ -17,14 +17,13 @@
 
 cmake_minimum_required(VERSION 3.5)
 
-set(OLP_SDK_FUNCTIONAL_TESTS_SOURCES
-    ./olp-cpp-sdk-core/NetworkTest.cpp
-    ./olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
+set(OLP_SDK_INTEGRATIONAL_TESTS_SOURCES
+    ./placeholder.cpp
 )
 
 if (ANDROID OR IOS)
-    add_library(olp-cpp-sdk-functional-tests-lib ${OLP_SDK_FUNCTIONAL_TESTS_SOURCES})
-    target_link_libraries(olp-cpp-sdk-functional-tests-lib
+    add_library(olp-cpp-sdk-integration-tests-lib ${OLP_SDK_INTEGRATIONAL_TESTS_SOURCES})
+    target_link_libraries(olp-cpp-sdk-integration-tests-lib
         PRIVATE
             gtest
             olp-cpp-sdk-core
@@ -34,24 +33,24 @@ if (ANDROID OR IOS)
     )
     if (ANDROID)
         include(${CMAKE_SOURCE_DIR}/cmake/android/gen_android_test.cmake)
-        gen_android_test_runner(olp-cpp-sdk-functional-tests
-            olp-cpp-sdk-functional-tests-lib)
+        gen_android_test_runner(olp-cpp-sdk-integration-tests
+            olp-cpp-sdk-integration-tests-lib)
         add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/android
             ${CMAKE_CURRENT_BINARY_DIR}/android)
 
     else()
         include(${CMAKE_SOURCE_DIR}/cmake/ios/gen_ios_test.cmake)
-        gen_ios_test_runner(olp-cpp-sdk-functional-tests olp-cpp-sdk-functional-tests-lib)
+        gen_ios_test_runner(olp-cpp-sdk-integration-tests olp-cpp-sdk-integration-tests-lib)
         add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/ios
             ${CMAKE_CURRENT_BINARY_DIR}/ios)
 
     endif()
 
 else()
-    add_executable(olp-cpp-sdk-functional-tests ${OLP_SDK_FUNCTIONAL_TESTS_SOURCES})
-    target_link_libraries(olp-cpp-sdk-functional-tests
+    add_executable(olp-cpp-sdk-integration-tests ${OLP_SDK_INTEGRATIONAL_TESTS_SOURCES})
+    target_link_libraries(olp-cpp-sdk-integration-tests
         PRIVATE
-            gtest_main
+            gmock_main
             olp-cpp-sdk-core
             olp-cpp-sdk-authentication
             olp-cpp-sdk-dataservice-read
--- a/tests/integration/placeholder.cpp
+++ b/tests/integration/placeholder.cpp
@@ -1,0 +1,18 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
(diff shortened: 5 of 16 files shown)


CMakeLists.txt
```diff
@@ -90,3 +90,7 @@ endif()
 # Add uninstall script
 add_custom_target(uninstall "${CMAKE_COMMAND}" -P "${CMAKE_MODULE_PATH}/uninstall.cmake")
 
+if(EDGE_SDK_ENABLE_TESTING)
+    # Add tests
+    add_subdirectory(tests)
+endif(EDGE_SDK_ENABLE_TESTING)
```
andescu:
```suggestion
endif()
```
To keep consistent

scripts/linux/fv/olp-cpp-sdk-integration-test.variables
```diff
@@ -0,0 +1,6 @@
+export dataservice_read_test_secret="${dataservice_read_test_secret}" #defined previously as protected value on Gitlab
```
andescu: exporting an environment variable from itself? Does this even work?

scripts/linux/fv/olp-dataservice-read-test.variables
```diff
@@ -1,3 +1,3 @@
-export secret="${dataservice_read_test_secret}" #defined previously as protected value on Gitlab
-export appid="${dataservice_read_test_appid}" #defined previously as protected value on Gitlab
-export catalog="hrn:here:data:::hereos-internal-test-v2"
\ No newline at end of file
+export dataservice_read_test_secret="${dataservice_read_test_secret}" #defined previously as protected value on Gitlab
```
andescu: same here

scripts/linux/fv/olp-dataservice-write-test.variables
```diff
@@ -1,3 +1,3 @@
-export secret="${dataservice_write_test_secret}" #defined previously as protected value on Gitlab
-export appid="${dataservice_write_test_appid}" #defined previously as protected value on Gitlab
-export catalog="hrn:here:data:::olp-cpp-sdk-ingestion-test-catalog"
\ No newline at end of file
+export dataservice_write_test_secret="${dataservice_write_test_secret}" #defined previously as protected value on Gitlab
```
andescu: probably also here

tests/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+cmake_minimum_required(VERSION 3.5)
+
+add_subdirectory(integration)
```
andescu: Do we need this file? Can we not just include directly the subfolders?

tests/integration/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+cmake_minimum_required(VERSION 3.5)
+
+SET(OLP_CPP_SDK_INTEGRATION_TESTS_SOURCES
```
andescu:
```suggestion
SET(OLP_SDK_INTEGRATION_TESTS_SOURCES
```

tests/integration/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+cmake_minimum_required(VERSION 3.5)
```
andescu: probably this should be enough in root CMakeLists.txt

tests/integration/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
+)
+
+add_executable(olp-cpp-sdk-integration-tests
+    ${OLP_CPP_SDK_INTEGRATION_TESTS_SOURCES}
+)
+
+target_link_libraries(olp-cpp-sdk-integration-tests
+PRIVATE
+    gtest
```
andescu: gtest_main already includes gtest. You only need gtest_main.
You are not linking twice to the same lib.

But i am pretty sure that online tests are not integration tests. Integration tests are e.g. CatalogClientTest with mocked Network. Online tests are functional tests. So probably you also need gmock lib

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
```diff
@@ -0,0 +1,78 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include <gtest/gtest.h>
+
```
andescu: remove extra newline

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include <future>
+
+#include <olp/core/client/OlpClient.h>
+#include <olp/core/client/OlpClientFactory.h>
+#include <olp/core/client/OlpClientSettingsFactory.h>
+
```
andescu: remove extra newline

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
```diff
@@ -0,0 +1,78 @@
...
+class OlpClientDefaultAsyncHttp : public ::testing::Test {
+ protected:
+  olp::client::OlpClientSettings client_settings_;
+  olp::client::OlpClient client_;
+};
+
+TEST_F(OlpClientDefaultAsyncHttp, GetGoogleWebsite) {
+  client_.SetBaseUrl("https://www.google.com");
+
+  client_settings_.network_request_handler = olp::client::
```
andescu: same mistake as before?
Common parts move into SetUp/TearDown

scripts/linux/fv/olp-cpp-sdk-integration-test.variables
ystefinko: This is approach we use on FV to exclude hardcoded tokens from scripts. Those variables are setup on Gitlab Project Settings Page.

serhiikobalt: Yes, it works.  I want all valiables to be in the same place as it was before. Having 2 different names for variable that represent one entity is more confusing than this.

tests/integration/CMakeLists.txt
serhiikobalt: every other cmake list explicitly states required minimum version of cmake, why this should be different?

scripts/linux/fv/olp-cpp-sdk-integration-test.variables
andescu: To be sure we understand each other. You are exporting the same environment variable again as string. That was the reason why it had a different name. It is not very nice that you override env variables passed down from above.

tests/integration/CMakeLists.txt
andescu: Might be a copy paste error.
If you only go trough root CMakeLists.txt then you dont need it. And currently we only go trough root CMakeLists.txt

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
serhiikobalt: Tests refactoring is another huge topic and not in scope of OLPEDGE-717.

tests/integration/CMakeLists.txt
serhiikobalt: I'd say consistency is more important that avoiding style that does no harm.

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
andescu: But this is not an integration test. This is a functional test, resp. online test.
Also not adding the network to as static will crash again. Means separate, wait for crash, fix crash is the workflow?

scripts/linux/fv/olp-cpp-sdk-integration-test.variables
serhiikobalt: >  exporting the same environment variable again as string.
I'm not sure what you mean. Environment variables are strings, as well as any other shell variable.

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
serhiikobalt: How moving client settings initialization to SetUp will prevent network from being static?  

Yes, wait for crash -> fix crash is the workflow.

scripts/linux/fv/olp-cpp-sdk-integration-test.variables
andescu: Ok, let me talk trough code
```bash
# Envirnoment variable set by GitLab CI
export dataservice_read_test_secret = "token_sercret_1"

# Your changes in scripts/linux/fv/olp-cpp-sdk-integration-test.variables
# which export the same environment variable again.
export dataservice_read_test_secret = {$dataservice_read_test_secret}
```
Why?

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
andescu: Two different topics:
1. Network static as in other online tests to avoid crash. Why not also here?
2. As you create a new fixture, move common parts, e.g. client_.SetSettings(client_settings_) to SetUp().

And btw. why is client_settings_ a member?

scripts/linux/fv/olp-cpp-sdk-integration-test.variables
serhiikobalt: Because this way all variables are visible to the shell in one place and in case it will ever change no need to look for them across whole codebase on multiple CI environments.

andescu: Seems more like 3 places, 3 scripts, not one. And it is only one CI environment, gitlab. A comment would suffice instead of reexporting them again

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
serhiikobalt: Good question, idk. But it didn't fail once so far. 
I don't create new fixture is was there before. 
Someboby made client_settings_ a member long time ago.

OLPEDGE-717 has a clear scope. I also don't like what this test look like but fixing tests that are not even failing is clearly not in that scope. You are welcome to create another ticket to refactor the tests.

scripts/linux/fv/olp-cpp-sdk-integration-test.variables
serhiikobalt: For every test there is one .variables file that explicitly lists all required variables. These files are visible to both GitLab and Travis CI. ```export``` is as cheap as variable assignment. Why do you think it is better to remove explicit code in favor of comments that people have tendency to not read?

tests/integration/olp-cpp-sdk-core/OlpClientDefaultAsyncHttp.cpp
andescu: Seems tickets scope is very subjective, especially when I see changes here that are also not part of the scope.

scripts/linux/fv/olp-cpp-sdk-integration-test.variables
andescu: Ok, lets park the discussion for now.
I would propose in the future to focus on the ticket itself instead of refactore every piece of code on the way, as this is an endless loop.

tests/integration/placeholder.cpp
```diff
@@ -0,0 +1,18 @@
+/*
```
andescu: Remove, we dont need placeholders.

andescu: Only thing blocking the merge is the removal of this placeholder empty file.

serhiikobalt: I couldn't have integration tests binary without any source file.

andescu: But then I would say better is to postpone the creation of the integration test binary instead of having shady dummy files, right? Anyway, to not keep this endless loop going. Please avoid adding dummy files in the future!

PR #191, opened 2019-09-23
Add performance test for CatalogClient

serhiikobalt:
* ReadNPartitions parameterized test.
 * Common mock class for KeyValueCache.

Relates-to: OLPEDGE-745

### DIFF
--- a/tests/performance/CMakeLists.txt
+++ b/tests/performance/CMakeLists.txt
@@ -1,0 +1,30 @@
+# Copyright (C) 2019 HERE Europe B.V.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+cmake_minimum_required(VERSION 3.5)
+
+set(OLP_SDK_PERFORMANCE_TESTS_SOURCES
+    ./CatalogClientTest.cpp
+)
+
+add_executable(olp-cpp-sdk-performance-tests ${OLP_SDK_PERFORMANCE_TESTS_SOURCES})
+target_link_libraries(olp-cpp-sdk-performance-tests
+    PRIVATE
+        gtest_main
+        olp-cpp-sdk-authentication
+        olp-cpp-sdk-dataservice-read
+)
--- a/tests/performance/CatalogClientTest.cpp
+++ b/tests/performance/CatalogClientTest.cpp
@@ -1,0 +1,177 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <memory>
+#include <mutex>
+
+#include <olp/authentication/TokenProvider.h>
+
+#include <olp/core/client/HRN.h>
+#include <olp/core/client/OlpClientSettings.h>
+#include <olp/core/client/OlpClientSettingsFactory.h>
+
+#include <olp/dataservice/read/CatalogClient.h>
+#include <olp/dataservice/read/CatalogRequest.h>
+#include <olp/dataservice/read/DataRequest.h>
+#include <olp/dataservice/read/PartitionsRequest.h>
+
+namespace {
+struct CatalogClientTestConfiguration {
+  int requests_per_second;
+  int calling_thread_count;
+  int task_scheduler_capacity;
+  int runtime;
+};
+
+std::ostream& operator<<(std::ostream& stream,
+                         const CatalogClientTestConfiguration& config) {
+  return stream << "CatalogClientTestConfiguration("
+                << ".calling_thread_count=" << config.calling_thread_count
+                << ", .task_scheduler_capacity="
+                << config.task_scheduler_capacity
+                << ", .requests_per_second=" << config.requests_per_second
+                << ", .runtime=" << config.runtime << ")";
+}
+
+const std::string kKeyId("123");
+const std::string kKeySecret("456");
+const std::string kLayerId("versioned_test_layer");
+const std::string kPartitionId("123456789");
+}  // namespace
+
+class CatalogClientTest
+    : public ::testing::TestWithParam<CatalogClientTestConfiguration> {
+ public:
+  static void SetUpTestSuite() {
+    s_network = olp::client::OlpClientSettingsFactory::
+        CreateDefaultNetworkRequestHandler();
+  }
+  static void TearDownTestSuite() { s_network.reset(); }
+
+  olp::http::NetworkProxySettings GetLocalhostProxySettings() {
+    return olp::http::NetworkProxySettings()
+        .WithHostname("http://localhost:3000")
+        .WithUsername("test_user")
+        .WithPassword("test_password")
+        .WithType(olp::http::NetworkProxySettings::Type::HTTP);
+  }
+
+  std::shared_ptr<olp::client::OlpClientSettings>
+  CreateCatalogClientSettings() {
+    auto parameter = GetParam();
+
+    std::shared_ptr<olp::thread::TaskScheduler> task_scheduler = nullptr;
+    if (parameter.task_scheduler_capacity != 0) {
+      task_scheduler =
+          olp::client::OlpClientSettingsFactory::CreateDefaultTaskScheduler(
+              parameter.task_scheduler_capacity);
+    }
+    olp::authentication::Settings settings;
+    settings.task_scheduler = task_scheduler;
+    settings.network_request_handler = s_network;
+
+    olp::client::AuthenticationSettings auth_settings;
+    auth_settings.provider = []() { return "invalid"; };
+
+    auto client_settings = std::make_shared<olp::client::OlpClientSettings>();
+    client_settings->authentication_settings = auth_settings;
+    client_settings->task_scheduler = task_scheduler;
+    client_settings->network_request_handler = s_network;
+    client_settings->proxy_settings = GetLocalhostProxySettings();
+
+    return client_settings;
+  }
+
+ protected:
+  static std::shared_ptr<olp::http::Network> s_network;
+};
+
+std::shared_ptr<olp::http::Network> CatalogClientTest::s_network;
+
+namespace {
+
+void TestClientThread(
+    std::shared_ptr<olp::dataservice::read::CatalogClient> service_client,
+    std::chrono::duration<double, std::milli> sleep_interval, int runtime) {
+  auto start = std::chrono::steady_clock::now();
+  int partition_id = 0;
+
+  while (true) {
+    partition_id++;
+
+    auto request = olp::dataservice::read::DataRequest()
+                       .WithLayerId(kLayerId)
+                       .WithPartitionId(std::to_string(partition_id))
+                       .WithBillingTag(boost::none);
+
+    service_client->GetData(
+        request, [&](olp::dataservice::read::DataResponse response) {
+          if (!response.IsSuccessful()) {
+            auto error_string = response.GetError().GetMessage();
+            std::cout << "Error: " << error_string << "\n";
+          }
+        });
+
+    std::this_thread::sleep_for(sleep_interval);
+
+    auto end = std::chrono::steady_clock::now();
+    std::chrono::duration<double, std::milli> elapsed = end - start;
+    if (elapsed.count() > runtime * 1000) {
+      return;
+    }
+  }
+}
+}  // namespace
+
+TEST_P(CatalogClientTest, ReadPartitions) {
+  auto parameter = GetParam();
+  std::chrono::duration<double, std::milli> sleep_interval(
+      1000. / parameter.requests_per_second);
+
+  auto olp_client_settings = CreateCatalogClientSettings();
+  olp::client::HRN hrn("hrn:here:data:::testhrn");
+
+  std::vector<std::thread> client_threads;
+
+  for (int client_id = 0; client_id < parameter.calling_thread_count;
+       client_id++) {
+    auto service_client =
+        std::make_shared<olp::dataservice::read::CatalogClient>(
+            hrn, olp_client_settings);
+    auto thread = std::thread(TestClientThread, std::move(service_client),
+                              sleep_interval, parameter.runtime);
+    client_threads.push_back(std::move(thread));
+  }
+
+  for (auto& thread : client_threads) {
+    thread.join();
+  }
+}
+
+//   int requests_per_second;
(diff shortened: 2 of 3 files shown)


tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  for (auto& thread : client_threads) {
+    thread.join();
+  }
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    Performance, CatalogClientTest,
+    ::testing::Values(CatalogClientTestConfiguration{10, 1, 1, true},
```
borzun: What about synchronous execution (i.e. without task scheduler)? Do we need to cover that as well? 
@mykhailo-kuchma

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+ protected:
+  static std::shared_ptr<olp::http::Network> s_network;
+};
+
+std::shared_ptr<olp::http::Network> CatalogClientTest::s_network;
+
+TEST_P(CatalogClientTest, ReadNPartitions) {
+  auto parameter = GetParam();
```
borzun:
```suggestion
  auto config = GetParam();
```
same here

serhiikobalt:
@leonid-ziskel memory profiler will show heap usage by in-memory cache if it's not disabled. I assume we might want to measure memory consumption without cache part. 
`10` is just an arbitrary number, I'll pick better one when there will be results for local server. 

@borzun I'd rather stuck to `parameter` name as it is parameterized test and function name is `GetParam()`

tests/common/mocks/KeyValueCache.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <gmock/gmock.h>
+
+#include <olp/core/cache/KeyValueCache.h>
+
+namespace olp {
+namespace testing {
+namespace mocks {
+
+class KeyValueCache : public olp::cache::KeyValueCache {
```
andescu:
```suggestion
class MockKeyValueCache : public olp::cache::KeyValueCache {
```
But the more urging question that is to be solved before: Why are we testing real behavior with mocks?

tests/performance/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+cmake_minimum_required(VERSION 3.5)
+
+set(OLP_SDK_PERFORMANCE_TESTS_SOURCES
+    ./PerformanceTest.cpp
```
andescu: PerformanceTest? Means we will add all performance tests to one file only in the future?
Change please.

tests/performance/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+)
+
+if (ANDROID OR IOS)
```
andescu: Why do we need android and ios? remove.

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <olp/dataservice/read/CatalogClient.h>
+#include <olp/dataservice/read/CatalogRequest.h>
+#include <olp/dataservice/read/DataRequest.h>
+#include <olp/dataservice/read/PartitionsRequest.h>
+
+#include <mocks/KeyValueCache.h>
+
+using namespace olp;
+
+struct CatalogClientTestConfiguration {
```
andescu: anonymous namespace

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,179 @@
...
+  int number_of_requests;
+  int calling_thread_count;
+  int task_scheduler_capacity;
+  bool use_no_cache;
+};
+
+std::ostream& operator<<(std::ostream& stream,
```
andescu: anonymous namespace

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+class CatalogClientTest
+    : public ::testing::TestWithParam<CatalogClientTestConfiguration> {
+ public:
+  static void SetUpTestSuite() {
+    s_network =
+        client::OlpClientSettingsFactory::CreateDefaultNetworkRequestHandler();
+  }
+  static void TearDownTestSuite() {
+    OLP_SDK_LOG_INFO_F("CatalogClientTest", "Network use count is %ld",
```
andescu: Remove logs

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,179 @@
...
+                       s_network.use_count());
+    s_network = nullptr;
```
andescu: s_network.reset();

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,179 @@
...
+    client_settings->network_request_handler = s_network;
+    client_settings->proxy_settings = GetLocalhostProxySettings();
+
+    std::shared_ptr<olp::cache::KeyValueCache> cache = CreateCache();
+
+    return std::make_shared<dataservice::read::CatalogClient>(
+        hrn, client_settings, cache);
+  }
+
+  std::shared_ptr<olp::cache::KeyValueCache> CreateCache() {
```
andescu: This is definitely not a real behavior testing strategy.
Mocked elements in a performance test is a very new experience for me, never seen this before.
Our key concept in the SDK is that we have cache to speed things up and basically everyone is using it, so why do you want to remove it?

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,179 @@
...
+std::shared_ptr<olp::http::Network> CatalogClientTest::s_network;
+
+TEST_P(CatalogClientTest, ReadNPartitions) {
+  auto parameter = GetParam();
+
+  auto service_client = CreateCatalogClient();
+
+  std::vector<std::thread> client_threads;
+
+  for (int client_id = 0; client_id < parameter.calling_thread_count;
```
andescu: where is client_id used? If it is not used then use a while loop with decrement

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,179 @@
...
+    auto thread = std::thread([&]() {
+      for (int iteration = 0; iteration < parameter.number_of_requests;
+           iteration++) {
+        auto request = olp::dataservice::read::DataRequest()
+                           .WithLayerId(kLayerId)
+                           .WithPartitionId(kPartitionId)
+                           .WithBillingTag(boost::none);
+
+        auto future = service_client->GetData(request);
+        olp::dataservice::read::DataResponse data_response =
```
andescu: what happens when 5 request succeed and 995 return kSlowDown?
Is this still a real performance test?
Please check error and react accordingly.

tests/performance/PerformanceTest.cpp
```diff
@@ -0,0 +1,179 @@
...
+                      CatalogClientTestConfiguration{10, 1, 10, true},
+                      CatalogClientTestConfiguration{10, 10, 1, true},
+                      CatalogClientTestConfiguration{10, 10, 10, true},
+
+                      CatalogClientTestConfiguration{10, 1, 1, false},
+                      CatalogClientTestConfiguration{10, 1, 10, false},
+                      CatalogClientTestConfiguration{10, 10, 1, false},
+                      CatalogClientTestConfiguration{10, 10, 10, false},
+
+                      CatalogClientTestConfiguration{5000, 1, 1, false}));
```
andescu: 5000 requests, one thread, all wait for each other. This is a pipeline not a performance test.

tests/common/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+cmake_minimum_required(VERSION 3.5)
+
+add_library(olp-cpp-sdk-tests-common-lib INTERFACE)
```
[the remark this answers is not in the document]
borzun: Actually, adding -lib suffix might be handy - it is easier to switch between targets in IDE - you better understand where is the executable and where is the library.

tests/common/mocks/KeyValueCache.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+#pragma once
+
+#include <gmock/gmock.h>
+
+#include <olp/core/cache/KeyValueCache.h>
+
+namespace olp {
+namespace testing {
```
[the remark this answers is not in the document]
serhiikobalt: that's why it is in olp namespace :-)

tests/common/CMakeLists.txt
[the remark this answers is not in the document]
serhiikobalt: I recall a strong desire to avoid placeholder files and now this library has nothing that could be compiled in it so I've made it a header only. 

Functional and integration tests use -lib suffix in target names and that's ok.

tests/common/mocks/KeyValueCache.h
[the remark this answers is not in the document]
andescu: 1. You don't put tests in the same namespace as the classes you are testing.
2. You don't use existing namespaces.

tests/common/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+target_link_libraries(olp-cpp-sdk-tests-common-lib
+        INTERFACE
+                gmock
+                olp-cpp-sdk-core
+)
+
+target_include_directories(olp-cpp-sdk-tests-common-lib
+        INTERFACE
+                .
```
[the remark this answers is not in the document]
serhiikobalt: We don't use variables for list of targets include directories everywhere else. I believe hiding list of directories as well as list of source files behind some mutable variable is a bad practice because you newer realy know what is in there by just looking at it.

tests/common/CMakeLists.txt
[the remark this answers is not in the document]
andescu: This common lib will disappear anyway as the cache mock will disappear.
And even if you want to keep it, by some strange reason, you should, as recommended by gmock and also mentioned in the confluence page, put the destructor and constructor of the mock in a cpp file.

@borzun So you want to tell me that instead of using your IDE to get the build targets available, you rather add an lib at the end of an lib? libolp-cpp-sdk-tests-common-lib just to know that it is a lib in IDE? In a public repository?

tests/common/CMakeLists.txt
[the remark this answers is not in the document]
andescu: So using ${CMAKE_CURRENT_SOURCE_DIR} instead of `.` is a bad practice?

tests/common/mocks/KeyValueCache.h
serhiikobalt: `MockKeyValueCache `? Not `KeyValueCacheMock`?

serhiikobalt: People may be interesting in measuring memory consumption with in-memory cache disabled.

andescu: Which people? And gmock works without memory usage?
If you want to test without cache you set cache to nullptr but as this is not possible currently inside the OLP SDK you want to hide the unavoidable memory usage of the in-memory cache? I didn't recollect that we want to test OLP SDK without cache.

tests/performance/CMakeLists.txt
mykhailo-kuchma: Why not? ios and android devices are real hardware and memory/cpu data very representative.

tests/performance/CatalogClientMemoryConsumptionTest.cpp
```diff
@@ -0,0 +1,178 @@
...
+  olp::http::NetworkProxySettings GetLocalhostProxySettings() {
+    return olp::http::NetworkProxySettings()
+        .WithHostname("http://localhost:3000/http_proxy")
+        .WithUsername("test_user")
+        .WithPassword("test_password")
+        .WithType(olp::http::NetworkProxySettings::Type::HTTP);
+  }
+
+  std::shared_ptr<olp::dataservice::read::CatalogClient> CreateCatalogClient() {
+    olp::client::HRN hrn("fake");
```
mykhailo-kuchma: not used

tests/performance/CMakeLists.txt
andescu: As we do not execute them yet on CI. If we will integrate iOS & Android tests in FV/NV on CI we can add the additional cmake entries pretty quick.

tests/common/CMakeLists.txt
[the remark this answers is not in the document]
serhiikobalt: Usually in places where we use variables for e.g. sources we don't use `${CMAKE_CURRENT_SOURCE_DIR}`, instead there is some kind of `${OLP_SDK_SOMETHING_SOMETHING_SOURCES}`. I thought about them first.

I don't see any difference between `${CMAKE_CURRENT_SOURCE_DIR}` and `.` and don't have any preference for one over another.

tests/performance/CMakeLists.txt
mykhailo-kuchma: We don't execute anything on iOS/Android PSV/NV/etc. But our cmake offers you to generate test app for that. At least now we can do that manually.

tests/common/mocks/KeyValueCache.h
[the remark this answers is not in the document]
serhiikobalt: `olp::testing::mocks` is not the same as `olp::cache`. Also I haven't seen 
> don't use `olp` namespace for tests

anywhere in code convention. Is there a good reason for that? 

I don't extend existing namespaces anywhere in tests/mocks/wherever

tests/performance/CMakeLists.txt
andescu: At least to me as a user of the SDK tests are not very interesting.
Did you ever execute the tests from boost, leveldb, rapidjson or any other public repo that you used?

tests/common/mocks/KeyValueCache.h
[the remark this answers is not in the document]
andescu: You extend the olp namespace.
I think the main reason is the namespace pollution, and if this is not enough it is the mear reason for the namespace itself. Using `olp` means you add the "feature" of mocked classes to the olp scope.

tests/common/CMakeLists.txt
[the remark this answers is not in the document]
borzun: @andescu 
What I wanted to say is the pros of having such prefix. And keep in mind, that this approach with `-lib` suffix, is currently used in our repo.

tests/performance/CMakeLists.txt
andescu: Yes, we should. But not part of this ticket.

tests/common/mocks/KeyValueCache.h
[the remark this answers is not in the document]
serhiikobalt: ok, I was wrong. I meant to say I don't extend "leaf" namespaces. How I see the alternative:

I put tests into `::test`, because `::testing` is not good enought and is taken by Google. The other day somebody I've never met will use SDK as subproject, write in CMakeLists.txt `add_subdirectory(here_olp_sdk)` and write their own tests and put them into `::test`. Why? Because `::testing` is already taken by Google and nobody else will ever put anything in `::test` namespace, right?

On the other side, `::olp` is our safe bay which we have full control over and we can agree on `::olp::testing` for tests.

leonid-ziskel:
> @leonid-ziskel memory profiler will show heap usage by in-memory cache if it's not disabled. I assume we might want to measure memory consumption without cache part.

We have a setting to disable the in-memory cache. You don't need to mock it. Furthermore, comparing with the data we download I guess the neglate the application itself.

tests/performance/PerformanceTest.cpp
borzun: Agreed to include the sync. execution to stress tests

tests/performance/CatalogClientTest.cpp
```diff
@@ -0,0 +1,170 @@
...
+                              sleep_interval, parameter.runtime);
+    client_threads.push_back(std::move(thread));
+  }
+
+  for (auto& thread : client_threads) {
+    thread.join();
+  }
+}
+
+//   int requests_per_second;
```
borzun: Better to place these comments to the left of actual values (see lines 169 - 170) in the format (sorry I didn't run clang-format on this code :) ):

```
::testing::Values(CatalogClientTestConfiguration{1000, /* requests_per_second */
10,      /* calling_thread_count */
20,     /* task_scheduler_capacity  */
3600  /* runtime */},
```

PR #209, opened 2019-10-02
Common tests mocks refactoring

sergii-vostrikov:
Network and Cache mocks moved to tests/common
IsGetRequest matcher moved from inline definition to a dedicated place

Relates-to: OLPEDGE-768


sergii-vostrikov:
implementation of test projects to be implemented as separate pull-requests

PR #220, opened 2019-10-03
Copy ArcGis functional tests to <root>/tests

borzun:
Copied the ArcGisAuthenticationOnlineTest from olp-sdk-authentication/tests to <root>/tests/functional folder with following changes:
 - renamed ArcGisAuthenticationOnlineTest to ArcGisAuthenticationTest;
 - moved ArcGisAuthenticationTest class fixture to cpp file;

Relates to: OLPEDGE-718

### DIFF
--- a/tests/functional/olp-cpp-sdk-authentication/ArcGisAuthenticationTest.cpp
+++ b/tests/functional/olp-cpp-sdk-authentication/ArcGisAuthenticationTest.cpp
@@ -1,0 +1,153 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


tests/functional/olp-cpp-sdk-authentication/ArcGisAuthenticationTest.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include "AuthenticationCommonTestFixture.h"
+#include "ArcGisTestUtils.h"
+#include "TestConstants.h"
+
+namespace {
+constexpr auto kErrorArcGisFailedCode = 400300;
+
+const std::string kErrorArcGisFailedMessage = "Invalid token.";
+
+}  // namespace
```
serhiikobalt: you could place everything inside anonymous namespace and make it a bit easier for linker

borzun: Ok, will do in a separate changes after I done with moving of all original tests

PR #244, opened 2019-10-07
ApiClientLookup cancel-before-execution case

sergii-vostrikov:
Added the condition's unblock on cancel event to cover a case when user
cancelled the operation before it actually was set up.

Relates-to: OLPEDGE-832


olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
```diff
@@ -164,24 +164,31 @@ ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApi(
...
-        }
-      };
-
-  cancellation_context.ExecuteOrCancelled([&, flag]() {
-    auto token = ApiClientLookup::LookupApiClient(
-        client, service, service_version, catalog, std::move(apiClientCallback));
-    return client::CancellationToken([&, token, flag]() {
-      if (flag->exchange(false)) {
-        token.cancel();
+  auto apiClientCallback = [&,
```
borzun:
```suggestion
  auto api_client_callback = [&,
```

PR #256, opened 2019-10-08
Add performance test for CatalogClient GetData.

mykhailo-kuchma:
The test performs N request per second for M seconds.
Valgrind, heaptrack, other tools are used to collect the output.
Local OLP server required.

Relates-to: OLPEDGE-745

### DIFF
--- a/tests/performance/CatalogClientTest.cpp
+++ b/tests/performance/CatalogClientTest.cpp
@@ -1,0 +1,213 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include <chrono>
+#include <memory>
+
+#include <gtest/gtest.h>
+
+#include "olp/authentication/TokenProvider.h"
+#include "olp/core/client/HRN.h"
+#include "olp/core/client/OlpClientSettings.h"
+#include "olp/core/client/OlpClientSettingsFactory.h"
+#include "olp/core/logging/Log.h"
+#include "olp/dataservice/read/CatalogClient.h"
+#include "olp/dataservice/read/DataRequest.h"
+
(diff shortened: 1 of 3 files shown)


tests/performance/CatalogClientTest.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <olp/core/client/HRN.h>
+#include <olp/core/client/OlpClientSettings.h>
+#include <olp/core/client/OlpClientSettingsFactory.h>
+
+#include <olp/core/logging/Log.h>
+
+#include <olp/dataservice/read/CatalogClient.h>
+#include <olp/dataservice/read/CatalogRequest.h>
+#include <olp/dataservice/read/DataRequest.h>
+#include <olp/dataservice/read/PartitionsRequest.h>
```
serhiikobalt: no need for blank lines

tests/performance/CatalogClientTest.cpp
```diff
@@ -0,0 +1,218 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include <gtest/gtest.h>
```
serhiikobalt: gtest should go after c++ headers

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

tests/performance/CatalogClientTest.cpp
```diff
@@ -0,0 +1,218 @@
...
+ * Test performs N requests per second for M duration. All requests are unique.
+ * Total requests number calculated as:
+ *
+ * total_count = requests_per_second * calling_thread_count
+ *
+ * By default, SDK initialized in-memory cache with 1 MB capacity.
+ * To run the test, you need to start a local OLP mock server first.
+ * Valgrind, heaptrack, other tools are used to collect the output.
+ */
+TEST_P(CatalogClientTest, ReadNPartitions_VersionedLayer) {
```
serhiikobalt: underscores in test names are dangerous
```suggestion
TEST_P(CatalogClientTest, ReadNPartitionsFromVersionedLayer) {
```

tests/performance/CatalogClientTest.cpp
```diff
@@ -0,0 +1,213 @@
...
+            const auto error_it = errors.find(error_code);
+            if (error_it != errors.end()) {
+              error_it->second++;
+            } else {
+              errors[error_code] = 0;
+            }
+          }
+        });
+
+    std::this_thread::sleep_for(sleep_interval);
```
diachenko-mischa: In case of sleep, we won't keep requested frequency of requests. Consider to calculate expected end time of iteration at the start of the iteration, and at the end of iteration we can wait remaining time.

mykhailo-kuchma: We don't need to be so precise.

PR #273, opened 2019-10-10
Make ReturnHttpResponse action a free function

serhiikobalt:
Instead of ReturnHttpResponse being a static member function of a
NetworkMock class make it a free function.

Relates-to: OLPEDGE-768


sergii-vostrikov:
Do we really need to clutter global namespace? Having ReturnHttpResponse as a static member of NetworkMock syntactically ties mocks of the class and its response.

PR #277, opened 2019-10-10
Pass arguments to CatalogClient by value

diachenko-mischa:
Pass HRN and settings to CatalogClient by value. Cache is passed as
OlpClientSettings. This aligns interface with read dataservice API.

Relates-to: OLPEDGE-798


tests/integration/olp-cpp-sdk-dataservice-read/CatalogClientCacheTest.cpp
```diff
@@ -60,6 +60,7 @@ class CatalogClientCacheTest : public CatalogClientTestBase {
     cache_ = std::make_shared<olp::cache::DefaultCache>(settings);
```
serhiikobalt: check if you need cache_ as a member

diachenko-mischa: In tearDown we use ::Close() method from DefaultCache, which is not present in setting_.cache

diachenko-mischa: Yes, it is used in TeadDown, cannot be removed

PR #281, opened 2019-10-11
Implemented VolatileLayerClient::GetData() method

meravingen123:
Added service parameter to DataRepository::GetBlobData() to switch between blob and volatile-blob REST API
Added Tests for VolatileLayerClientImpl class

Resolves: OLPEDGE-760

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
@@ -19,20 +19,70 @@
 
 #include "VolatileLayerClientImpl.h"
 
+#include <olp/core/client/CancellationContext.h>
+#include <olp/dataservice/read/PartitionsRequest.h>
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,20 +19,70 @@
...
+        callback(client::ApiError(client::ErrorCode::NotFound,
+                                  "Partition not found"));
+        return;
+      }
+
+      data_request.WithDataHandle(partitions.front().GetDataHandle());
+    }
+
+    auto blob_response = repository::DataRepository::GetBlobData(
+        catalog_, layer_id_, "volatile-blob", data_request, context,
```
sergii-vostrikov: please avoid capturing "this", use copies instead

meravingen123:
Closing because it will be split into 2 parts.

PR #284, opened 2019-10-11
Added service parameter to DataRepository::GetBlobData()

meravingen123:
Added service parameter to DataRepository::GetDataBlob() to switch between blob and volatile-blob REST API
Adjusted tests to use blob service

Relates-to: OLPEDGE-760

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
@@ -328,9 +328,9 @@ void DataRepository::GetData(
                   *cache_);
 }
 
-BlobApi::DataResponse DataRepository::GetBlobData(
+DataRepository::DataResponse DataRepository::GetBlobData(
     const client::HRN& catalog, const std::string& layer,
-    const DataRequest& data_request,
+    const std::string& service, const DataRequest& data_request,
     client::CancellationContext cancellation_context,
     client::OlpClientSettings settings) {
   using namespace client;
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
```diff
@@ -328,13 +328,15 @@ void DataRepository::GetData(
...
-BlobApi::DataResponse DataRepository::GetBlobData(
+DataRepository::DataResponse DataRepository::GetBlobData(
     const client::HRN& catalog, const std::string& layer,
-    const DataRequest& data_request,
+    const std::string& service, const DataRequest& data_request,
     client::CancellationContext cancellation_context,
     client::OlpClientSettings settings) {
   using namespace client;
 
+  std::string blob_api_service = "blob";  // "volatile-blob
```
diachenko-mischa: Unused

olp-cpp-sdk-dataservice-read/tests/DataRepositoryTest.cpp
```diff
@@ -93,7 +93,7 @@ TEST_F(DataRepositoryTest, GetBlobData) {
 
   auto response =
       olp::dataservice::read::repository::DataRepository::GetBlobData(
-          hrn, "testlayer", request, context, *settings_);
+          hrn, "testlayer", "blob", request, context, *settings_);
```
diachenko-mischa: Extract const instead of "blob"

PR #287, opened 2019-10-11
Improve FV test script

ystefinko:
Add more details in case of failure
Generate better data at the end of script
That info become more useful(with summarized data) when we receive failure emails from GitLab.

Resolves: OLPEDGE-473

### DIFF
--- a/scripts/linux/fv/gitlab_test_fv.sh
+++ b/scripts/linux/fv/gitlab_test_fv.sh
@@ -1,31 +1,57 @@
 #!/bin/bash -e
-#Running every test one by one
+# Running every test one by one
 
 export REPO_HOME=$PWD
-export FV_HOME=$REPO_HOME/scripts/linux/fv
-rm -rf reports && mkdir -p reports #for reports storage
+export FV_HOME=${REPO_HOME}/scripts/linux/fv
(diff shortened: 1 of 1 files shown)


scripts/linux/fv/gitlab_test_fv.sh
```diff
@@ -1,31 +1,57 @@
...
 
 TEST_FAILURE=0
+EXPECTED_REPORT_COUNT=6                         # expected that we generate 6 reports
 
 # Run unit tests
-$FV_HOME/gitlab-olp-cpp-sdk-authentication-test.sh || TEST_FAILURE=1
-$FV_HOME/gitlab-olp-cpp-sdk-core-test.sh || TEST_FAILURE=1
-$FV_HOME/gitlab-olp-dataservice-read-test.sh || TEST_FAILURE=1
-$FV_HOME/gitlab-olp-dataservice-write-test.sh || TEST_FAILURE=1
+${FV_HOME}/gitlab-olp-cpp-sdk-authentication-test.sh 2> errors.txt || TEST_FAILURE=1
```
diachenko-mischa:
```suggestion
${FV_HOME}/gitlab-olp-cpp-sdk-authentication-test.sh 2>> errors.txt || TEST_FAILURE=1
```
otherwise errors will be overwritten

PR #288, opened 2019-10-11
Cancel pending requests when clients are destroyed.

mykhailo-kuchma:
Cancel pending requests when Versioned/Volatile/Index clients are destroyed.

Resolves: OLPEDGE-880

### DIFF
--- a/olp-cpp-sdk-dataservice-write/include/olp/dataservice/write/VersionedLayerClient.h
+++ b/olp-cpp-sdk-dataservice-write/include/olp/dataservice/write/VersionedLayerClient.h
@@ -88,7 +88,11 @@ class DATASERVICE_WRITE_API VersionedLayerClient {
    * instance
    */
   VersionedLayerClient(client::HRN catalog, client::OlpClientSettings settings);
-  ~VersionedLayerClient();
+  
+  /**
+   * @brief Cancel all pending requests.
+   */
+  void CancellAll();
 
   /**
    * @brief Start a batch operation.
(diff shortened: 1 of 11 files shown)


olp-cpp-sdk-dataservice-write/src/IndexLayerClientImpl.h
```diff
@@ -50,6 +50,8 @@ class IndexLayerClientImpl
  public:
   IndexLayerClientImpl(client::HRN catalog, client::OlpClientSettings settings);
 
+  ~IndexLayerClientImpl();
```
borzun:
```suggestion
virtual ~IndexLayerClientImpl();
```
I suggest to mark it as virtual in case when we will add proper unit tests for those Implementation files and mock them. Same with other implementation files.

What do you think?

mykhailo-kuchma: Good idea, I'll add it.

olp-cpp-sdk-dataservice-write/include/olp/dataservice/write/VersionedLayerClient.h
```diff
@@ -88,7 +88,11 @@ class DATASERVICE_WRITE_API VersionedLayerClient {
    * instance
    */
   VersionedLayerClient(client::HRN catalog, client::OlpClientSettings settings);
-  ~VersionedLayerClient();
+  
+  /**
+   * @brief Cancel all pending requests.
+   */
+  void CancellAll();
```
andescu: Typo!
CancellAll() -> CancelAll()

olp-cpp-sdk-dataservice-write/include/olp/dataservice/write/VolatileLayerClient.h
```diff
@@ -90,7 +90,7 @@ class DATASERVICE_WRITE_API VolatileLayerClient {
   /**
    * @brief Cancel all pending requests.
    */
-  void cancellAll();
+  void CancellAll();
```
andescu: Same here -> CancelAll()

PR #290, opened 2019-10-11
Implemented VolatileLayerClient::GetData() method

meravingen123:
Added Tests for VolatileLayerClientImpl class

Resolves: OLPEDGE-760

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
@@ -19,19 +19,80 @@
 
 #include "VolatileLayerClientImpl.h"
 
+#include <olp/core/cache/DefaultCache.h>
+#include <olp/core/client/CancellationContext.h>
+#include <olp/core/client/OlpClientSettingsFactory.h>
+#include <olp/dataservice/read/PartitionsRequest.h>
+#include "repositories/DataRepository.h"
+#include "repositories/ExecuteOrSchedule.inl"
+#include "repositories/PartitionsRepository.h"
+
 namespace olp {
 namespace dataservice {
 namespace read {
 
+namespace {
+constexpr auto kLogTag = "VolatileLayerClientImpl";
+}  // namespace
+
 VolatileLayerClientImpl::VolatileLayerClientImpl(
-    olp::client::HRN catalog, std::string layer_id,
-    olp::client::OlpClientSettings client_settings) {}
+    client::HRN catalog, std::string layer_id,
+    client::OlpClientSettings settings)
+    : catalog_(std::move(catalog)),
+      layer_id_(std::move(layer_id)),
+      settings_(std::move(settings)),
+      pending_requests_(std::make_shared<PendingRequests>()) {
+  if (!settings_.cache) {
+    settings_.cache = client::OlpClientSettingsFactory::CreateDefaultCache({});
+  }
+  // to avoid capturing task scheduler inside a task, we need a copy of settings
+  // without the scheduler
+  task_scheduler_ = std::move(settings_.task_scheduler);
+}
 
-VolatileLayerClientImpl::~VolatileLayerClientImpl() = default;
+VolatileLayerClientImpl::~VolatileLayerClientImpl() {
+  pending_requests_->CancelPendingRequests();
+}
 
-olp::client::CancellationToken VolatileLayerClientImpl::GetData(
-    DataRequest data_request, Callback callback) {
-  return olp::client::CancellationToken{[]() {}};
+client::CancellationToken VolatileLayerClientImpl::GetData(
+    DataRequest request, Callback<DataResponse> callback) {
+  auto add_task = [&](DataRequest& request,
+                      Callback<DataResponse> callback) mutable {
+    client::CancellationContext context;
+    auto cancellation_token = client::CancellationToken(
+        [context]() mutable { context.CancelOperation(); });
+
+    int64_t request_key = pending_requests_->GenerateRequestPlaceholder();
+    pending_requests_->Insert(cancellation_token, request_key);
+
+    auto catalog = catalog_;
+    auto layer_id = layer_id_;
+    auto settings = settings_;
+
+    repository::ExecuteOrSchedule(task_scheduler_, [=]() mutable {
+      auto response = repository::DataRepository::GetVolatileData(
+          catalog, layer_id, request, context, settings);
+      pending_requests_->Remove(request_key);
+      if (callback) {
+        callback(std::move(response));
+      }
+    });
+    return cancellation_token;
+  };
+
+  if (request.GetFetchOption() == FetchOptions::CacheWithUpdate) {
+    auto cache_token = add_task(
+        request.WithFetchOption(FetchOptions::CacheOnly), std::move(callback));
+    auto online_token =
+        add_task(request.WithFetchOption(FetchOptions::OnlineOnly), nullptr);
+
+    return client::CancellationToken([=]() {
+      cache_token.cancel();
+      online_token.cancel();
+    });
+  }
+
+  return add_task(request, std::move(callback));
 }
 
 }  // namespace read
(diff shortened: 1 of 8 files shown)


olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,20 +19,89 @@
...
+}
+
 VolatileLayerClientImpl::VolatileLayerClientImpl(
     olp::client::HRN catalog, std::string layer_id,
-    olp::client::OlpClientSettings client_settings) {}
+    olp::client::OlpClientSettings client_settings)
+    : catalog_(std::move(catalog)),
+      layer_id_(std::move(layer_id)),
+      client_settings_(std::move(client_settings)),
+      pending_requests_(new PendingRequests()) {}
```
mykhailo-kuchma: I would suggest using make_shared

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,20 +19,89 @@
...
+    }
+
+    auto blob_response = repository::DataRepository::GetBlobData(
+        catalog_, layer_id_, "volatile-blob", data_request, context,
+        client_settings_);
+
+    pending_requests->Remove(request_key);
+    if (callback) {
+      callback(blob_response);
+    }
```
mykhailo-kuchma: We can override a call to callback and call pending_requests->Remove() one time.
Fetch option Cache with update is not handled.

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,19 +19,107 @@
...
 olp::client::CancellationToken VolatileLayerClientImpl::GetData(
     DataRequest data_request, Callback callback) {
-  return olp::client::CancellationToken{[]() {}};
+  auto catalog = catalog_;
+  auto client_settings = client_settings_;
+  auto layer_id = layer_id_;
+
+  auto get_data = [=](DataRequest data_request,
+                      olp::client::CancellationContext context,
+                      Callback callback) mutable {
```
borzun:
```suggestion
                      Callback callback) {
```
Should it be mutable?

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,19 +19,107 @@
...
+                               ? data_request.GetPartitionId().get().c_str()
+                               : "<none>");
+        if (callback) {
+          callback(client::ApiError(client::ErrorCode::NotFound,
+                                    "Partition not found"));
+        }
+        return;
+      }
+
+      data_request.WithDataHandle(partitions.front().GetDataHandle());
```
borzun: Just curious - can we have more than 1 partition with same id?

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,19 +19,107 @@
...
+    }
+
+    auto blob_response = repository::DataRepository::GetBlobData(
+        catalog_, layer_id_, "volatile-blob", data_request, context,
```
borzun:
```suggestion
        catalog, layer_id, "volatile-blob", data_request, context,
```

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,19 +19,107 @@
...
+        client_settings_);
```
borzun:
```suggestion
        client_settings);
```

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
meravingen123: This should not happen as i understand.
@mykhailo-kuchma Can you please clarify?

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,19 +19,107 @@
...
+}
 
 olp::client::CancellationToken VolatileLayerClientImpl::GetData(
     DataRequest data_request, Callback callback) {
-  return olp::client::CancellationToken{[]() {}};
+  auto catalog = catalog_;
+  auto client_settings = client_settings_;
+  auto layer_id = layer_id_;
+
+  auto get_data = [=](DataRequest data_request,
```
andescu:
```suggestion
  auto get_data = [=](DataRequest request,
```

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,19 +19,107 @@
...
+  auto add_task = [&](DataRequest& data_request, Callback callback) mutable {
+    olp::client::CancellationContext context;
+    auto cancellation_token = olp::client::CancellationToken(
+        [context]() mutable { context.CancelOperation(); });
+
+    int64_t request_key = pending_requests_->GenerateRequestPlaceholder();
+    pending_requests->Insert(cancellation_token, request_key);
+
+    repository::ExecuteOrSchedule(&client_settings, [=]() mutable {
+      get_data(data_request, context, std::move(callback));
```
andescu: A little bit to many copies of the callback.
Why do we need it? Can we not just capture it?

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,19 +19,107 @@
...
+        catalog, layer_id, "volatile-blob", data_request, context,
+        client_settings);
+
+    if (callback) {
+      callback(blob_response);
+    }
+  };
+
+  auto pending_requests = pending_requests_;
+  auto add_task = [&](DataRequest& data_request, Callback callback) mutable {
```
andescu: why mutable?
You only alter pending_requests which is a shared_ptr so you should be able to remove the mutable.

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -19,19 +19,107 @@
...
+  };
+
+  if (data_request.GetFetchOption() == FetchOptions::CacheWithUpdate) {
+    auto cache_token =
+        add_task(data_request.WithFetchOption(FetchOptions::CacheOnly),
+                 std::move(callback));
+    auto online_token = add_task(
+        data_request.WithFetchOption(FetchOptions::OnlineOnly), nullptr);
+
+    return olp::client::CancellationToken([=]() {
```
andescu: We don't need to add this in the pending_requests also?

olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
```diff
@@ -18,24 +18,168 @@
...
+      olp::client::ApiResponse<VolatileLayerClientImpl::DataResult,
+                               olp::client::ApiError>;
+  std::promise<DataResponse> promise;
+  std::future<DataResponse> future = promise.get_future();
+
+  auto token = client.GetData(
+      request, [&](DataResponse response) { promise.set_value(response); });
+
+  const auto& response = future.get();
+  ASSERT_FALSE(response.IsSuccessful());
```
mykhailo-kuchma: Please add assertion on the response code, as the user may rely on it

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
mykhailo-kuchma: There is no easy way to move capture it in lambda.

andescu: As you already copy capture the callback you actually only move the copy.
At least in most of the cases. So if you already copied it you can work with the copy directly.

PR #292, opened 2019-10-15
Fix examples compilation on psv.

mykhailo-kuchma:
Resolves: OLPEDGE-887


mykhailo-kuchma:
> So, after removing `find_package`, we assume that the user will build examples only with all other libraries and not separately, right?

Yes

PR #297, opened 2019-10-16
Added a DEPRECATED macro

mykhailo-kuchma:
Added a macro to deprecate classes/functions.

Resolves: OLPEDGE-877

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/porting/deprecated.h
+++ b/olp-cpp-sdk-core/include/olp/core/porting/deprecated.h
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/include/olp/core/porting/deprecated.h
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#if defined(__GNUC__) || defined(__clang__)
+#define DEPRECATED __attribute__((deprecated))
```
andescu:
```suggestion
#define OLP_SDK_DEPRECATED(text) __attribute__((deprecated))
```
Also add text (we dont need to use it) but then the users see a text:

```cpp
OLP_SDK_DEPRECATED("Will be removed by 11.2019, please use GetData(request, callback) instead")
ErrorCode GetData(callback, request);
```

PR #302, opened 2019-10-17
Align cache according to requirements, dataservice write.

mykhailo-kuchma:
If the user do not provide a default cache, SDK will create a default one.

Resolves: OLPEDGE-873

### DIFF
--- a/olp-cpp-sdk-dataservice-write/src/StreamLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-write/src/StreamLayerClientImpl.cpp
@@ -74,13 +74,12 @@ void ExecuteOrSchedule(const OlpClientSettings& settings,
 }
 }  // namespace
 
-StreamLayerClientImpl::StreamLayerClientImpl(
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-write/src/StreamLayerClientImpl.cpp
```diff
@@ -34,16 +35,9 @@
...
-
-// clang-format off
-#include <generated/serializer/PublishDataRequestSerializer.h>
-#include <generated/serializer/JsonSerializer.h>
-// clang-format on
-
-// clang-format off
-#include <generated/parser/PublishDataRequestParser.h>
-#include <olp/core/generated/parser/JsonParser.h>
-// clang-format on
```
andescu: This order and clang-format off is necessary to be able to compile, as one defines the to_json method and the other the serializer that uses this to_json()

diachenko-mischa: Done

PR #303, opened 2019-10-17
VersionedLayerClient class implementation

sergii-vostrikov:
The class' GetData method is introduced.
It fetches data from a layer using a partition id or data handle.
The data is fetched asynchronously if the task scheduler is set
properly. The synchronous implementation of the logic is placed to
DataRepository::GetData() as it fits better from an architectural
point of view.

Resolves: OLPEDGE-759

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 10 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
```diff
@@ -0,0 +1,69 @@
...
+
+ private:
+  olp::client::CancellationToken AddGetDataTask(DataRequest data_request,
+                                                Callback callback) const;
+
+ protected:
+  olp::client::HRN catalog_;
+  std::string layer_id_;
+  olp::client::OlpClientSettings sync_settings_;
+  olp::client::OlpClientSettings scheduler_only_settings_;
```
mykhailo-kuchma: Please store just task scheduler here, and provide a new ExecuteOrSchedule function that works with task scheduler as an input

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
```diff
@@ -0,0 +1,69 @@
...
+                                                 Callback callback) const;
+
+ private:
+  olp::client::CancellationToken AddGetDataTask(DataRequest data_request,
+                                                Callback callback) const;
+
+ protected:
+  olp::client::HRN catalog_;
+  std::string layer_id_;
+  olp::client::OlpClientSettings sync_settings_;
```
mykhailo-kuchma: please name it as just settings

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
sergii-vostrikov: I will create a separate ticket for ExecuteOrSchedule rework

sergii-vostrikov: Please add your suggestions and confirm the scope of OLPEDGE-902

PR #306, opened 2019-10-18
Added VersionedLayerClient::PrefetchTiles method

borzun:
Implemented the VersionedLayerClient::PrefetchTiles method in dataservice-read component, which is based on the CatalogClient::PrefetchTiles implementation.

Also, moved the integration and functional tests regarding the pre-fetch functionality from CatalogClientTest with following changes:
- added DataserviceReadVersionedLayerClientTest.cpp file to integration tests target;
- removed the PrefetchEmbedded integration test from tests in DataserviceReadVersionedLayerClientTest test suite (see OLPEDGE-906);
- re-enabled the PrefetchBusy test in both CatalogClientTest and DataserviceReadVersionedLayerClientTest.

Relates to: OLPEDGE-795

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/CatalogClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/CatalogClientImpl.cpp
@@ -249,7 +249,7 @@ client::CancellationToken CatalogClientImpl::PrefetchTiles(
     callback(response);
   };
 
-  auto token = prefetch_provider_->PrefetchTiles(request, callback);
+  auto token = prefetch_provider_->PrefetchTiles(request, request_callback);
 
   pending_requests_->Insert(token, request_key);
   return token;
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClient.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClient.cpp
@@ -46,6 +46,11 @@ client::CancellationToken VersionedLayerClient::GetPartitions(
                               std::move(callback));
 }
 
+olp::client::CancellationToken VersionedLayerClient::PrefetchTiles(
+    PrefetchTilesRequest request, PrefetchTilesResponseCallback callback) {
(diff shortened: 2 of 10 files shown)


olp-cpp-sdk-dataservice-read/src/CatalogClientImpl.cpp
```diff
@@ -249,7 +249,7 @@ client::CancellationToken CatalogClientImpl::PrefetchTiles(
     callback(response);
   };
 
-  auto token = prefetch_provider_->PrefetchTiles(request, callback);
+  auto token = prefetch_provider_->PrefetchTiles(request, request_callback);
```
mykhailo-kuchma: Typo?

olp-cpp-sdk-dataservice-read/src/VersionedLayerClient.cpp
```diff
@@ -40,6 +40,12 @@ olp::client::CancellationToken VersionedLayerClient::GetData(
   return impl_->GetData(std::move(data_request), std::move(callback));
 }
 
+olp::client::CancellationToken VersionedLayerClient::PrefetchTiles(
+    PrefetchTilesRequest request,
+    PrefetchTilesResponseCallback callback) {
+  return impl_->PrefetchTiles(request, callback);
```
mykhailo-kuchma: Please use move

olp-cpp-sdk-dataservice-read/src/CatalogClientImpl.cpp
borzun: Can you explain? In case we will pass only `callback` there, the pending requests will contain 'dead' request keys of already processed PrefetchTiles requests. Thus, we need to pass the `request_callback` instead.

mykhailo-kuchma: Get it now

PR #309, opened 2019-10-22
Added missing auth variable aliases

sergii-vostrikov:
Masked appid/secret variables added to common test vars script

Relates-to: OLPEDGE-907


sergii-vostrikov:
the variables already present in dedicated "read" section

PR #317, opened 2019-10-24
Align VersionedLayerClient with VolatileLayerClient

mykhailo-kuchma:
Remove olp namespace for parameters.
Align GetData implementation.
Align the order of parameters in DataRepository.

Relates-To: OLPEDGE-957

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -37,16 +37,15 @@ namespace dataservice {
 namespace read {
 
 VersionedLayerClientImpl::VersionedLayerClientImpl(
-    olp::client::HRN catalog, std::string layer_id,
-    olp::client::OlpClientSettings client_settings)
+    client::HRN catalog, std::string layer_id,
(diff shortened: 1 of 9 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -107,48 +106,48 @@ client::CancellationToken VersionedLayerClientImpl::GetPartitions(
...
-      cache_token.cancel();
-      online_token.cancel();
-    });
-  } else {
-    return AddGetDataTask(data_request, std::move(callback));
-  }
-}
-
-client::CancellationToken VersionedLayerClientImpl::AddGetDataTask(
+client::CancellationToken VersionedLayerClientImpl::GetData(
```
borzun: Why you moved the `AddGetDataTask` to the lambda? For me previous approach with separate method was better than creating the lambda - easier to write unit tests.

mykhailo-kuchma: It's easier to take FetchOptions block to another place and leave a lambda inside, in this scenario 1 method per 1 API call.

PR #319, opened 2019-10-25
Moved the auto-flush props from FlushSettings into AutoFlushSettings

borzun:
Moved the auto-flush specific properties (auto_flush_num_events, auto_flush_interval, events_per_single_flush) from FlushSettings into AutoFlushSettings with following changes:
- the return type of `events_per_single_flush` changed to int;
- the AutoFlushSettings is now part of the private API.

Also, removed all the auto-flush functional and integration tests, because the auto-flush mechanism is no longer part of public API.

Relates to: OLPEDGE-826


olp-cpp-sdk-dataservice-write/src/StreamLayerClientImpl.cpp
```diff
@@ -90,7 +90,7 @@ StreamLayerClientImpl::StreamLayerClientImpl(HRN catalog,
       cache_(settings_.cache),
       cache_mutex_(),
       flush_settings_(std::move(flush_settings)),
-      auto_flush_controller_(new AutoFlushController(flush_settings_)) {}
+      auto_flush_controller_(new AutoFlushController(AutoFlushSettings{})) {}
```
sergii-vostrikov: pass the settings?

borzun: You mean to pass `flush_settings_`? Actually, it is not needed, because auto-flush is disabled. I will remove the `auto_flush_controller` from this class in next PRs

PR #322, opened 2019-10-25
Reduce flakiness in TaskContext tests

diachenko-mischa:
Introduce sequence of operations in test on Condition instead of sleep.

Relates-To: OLPEDGE-909


mykhailo-kuchma:
please also add a log in layer clients when using it

PR #330, opened 2019-10-28
Move Types to the separate header.

mykhailo-kuchma:
Relates-To: OLPEDGE-957


mykhailo-kuchma:
> Can we not gradually also move towards template alias.
> Even if we keep the callback aliases we can still use underneath a template alias for std::function<void(client::ApiResponse<..., client::ApiError> response)>

Updated commit

PR #349, opened 2019-10-31
HttpResponse now holds data as a stream

sergii-vostrikov:
The class now hold the stream to response data rather a string.
JsonParser.h extended with stream variant of the parser.

Resolves: OLPEDGE-684

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
@@ -19,7 +19,7 @@
 
 #pragma once
 
-#include <string>
+#include <sstream>
 
 #include <olp/core/CoreApi.h>
 #include <olp/core/http/NetworkTypes.h>
@@ -35,6 +35,8 @@ class CORE_API HttpResponse {
   HttpResponse() = default;
   HttpResponse(int status, std::string response = std::string())
       : status(status), response(std::move(response)) {}
+  HttpResponse(int status, std::stringstream&& response)
+      : status(status), response(std::move(response)) {}
 
   /**
    * @brief status Status of the HTTP request.
@@ -42,7 +44,7 @@ class CORE_API HttpResponse {
    * @see olp::network::HttpStatusCode for positive status numbers
    */
   int status{static_cast<int>(olp::http::ErrorCode::UNKNOWN_ERROR)};
-  std::string response;
+  std::stringstream response;
 };
 
 }  // namespace client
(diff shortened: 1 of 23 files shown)


olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
```diff
@@ -33,7 +34,7 @@ namespace client {
 class CORE_API HttpResponse {
  public:
   HttpResponse() = default;
-  HttpResponse(int status, std::string response = std::string())
+  HttpResponse(int status, std::shared_ptr<std::stringstream> response = nullptr)
```
mykhailo-kuchma: can we remove shared_ptr and pass std::stringstream directly with std::move?
alternatively, we can also have a constructor with a std::string

sergii-vostrikov: Currently, user creates the stream and passes it to us. moving the stream would discard user's instance of it. I doubt we can afford that right now.
See usage example in NetworkMock

sergii-vostrikov: We could use moves if the stream was created by us and passed as an outcome, but since that's a user-initialized value, it's too risky to move it.

olp-cpp-sdk-dataservice-read/src/generated/api/PlatformApi.cpp
```diff
@@ -19,6 +19,9 @@
 
 #include "PlatformApi.h"
 
+#include <memory>
```
diachenko-mischa: Leftover from previous patch-set

olp-cpp-sdk-dataservice-read/src/generated/api/BlobApi.cpp
```diff
@@ -20,6 +20,8 @@
 #include "BlobApi.h"
 
 #include <map>
+#include <memory>
```
diachenko-mischa: Leftover from previous patch-set

sergii-vostrikov: nevertheless, it's required for shared_ptr used in the file.

olp-cpp-sdk-dataservice-read/src/generated/api/PlatformApi.cpp
sergii-vostrikov: nevertheless, it's required for shared_ptr used in the file.

PR #351, opened 2019-11-01
Refactored the bridge between C++ and Java using peer pattern

borzun:
- Used peer native handle pattern to refactor the communication between the C++ JNI side (`NetworkAndroid.cpp`) and Java side (`HttpClient.java`).
- Removed static container of `NetworkAndroid` objects.
- Fixed memory leak of `NetworkAndroid` by always joining the working thread.

Relates to: OLPEDGE-735


olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
```diff
@@ -289,10 +275,18 @@ bool NetworkAndroid::Initialize() {
...
+    gJniNativePtrField = env->GetFieldID(java_self_class_, "nativePtr", "J");
+    if (env->ExceptionOccurred()) {
+      OLP_SDK_LOG_ERROR(kLogTag,
+                        "initialize failed to get HttpClient.nativePtr");
+      env->ExceptionDescribe();
+      env->ExceptionClear();
+      return false;
+    }
   }
+  env->SetLongField(obj_, gJniNativePtrField, (jlong)this);
```
diachenko-mischa: Write to nativePtr is not protected here

borzun: I don't think we need to protect this code, because the `Initialize` method is already protected by a mutex and called during first `Send` method. 

So, I think, there are no potential data race, cause during `Initialize` call, the`HttpClient.java` doesn't process any requests, thus, no way it will access the `nativePtr` in other thread.

PR #352, opened 2019-11-01
Add cmake target URL & HASH hook

diachenko-mischa:
Build of external dependencies is decoupled from main build script, so
that is not possible to share variables, other that separate
configuration stage. Substitution variables are used as means to pass
correct version.

Relates-To: OLPEDGE-221

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -35,6 +35,21 @@ if(NOT N EQUAL 0)
     endif()
 endif()
 
+# Setup versions for external project 
+set(OLP_SDK_CPP_GOOGLETEST_URL "https://github.com/google/googletest.git")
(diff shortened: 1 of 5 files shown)


external/CMakeLists.txt
```diff
@@ -35,6 +35,21 @@ if(NOT N EQUAL 0)
     endif()
 endif()
 
+# Setup versions for external project 
+set(EDGE_SDK_CPP_GOOGLETEST_URL "https://github.com/google/googletest.git")
+set(EDGE_SDK_CPP_GOOGLETEST_TAG "6d668180665e313bbda19f8187abf6140e80a7d7")
```
andescu: I find it strange to have an SHA hash instead of a tag.
Please change to "release-1.10.0" or to " release-1.8.1" if it does not build with 1.10.0.

PR #359, opened 2019-11-05
Fixed Integration tests for VersionedLayerClient

meravingen123:
Now tests are using shared pointer to promise to prevent crash on timeout while waiting result.
Increased result waiting timeout to 3 seconds.

Resolves: OLPEDGE-996

### DIFF
--- a/tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerClientTest.cpp
+++ b/tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerClientTest.cpp
@@ -28,6 +28,7 @@
 #include <olp/core/client/OlpClientSettingsFactory.h>
 #include <olp/core/porting/make_unique.h>
 #include <olp/dataservice/read/VersionedLayerClient.h>
+
 #include "HttpResponses.h"
 
 using namespace olp::dataservice::read;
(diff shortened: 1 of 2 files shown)


tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerClientTest.cpp
```diff
@@ -18,16 +18,17 @@
...
 #include <matchers/NetworkUrlMatchers.h>
 #include <mocks/NetworkMock.h>
 #include <olp/authentication/Settings.h>
 #include <olp/core/client/OlpClientSettings.h>
 #include <olp/core/client/OlpClientSettingsFactory.h>
 #include <olp/core/porting/make_unique.h>
 #include <olp/dataservice/read/VersionedLayerClient.h>
+
+#include <chrono>
+#include <string>
```
borzun: I think previous layout was correct

mykhailo-kuchma:
> There are also 10 places in VersionedLayerClient integration tests, which uses `&promise` instead `shared_ptr<std::promise>`. I guess, we also need to fix them.

Disagree on that. Layer clients have a requirement to perform a callback

PR #380, opened 2019-11-11
Adapts InMemoryCache class to coding style.

andescu:
Adapts most of the internals of the InMemoryCache to the coding style.
Also removes static methods from InMemoryCache and replaces them with
nested structures to be used by the LRUCache and time provider default
parameter values. Add mutex lock for Size() method.
Promote private members to protected to be able to test via testable.
Adapts unit tests to coding style.

Relates-to: OLPEDGE-566

### DIFF
--- a/olp-cpp-sdk-core/src/cache/InMemoryCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/InMemoryCache.cpp
@@ -17,104 +17,88 @@
  * License-Filename: LICENSE
  */
 
-#include <ctime>
-
 #include "InMemoryCache.h"
 
 namespace olp {
 namespace cache {
-InMemoryCache::TimeProvider& InMemoryCache::DefaultTimeProvider() {
-  static InMemoryCache::TimeProvider defaultTimeProvider = []() {
-    return std::time(nullptr);
-  };
-  return defaultTimeProvider;
+namespace {
+inline bool HasExpiry(time_t expiry_seconds) {
+  return (expiry_seconds != InMemoryCache::kExpiryMax);
 }
+}  // namespace
 
-InMemoryCache::ModelCacheCostFunc& InMemoryCache::EqualityCacheCost() {
-  static InMemoryCache::ModelCacheCostFunc equalityCacheCost =
-      [](const ItemTuple& tuple) { return 1; };
-  return equalityCacheCost;
-}
-
-InMemoryCache::ModelCacheCostFunc& InMemoryCache::PresetModelSizeCacheCost() {
-  static InMemoryCache::ModelCacheCostFunc presetModelSizeCacheCost =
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/src/cache/InMemoryCache.cpp
```diff
@@ -123,42 +107,48 @@ void InMemoryCache::RemoveKeysWithPrefix(const std::string& keyPrefix) {
...
+      break;
+    }
   }
 
-  for (auto key : expiredKeys) ret &= PurgeExpired(key);
+  for (auto& key : expired_keys) {
+    ret &= PurgeExpired(key);
+  }
 
   return ret;
```
diachenko-mischa: Seems like `return true;` maybe used instead.

olp-cpp-sdk-core/src/cache/InMemoryCache.cpp
```diff
@@ -17,104 +17,88 @@
...
 
-InMemoryCache::InMemoryCache(size_t maxSize, ModelCacheCostFunc cacheCost,
-                             TimeProvider timeProvider)
-    : m_itemTuples(maxSize, std::move(cacheCost)),
-      m_timeProvider(std::move(timeProvider)) {
-  m_itemTuples.SetEvictionCallback(
+InMemoryCache::InMemoryCache(size_t max_size, ModelCacheCostFunc cache_cost,
+                             TimeProvider time_provider)
+    : item_tuples_(max_size, std::move(cache_cost)),
+      time_provider_(std::move(time_provider)) {
```
borzun:
```suggestion
      time_provider_(std::move(time_provider)) {
```
It is out of the scope of this ticket - but what if user pass a `nullptr` TimeProvider - should we roll back to the `DefaultTimeProvider` in that case?

olp-cpp-sdk-core/src/cache/InMemoryCache.cpp
andescu: It returns only true if all PurgeExpired returns true.

olp-cpp-sdk-core/src/cache/InMemoryCache.cpp
andescu: Considering it is a private class and not exposed trough public includes I would assume we are the users and we should avoid that. Either way we have a issue in the backlog where we need to expose this so that we can inject a TimeProvider from ourside as default one does not work on most automotive use-cases.

olp-cpp-sdk-core/src/cache/InMemoryCache.cpp
diachenko-mischa: InMemoryCache::PurgeExpired(time_t expire_time) returns true unconditionally

PR #384, opened 2019-11-12
Clean up CatalogClient

diachenko-mischa:
Add includes, that are mandatory for non-trivial class usage. Remove
redundant includes from examples and tests. Remove redundant forward
declarations.

Relates-To: OLPEDGE-798

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/CatalogClient.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/CatalogClient.cpp
@@ -20,8 +20,8 @@
 #include "olp/dataservice/read/CatalogClient.h"
 
 #include "CatalogClientImpl.h"
-#include "olp/core/cache/DefaultCache.h"
-#include "olp/core/client/OlpClientSettingsFactory.h"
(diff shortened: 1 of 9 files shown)


olp-cpp-sdk-dataservice-read/src/CatalogClient.cpp
```diff
@@ -17,19 +17,19 @@
  * License-Filename: LICENSE
  */
 
-#include "olp/dataservice/read/CatalogClient.h"
+#include <olp/dataservice/read/CatalogClient.h>
```
andescu: For own library includes you dont need to use <>.
Only for different libs, e.g. core.

tests/functional/olp-cpp-sdk-dataservice-read/CatalogClientTest.cpp
```diff
@@ -27,6 +27,7 @@
 
 #include <testutils/CustomParameters.hpp>
 
+#include "Utils.h"
```
andescu: Are you sure this sorting is correct?

PR #407, opened 2019-11-15
Move OSX/IOS builds from Travis to Azure

ystefinko:
Relates-to: OLPEDGE-1064

### DIFF
--- a/scripts/macos/psv/azure_macos_build_psv.sh
+++ b/scripts/macos/psv/azure_macos_build_psv.sh
@@ -1,0 +1,29 @@
+#!/bin/bash -e
+#
+# Copyright (C) 2019 HERE Europe B.V.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


scripts/macos/psv/azure_macos_build_psv.sh
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+
+mkdir -p build
+cd build
+cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
```
andescu: Add also -DOLP_SDK_ENABLE_TESTING=NO
Dont see a reason why we should also build tests

ystefinko: Thanks. Added.

PR #408, opened 2019-11-15
Add parallel build option for Visual Studio

diachenko-mischa:
Relates-To: OLPEDGE-0000


diachenko-mischa:
@andescu , yes you are right!

PR #412, opened 2019-11-18
Improved logging for NetworkCurl

borzun:
Improved logging in NetworkCurl implementation by adding request ids, this pointer and url when do logging.

Relates to: OLPEDGE-1051

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
@@ -222,9 +222,15 @@ static curl_code SslctxFunction(CURL* curl, void* sslctx, void*) {
 
 NetworkCurl::NetworkCurl(size_t max_requests_count)
     : handles_(max_requests_count),
-      static_handle_count_(std::max(1lu, max_requests_count / 4)) {}
+      static_handle_count_(std::max(1lu, max_requests_count / 4)) {
+  OLP_SDK_LOG_TRACE(kLogTag, "Created NetworkCurl with address="
+                                 << this
+                                 << ", handles_count=" << max_requests_count);
+}
 
 NetworkCurl::~NetworkCurl() {
+  OLP_SDK_LOG_TRACE(kLogTag, "Destroyed NetworkCurl object, this=" << this);
+
   if (state_ == WorkerState::STARTED) {
     Deinitialize();
   }
@@ -236,18 +242,18 @@ NetworkCurl::~NetworkCurl() {
 bool NetworkCurl::Initialize() {
   std::lock_guard<std::mutex> init_lock(init_mutex_);
   if (state_ != WorkerState::STOPPED) {
-    OLP_SDK_LOG_DEBUG(kLogTag, "Already initialized");
+    OLP_SDK_LOG_DEBUG(kLogTag, "Already initialized, this=" << this);
     return true;
   }
 
 #ifdef NETWORK_HAS_PIPE2
   if (pipe2(pipe_, O_NONBLOCK)) {
-    OLP_SDK_LOG_ERROR(kLogTag, "pipe2 failed");
+    OLP_SDK_LOG_ERROR(kLogTag, "pipe2 failed, this=" << this);
     return false;
   }
 #elif defined NETWORK_HAS_PIPE
   if (pipe(pipe_)) {
-    OLP_SDK_LOG_ERROR(kLogTag, "pipe failed");
+    OLP_SDK_LOG_ERROR(kLogTag, "pipe failed, this=" << this);
     return false;
   }
   // Set pipes non blocking
@@ -256,14 +262,14 @@ bool NetworkCurl::Initialize() {
     flags = 0;
   }
   if (fcntl(pipe_[0], F_SETFL, flags | O_NONBLOCK)) {
-    OLP_SDK_LOG_ERROR(kLogTag, "fcntl for pipe failed");
+    OLP_SDK_LOG_ERROR(kLogTag, "fcntl for pipe failed, this=" << this);
     return false;
   }
   if (-1 == (flags = fcntl(pipe_[1], F_GETFL, 0))) {
     flags = 0;
   }
   if (fcntl(pipe_[1], F_SETFL, flags | O_NONBLOCK)) {
-    OLP_SDK_LOG_ERROR(kLogTag, "fcntl for pipe failed");
+    OLP_SDK_LOG_ERROR(kLogTag, "fcntl for pipe failed, this=" << this);
     return false;
   }
 #endif
@@ -280,7 +286,7 @@ bool NetworkCurl::Initialize() {
   // cURL setup
   curl_ = curl_multi_init();
   if (!curl_) {
-    OLP_SDK_LOG_ERROR(kLogTag, "curl_multi_init failed");
+    OLP_SDK_LOG_ERROR(kLogTag, "curl_multi_init failed, this=" << this);
     return false;
   }
 
@@ -310,9 +316,12 @@ bool NetworkCurl::Initialize() {
 void NetworkCurl::Deinitialize() {
   // Stop worker thread
   if (state_ != WorkerState::STARTED) {
-    OLP_SDK_LOG_DEBUG(kLogTag, "Already deinitialized");
+    OLP_SDK_LOG_DEBUG(kLogTag, "Already deinitialized, this=" << this);
     return;
   }
+
+  OLP_SDK_LOG_TRACE(kLogTag, "Deinitialize NetworkCurl, this=" << this);
+
   {
     std::lock_guard<std::mutex> lock(event_mutex_);
     state_ = WorkerState::STOPPING;
@@ -331,7 +340,8 @@ void NetworkCurl::Teardown() {
 #if (defined NETWORK_HAS_PIPE) || (defined NETWORK_HAS_PIPE2)
   char tmp = 1;
   if (write(pipe_[1], &tmp, 1) < 0) {
-    OLP_SDK_LOG_INFO(kLogTag, "deinitialize - failed to write pipe " << errno);
+    OLP_SDK_LOG_INFO(kLogTag, "Deinitialize, failed to write pipe, err="
+                                  << errno << ", this=" << this);
   }
 #endif
 
@@ -410,6 +420,8 @@ SendOutcome NetworkCurl::Send(NetworkRequest request,
                               Network::DataCallback data_callback) {
   if (!Initialized()) {
     if (!Initialize()) {
+      OLP_SDK_LOG_ERROR(kLogTag, "Send failed - network is uninitialized, url="
+                                     << request.GetUrl());
       return SendOutcome(ErrorCode::OFFLINE_ERROR);
     }
   }
@@ -445,9 +457,9 @@ ErrorCode NetworkCurl::SendImplementation(
     const std::shared_ptr<std::ostream>& payload,
     Network::HeaderCallback header_callback,
     Network::DataCallback data_callback, Network::Callback callback) {
-  OLP_SDK_LOG_TRACE(kLogTag, "send with id = " << id);
-
   if (!IsStarted()) {
+    OLP_SDK_LOG_ERROR(
+        kLogTag, "Send failed - network is offline, url=" << request.GetUrl());
     return ErrorCode::IO_ERROR;
   }
 
@@ -456,9 +468,11 @@ ErrorCode NetworkCurl::SendImplementation(
   RequestHandle* handle = GetHandle(id, callback, header_callback,
                                     data_callback, payload, request.GetBody());
   if (!handle) {
-    OLP_SDK_LOG_WARNING(kLogTag, "curl handle is nullptr");
     return ErrorCode::NETWORK_OVERLOAD_ERROR;
   }
+
+  OLP_SDK_LOG_DEBUG(kLogTag, "Send request with url=" << request.GetUrl()
+                                                      << ", id=" << id);
 
   handle->transfer_timeout = config.GetTransferTimeout();
   handle->max_retries = config.GetRetries();
@@ -557,9 +571,10 @@ ErrorCode NetworkCurl::SendImplementation(
     CURLcode error = curl_easy_setopt(handle->handle, CURLOPT_CAINFO,
                                       curl_ca_bundle.c_str());
     if (CURLE_OK != error) {
+      OLP_SDK_LOG_ERROR(kLogTag, "Send failed - curl_easy_setopt error="
+                                     << error << ", id=" << id);
       return ErrorCode::UNKNOWN_ERROR;
     }
-    OLP_SDK_LOG_DEBUG(kLogTag, "curl bundle path: " << curl_ca_bundle);
   }
 #endif
 
@@ -592,10 +607,8 @@ ErrorCode NetworkCurl::SendImplementation(
   curl_easy_setopt(handle->handle, CURLOPT_ERRORBUFFER, handle->error_text);
 
 #if (LIBCURL_VERSION_MAJOR >= 7) && (LIBCURL_VERSION_MINOR >= 21)
-  // if (config && config->IsAutoDecompressionEnabled()) {
   curl_easy_setopt(handle->handle, CURLOPT_ACCEPT_ENCODING, "");
   curl_easy_setopt(handle->handle, CURLOPT_TRANSFER_ENCODING, 1L);
-  // }
 #endif
 
 #if (LIBCURL_VERSION_MAJOR >= 7) && (LIBCURL_VERSION_MINOR >= 25)
@@ -613,9 +626,8 @@ ErrorCode NetworkCurl::SendImplementation(
 }
 
 void NetworkCurl::Cancel(RequestId id) {
-  OLP_SDK_LOG_TRACE(kLogTag, "cancel with id = " << id);
-
   if (!IsStarted()) {
+    OLP_SDK_LOG_ERROR(kLogTag, "Cancel failed - network is offline, id=" << id);
     return;
   }
   std::lock_guard<std::mutex> lock(event_mutex_);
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -189,9 +189,15 @@ static curl_code SslctxFunction(CURL* curl, void* sslctx, void*) {
 
 NetworkCurl::NetworkCurl(size_t max_requests_count)
     : handles_(max_requests_count),
-      static_handle_count_(std::max(1lu, max_requests_count / 4)) {}
+      static_handle_count_(std::max(1lu, max_requests_count / 4)) {
+  OLP_SDK_LOG_TRACE(kLogTag, "Created NetworkCurl with address="
+                                 << this
+                                 << "; handles_count=" << max_requests_count);
```
andescu: Please use ',' as the separator, as we use it also everywhere else.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -203,18 +209,18 @@ NetworkCurl::~NetworkCurl() {
 bool NetworkCurl::Initialize() {
   std::lock_guard<std::mutex> init_lock(init_mutex_);
   if (state_ != WorkerState::STOPPED) {
-    OLP_SDK_LOG_DEBUG(kLogTag, "Already initialized");
+    OLP_SDK_LOG_DEBUG(kLogTag, "Already initialized this=" << this);
```
andescu: Missing separator.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -297,7 +306,8 @@ void NetworkCurl::Teardown() {
 #if (defined NETWORK_HAS_PIPE) || (defined NETWORK_HAS_PIPE2)
   char tmp = 1;
   if (write(pipe_[1], &tmp, 1) < 0) {
-    OLP_SDK_LOG_INFO(kLogTag, "deinitialize - failed to write pipe " << errno);
+    OLP_SDK_LOG_INFO(kLogTag, "deinitialize - failed to write pipe; err="
```
andescu:
```suggestion
    OLP_SDK_LOG_INFO(kLogTag, "Deinitialize, failed to write pipe, err="
```

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -376,6 +386,9 @@ SendOutcome NetworkCurl::Send(NetworkRequest request,
                               Network::DataCallback data_callback) {
   if (!Initialized()) {
     if (!Initialize()) {
+      OLP_SDK_LOG_ERROR(kLogTag, "Can't send the request with url="
```
andescu: Is this really a error or rather a warning?

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -579,21 +595,22 @@ ErrorCode NetworkCurl::SendImplementation(
 }
 
 void NetworkCurl::Cancel(RequestId id) {
-  OLP_SDK_LOG_TRACE(kLogTag, "cancel with id = " << id);
-
   if (!IsStarted()) {
+    OLP_SDK_LOG_ERROR(kLogTag, "Can't cancel the request id="
```
andescu: Why is this an error? I would say missing accomplished, silent exit.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -561,7 +577,7 @@ ErrorCode NetworkCurl::SendImplementation(
   // if (config && config->IsAutoDecompressionEnabled()) {
```
andescu: remove dead code please

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -422,10 +436,12 @@ ErrorCode NetworkCurl::SendImplementation(
   RequestHandle* handle = GetHandle(id, callback, header_callback,
                                     data_callback, payload, request.GetBody());
   if (!handle) {
-    OLP_SDK_LOG_WARNING(kLogTag, "curl handle is nullptr");
     return ErrorCode::NETWORK_OVERLOAD_ERROR;
   }
 
+  OLP_SDK_LOG_DEBUG(kLogTag, "send request with url=" << request.GetUrl()
```
andescu:
```suggestion
  OLP_SDK_LOG_DEBUG(kLogTag, "Send request with url=" << request.GetUrl()
```

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -411,9 +424,10 @@ ErrorCode NetworkCurl::SendImplementation(
     const std::shared_ptr<std::ostream>& payload,
     Network::HeaderCallback header_callback,
     Network::DataCallback data_callback, Network::Callback callback) {
-  OLP_SDK_LOG_TRACE(kLogTag, "send with id = " << id);
-
   if (!IsStarted()) {
+    OLP_SDK_LOG_ERROR(kLogTag, "can't send the request with url="
```
andescu: Try to find shorter messages, these are just for you to find bugs.
("Send - network offline, ulr=" << request.GetUrl()); Will also do the job and is not 3 lines long.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -579,21 +595,22 @@ ErrorCode NetworkCurl::SendImplementation(
...
       handle.cancelled = true;
       AddEvent(EventInfo::Type::CANCEL_EVENT, &handle);
+
+      OLP_SDK_LOG_TRACE(kLogTag, "cancel request id=" << id);
       return;
     }
   }
-  OLP_SDK_LOG_WARNING(kLogTag, "cancel for non-existing request " << id);
-  return;
+  OLP_SDK_LOG_WARNING(kLogTag, "cancel for non-existing request id=" << id);
```
andescu: Maybe this should not be an warning but rather an debug.
Nothing to cancel is not an error but rather: job done.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -602,10 +619,12 @@ void NetworkCurl::AddEvent(EventInfo::Type type, RequestHandle* handle) {
 #if (defined NETWORK_HAS_PIPE) || (defined NETWORK_HAS_PIPE2)
   char tmp = 1;
   if (write(pipe_[1], &tmp, 1) < 0) {
-    OLP_SDK_LOG_INFO(kLogTag, "addEvent - failed " << errno);
+    OLP_SDK_LOG_INFO(kLogTag, "addEvent - failed for id=" << handle->id
```
andescu:
```suggestion
    OLP_SDK_LOG_INFO(kLogTag, "AddEvent - failed for id=" << handle->id
```

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -602,10 +619,12 @@ void NetworkCurl::AddEvent(EventInfo::Type type, RequestHandle* handle) {
...
+                                                          << " err=" << errno);
   }
 #else
-  OLP_SDK_LOG_WARNING(kLogTag, "addEvent - no pipe");
+  OLP_SDK_LOG_WARNING(kLogTag, "addEvent for id=" << handle->id
```
andescu:
```suggestion
  OLP_SDK_LOG_WARNING(kLogTag, "AddEvent for id=" << handle->id
```

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -614,7 +633,8 @@ NetworkCurl::RequestHandle* NetworkCurl::GetHandle(
     Network::DataCallback data_callback, Network::Payload payload,
     NetworkRequest::RequestBodyType body) {
   if (!IsStarted()) {
-    OLP_SDK_LOG_ERROR(kLogTag, "Network is not started");
+    OLP_SDK_LOG_ERROR(kLogTag, "Can't get a handle for id="
```
andescu: "GetHandle - network offline, id="

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -623,7 +643,9 @@ NetworkCurl::RequestHandle* NetworkCurl::GetHandle(
       if (!handle.handle) {
         handle.handle = curl_easy_init();
         if (!handle.handle) {
-          OLP_SDK_LOG_ERROR(kLogTag, "curl_easy_init returns nullptr");
+          OLP_SDK_LOG_ERROR(kLogTag,
```
andescu: "GetHandle - curl_easy_init failed, id="

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -655,7 +677,8 @@ NetworkCurl::RequestHandle* NetworkCurl::GetHandle(
     }
   }
 
-  OLP_SDK_LOG_WARNING(kLogTag, "all curl handles are in use");
+  OLP_SDK_LOG_WARNING(kLogTag, "Can't get a handle for id="
```
andescu: Same here

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -705,9 +729,10 @@ size_t NetworkCurl::RxFunction(void* ptr, size_t size, size_t nmemb,
         if (handle->payload->tellp() != std::streampos(handle->count)) {
           handle->payload->seekp(handle->count);
           if (handle->payload->fail()) {
-            OLP_SDK_LOG_WARNING(
-                kLogTag,
-                "Reception stream doesn't support setting write point");
+            OLP_SDK_LOG_WARNING(kLogTag,
+                                "Reception stream doesn't support setting "
```
andescu: Too long!

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -740,6 +765,9 @@ size_t NetworkCurl::HeaderFunction(char* ptr, size_t size, size_t nitems,
                                    RequestHandle* handle) {
   const size_t len = size * nitems;
 
+  OLP_SDK_LOG_TRACE(kLogTag, "Received header for id=" << handle->id
+                                                       << "; len=" << len);
```
andescu: Does printing len provide important information?

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -831,7 +861,9 @@ void NetworkCurl::CompleteMessage(CURL* handle, CURLcode result) {
     std::string etag = handles_[index].etag;
     std::string contentType = handles_[index].content_type;
     if (!callback) {
-      OLP_SDK_LOG_WARNING(kLogTag, "Complete to request without callback");
+      OLP_SDK_LOG_WARNING(kLogTag,
```
andescu: CompleteMessage - no callback

I wonder why this is an warning? It is users decision.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -1091,19 +1127,21 @@ void NetworkCurl::Run() {
...
+            OLP_SDK_LOG_WARNING(kLogTag,
+                                "Releasing lost handle for url=" << url);
             CompleteMessage(handle, CURLE_OPERATION_TIMEDOUT);
           } else {
-            OLP_SDK_LOG_ERROR(
-                kLogTag, "lost handle curl_multi_remove_handle error %s "
-                             << std::to_string(remove_handle_status) << " for "
-                             << url);
+            OLP_SDK_LOG_ERROR(kLogTag,
+                              "lost handle curl_multi_remove_handle error "
```
andescu:
```suggestion
                              "Lost handle curl_multi_remove_handle error "
```

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -1091,19 +1127,21 @@ void NetworkCurl::Run() {
...
+                                  << std::to_string(remove_handle_status)
+                                  << " for url=" << url);
 
             int handle_index = GetHandleIndex(handle);
             if (handle_index >= 0) {
               if (!handles_[handle_index].callback) {
-                OLP_SDK_LOG_WARNING(kLogTag,
-                                    "Complete to request without callback");
+                OLP_SDK_LOG_WARNING(kLogTag, "Complete to request id="
```
andescu: "Complete to request" -> "Request completed, no callback, .... "

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: As we can't continue the operation, I thought about Error as it used also in other places in NetworkCurl.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: We can't continue the operation, thus it looks more as an error.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: At least, it is better to print the length of the headers received cause it proves that the SDK have received not-null data - just printing the id doesn't make sense to me.

You suggest to print all headers?

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: True, lets change the logging to Debug?

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
andescu: Ideally we will print headers and values. But this might be very verbose and will flood cout with countless logs. I dont think that headers will be empty so no need to print len. Better remove. We can add if desired.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
andescu: If it applies to all requests then error, if only to one request then warning.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: Any alternative? Note that this log is used in other Network implementations, so we have to align with them too.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: Maybe we should. But because `Cancel` doesn't provide any instant feedback for the user (like returning bool) I think, we should 'warn' the user that they is doing something wrong.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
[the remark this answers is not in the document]
borzun: Combining it with Google Code style - yes, it creates a lot of boilerplate code. But on contrary side, you don't have to deal with formatting mistake for parameters across different compliers.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: Not sure whether I correctly understood you. In this case, all requests will be failed, cause `Initialize() `most likely will be having some permanent problem.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: >> these are just for you to find bugs.
Not only for me, they are useful for other SDK users as well :)

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: I lowered level to Debug

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
borzun: Ok, removed

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
andescu: Most likely they will hand it over to us, if not still they are developers and they will look into the code. So the esthetical nature is not very important here. It is more important to log all the necessary information so that it can be traced back.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
andescu: But then you also have error logs in Initialize() which will be called on each Send() until it succeeds. This is a tricky one. Then leave error.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
andescu: Of course we need. This is legacy code and it is about time that we do some face lifting.
"Payload seekp() failed" for example.

borzun: Renaming this log in other network implementations will be done in corresponding commits.

PR #422, opened 2019-11-19
NetworkCurl now uses RequestId in RequestHandle

meravingen123:
Resolves: OLPEDGE-1077

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.h
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.h
@@ -19,6 +19,7 @@
 
 #pragma once
 
+#include <curl/curl.h>
 #include <olp/core/http/Network.h>
 #include <olp/core/http/NetworkRequest.h>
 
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/src/http/curl/NetworkCurl.h
```diff
@@ -19,6 +19,7 @@
 
 #pragma once
 
+#include <curl/curl.h>
```
andescu: is this needed?

borzun: I guess yes, cause we use CURL related types in `NeworkCurl::RequestHandle` struct and in other callbacks.

andescu: But <curl/curl.h> is part of libcurl and should not be related to RequestId, the only change i see here.

andescu: Missed that it was just moved. Ignore my comment.

PR #445, opened 2019-11-25
Fix client capturing in ApiClientLookup

mykhailo-kuchma:
Relates-To: OLPEDGE-1091

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
@@ -139,7 +139,7 @@ ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApi(
   if (options != OnlineOnly) {
     auto url = repository.Get(service, service_version);
     if (url) {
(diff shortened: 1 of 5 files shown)


olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
```diff
@@ -154,28 +154,36 @@ ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApi(
...
   cancellation_context.ExecuteOrCancelled(
       [&, flag]() {
-        auto token = ApiClientLookup::LookupApiClient(
-            client, service, service_version, catalog,
-            std::move(api_client_callback));
+        const auto& base_url = GetDatastoreServerUrl(catalog.partition);
+
+        client::OlpClient client;
+        client.SetBaseUrl(base_url);
+        client.SetSettings(std::move(settings));
```
diachenko-mischa: There's usages of settings after move

mykhailo-kuchma: might be a problem in the future, now there is no movement, a function takes a const ref.

PR #447, opened 2019-11-25
Align CancellationToken to google style

mykhailo-kuchma:
Resolves: OLPEDGE-1130

### DIFF
--- a/olp-cpp-sdk-core/src/client/CancellationToken.cpp
+++ b/olp-cpp-sdk-core/src/client/CancellationToken.cpp
@@ -21,13 +21,12 @@
 
 namespace olp {
 namespace client {
-CancellationToken::CancellationToken(const std::function<void()> func)
-  : func_(func) {}
+CancellationToken::CancellationToken(std::function<void()> func)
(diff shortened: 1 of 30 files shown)


olp-cpp-sdk-core/src/client/CancellationToken.cpp
```diff
@@ -22,16 +22,19 @@
 namespace olp {
 namespace client {
 CancellationToken::CancellationToken(const std::function<void()> func)
```
andescu: I wonder why this is a const copy and should not be moved.

olp-cpp-sdk-core/src/client/CancellationToken.cpp
```diff
@@ -22,16 +22,19 @@
...
-  : func_(func) {}
-
-CancellationToken::CancellationToken(const CancellationToken& other)
-  : func_(other.func_) {}
+    : func_(func) {}
 
 void CancellationToken::cancel() const {
   if (func_) {
```
andescu: Call Cancel() to avoid duplication.

PR #456, opened 2019-11-26
Add callback for outer-task in Prefetch

diachenko-mischa:
When task is schedulled and cancelled before execution,
no function executed and callback is called. Add callback
for covering such cases.

Relates-To: OLPEDGE-1005

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -170,8 +170,7 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchTiles(
           OLP_SDK_LOG_WARNING_F(kLogTag,
                                 "PrefetchTiles : invalid request, layer=%s",
                                 layer_id.c_str());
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -324,7 +324,14 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchTiles(
       // Because the handling of prefetch tiles responses is performed by the
       // inner-task, no need to set a callback here. Otherwise, the user would
       // be notified with empty results.
-      nullptr);
+      // It is possible to not invoke inner task, when it was cancelled before
+      // execution.
+      [callback](EmptyResponse response) {
```
andescu: Be aware that this is not enough. You need to remove additionaly, if possible, all the user callback triggers from the above lambda and fwd the errors only to this lambda.

andescu: Also please use OLPEDGE-1005 as ticket.

diachenko-mischa: Added error propagation from task.

PR #464, opened 2019-11-28
Added a random cancellation configuration for performance tests

mykhailo-kuchma:
Make a new network for every test run.

Relates-To: OLPEDGE-1156

### DIFF
--- a/tests/performance/MemoryTest.cpp
+++ b/tests/performance/MemoryTest.cpp
@@ -42,11 +42,14 @@ using TestFunction = std::function<void(uint8_t thread_id)>;
 
 struct TestConfiguration {
   std::string configuration_name;
-  std::uint16_t requests_per_second;
-  std::uint8_t calling_thread_count;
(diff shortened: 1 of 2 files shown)


tests/performance/MemoryTest.cpp
```diff
@@ -393,18 +409,20 @@ TestConfiguration ShortRunningTestWithMemoryCache() {
...
-    if (location.empty()) {
-      location = utils::Dir::TempDirectory() + "/performance_test";
-    }
-    settings.disk_path = location;
+
+  cache::CacheSettings settings;
+  auto location = CustomParameters::getArgument("cache_location");
+  if (location.empty()) {
+    location = utils::Dir::TempDirectory() + "/performance_test";
+  }
```
andescu: Can you create the dir in case it does not exist?

mykhailo-kuchma: Our cache will do that

PR #465, opened 2019-11-28
Propagate error_id to user.

mykhailo-kuchma:
Resolves: OLPEDGE-1157


tests/functional/olp-cpp-sdk-authentication/AuthenticationClientTest.cpp
```diff
@@ -173,6 +173,11 @@ class AuthenticationClientTest : public AuthenticationCommonTestFixture {
 
     return request_future.get();
   }
+
+  std::string PrintErrorId(
```
andescu:
```suggestion
  std::string GetErrorId(
```
As it is technically not printing anything but only returning.

PR #471, opened 2019-11-29
Fix retry mechanism in FV/Nv online tests

ystefinko:
Changed way we call retry in shell

Relates-to: OLPEDGE-1144

### DIFF
--- a/scripts/linux/fv/gitlab_test_fv.sh
+++ b/scripts/linux/fv/gitlab_test_fv.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash -ex
 #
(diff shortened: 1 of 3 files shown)


scripts/linux/fv/gitlab_test_fv.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash -x
```
andescu: Below is `-ex` why here only `-x`

ystefinko: Just added comments about that into script. Also added set +e for allowing online for exit not0.
Thanks for review.

PR #473, opened 2019-11-29
Added sync version of LookupApiClient in dataservice-write

borzun:
Implemented the sync version of ApiClientLookup::LookupApiClient in dataservice-write component in the same way it is done in dataservice-read component.
Also, covered this new function with unit tests.

Relates to: OLPEDGE-1103

### DIFF
--- a/olp-cpp-sdk-dataservice-write/src/ApiClientLookup.cpp
+++ b/olp-cpp-sdk-dataservice-write/src/ApiClientLookup.cpp
@@ -20,8 +20,10 @@
 #include "ApiClientLookup.h"
 
 #include <olp/core/client/ApiError.h>
+#include <olp/core/client/Condition.h>
 #include <olp/core/client/HRN.h>
 #include <olp/core/client/OlpClient.h>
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-dataservice-write/src/ApiClientLookup.cpp
```diff
@@ -126,6 +135,110 @@ client::CancellationToken ApiClientLookup::LookupApiClient(
       });
 }
 
+ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApiClient(
```
diachenko-mischa: Please, make order of arguments same as in dataservice-read :
```
  static ApiClientResponse LookupApi(
      const client::HRN& catalog,
      client::CancellationContext cancellation_context, std::string service,
      std::string service_version, FetchOptions options,
      client::OlpClientSettings settings);
```

borzun: We also need to +/- align other dataservice-write:: {{ApiClientLookup}} methods.

PR #478, opened 2019-12-03
Update CacheSettings structure.

mykhailo-kuchma:
* Add disk_path_mutable.
* Deprecate the disk_path in favor of a new disk_path_mutable.
* Update the SDK to use a new disk_path_mutable.
* When disk_path_mutable is empty, disk_path is used (backwards compatibility).

Resolves: OLPEDGE-1039

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
+++ b/olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
@@ -23,6 +23,8 @@
 
 #include <boost/optional.hpp>
 
+#include <olp/core/porting/deprecated.h>
+
 namespace olp {
 
 namespace cache {
@@ -40,7 +42,14 @@ struct CacheSettings {
   /**
    * @brief Path to save contents on disk
    */
+  OLP_SDK_DEPRECATED("Use disk_path_mutable instead. Will be removed 03.2020")
   boost::optional<std::string> disk_path = boost::none;
+
+  /**
+   * @brief Path to the mutable (read-write) cache, where SDK will cache and
+   * lookup the content. User should have write permissions.
+   */
+  boost::optional<std::string> disk_path_mutable = boost::none;
 
   /**
    * @brief Set the upper limit of disk space to use for persistent stores in
--- a/olp-cpp-sdk-core/src/cache/DefaultCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCache.cpp
@@ -17,7 +17,14 @@
  * License-Filename: LICENSE
  */
 
+#include "olp/core/porting/warning_disable.h"
+
+PORTING_PUSH_WARNINGS()
+PORTING_CLANG_GCC_DISABLE_WARNING("-Wdeprecated-declarations")
+// Generated class methods use a deprecated field and generate warning
 #include "olp/core/cache/DefaultCache.h"
+PORTING_POP_WARNINGS()
+
 #include "DiskCache.h"
 #include "InMemoryCache.h"
 #include "olp/core/logging/Log.h"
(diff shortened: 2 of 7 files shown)


r0busta:
I wonder how you guys usually handle this kind of issue when deprecating API and using it at the same time causes the CI to fail. Do you use pragmas to suppress the warnings?

mykhailo-kuchma:
> I wonder how you guys usually handle this kind of issue when deprecating API and using it at the same time causes the CI to fail. Do you use pragmas to suppress the warnings?

Usually, we don't use pragmas. However, it's the first time we have to deprecate a field in a struct.

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -281,7 +288,19 @@ DefaultCache::StorageOpenResult DefaultCache::SetupStorage() {
   if (settings_.max_memory_cache_size > 0) {
     memory_cache_.reset(new InMemoryCache(settings_.max_memory_cache_size));
   }
-  if (settings_.disk_path) {
+
+  // For backwards compatibility.
+  // If the user provided old disk_path instead of a new one, use it.
```
r0busta: I'm not sure about the team's agreement on this topic. So I'm not bikeshedding, rather just asking and thinking out loud.

Is this kind of comment necessary and useful? The comment explains the How part, meanwhile it's preferable for the code comments to explain the Why part of the code and for the code itself to explain the How, isn't it? 

That is. It's obvious from line 299 that we're going to use the deprecated setting, so the comment at line 293 loses its usefulness.

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -281,7 +288,19 @@ DefaultCache::StorageOpenResult DefaultCache::SetupStorage() {
...
+  if (!settings_.disk_path_mutable) {
+    PORTING_PUSH_WARNINGS()
+    PORTING_CLANG_GCC_DISABLE_WARNING("-Wdeprecated-declarations")
+    PORTING_MSVC_DISABLE_WARNINGS(4996)
+    // Direct usage of deprecated field
```
r0busta: Same question about the usage of comments.

olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
```diff
@@ -77,6 +85,11 @@ struct CacheSettings {
    * @brief Set the disk cache open options.
    */
   OpenOptions openOptions = OpenOptions::Default;
+
+  /**
+   * @brief Path to the readonly cache.
```
r0busta: Would it make sense to explain why the option is called blah_protected and not blah_readonly? Is the meaning of the word "protected" clear from some other place in the docs?

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
mykhailo-kuchma: Changed the comment.

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
mykhailo-kuchma: removed

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -281,7 +288,17 @@ DefaultCache::StorageOpenResult DefaultCache::SetupStorage() {
...
-  if (settings_.disk_path) {
+
+  // Temporary code for backwards compatibility.
+  if (!settings_.disk_path_mutable) {
+    PORTING_PUSH_WARNINGS()
+    PORTING_CLANG_GCC_DISABLE_WARNING("-Wdeprecated-declarations")
+    PORTING_MSVC_DISABLE_WARNINGS(4996)
+    settings_.disk_path_mutable = settings_.disk_path;
+    PORTING_POP_WARNINGS()
+  }
```
andescu: Please move to an anonymous helper so that we avoid this in the middle of the code.
Something like ValidateDiskPath(settings_).

olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
```diff
@@ -40,8 +42,16 @@ struct CacheSettings {
   /**
    * @brief Path to save contents on disk
    */
+  OLP_SDK_DEPRECATED(
+      "Deprecated. Use disk_path_mutable instead. Will be removed march 2020")
```
andescu:
```suggestion
      "Use disk_path_mutable instead. Will be removed 03.2020")
```

PR #504, opened 2019-12-12
Added sleep after DeleteUser() call in authentication tests

meravingen123:
Added sleep after DeleteUser call in authentication tests to
prevent server error when signin to deleted user to check that
it has been deleted.

Resolves: OLPEDGE-799

### DIFF
--- a/tests/functional/olp-cpp-sdk-authentication/AuthenticationClientTest.cpp
+++ b/tests/functional/olp-cpp-sdk-authentication/AuthenticationClientTest.cpp
@@ -27,6 +27,8 @@
 
 namespace {
 
+constexpr auto kDeleteUserWait = std::chrono::seconds(5);
+
 class AuthenticationClientTest : public AuthenticationCommonTestFixture {
  protected:
   void SetUp() override {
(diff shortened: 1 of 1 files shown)


tests/functional/olp-cpp-sdk-authentication/AuthenticationCommonTestFixture.cpp
```diff
@@ -98,6 +100,11 @@ AuthenticationCommonTestFixture::AcceptTerms(
 AuthenticationTestUtils::DeleteUserResponse
 AuthenticationCommonTestFixture::DeleteUser(
     const std::string& user_bearer_token) {
+  // Before deleting user we must wait some time before server finishes
+  // SignUp operation. In case we delete user just after SignUp
+  // ( which is unreal use case ) server will send us an error.
+  std::this_thread::sleep_for(kDeleteUserWait);
```
r0busta: Would it make sense replacing the sleep with something more reliable? E.g. looking up if the user successfully signed up.

tests/functional/olp-cpp-sdk-authentication/AuthenticationCommonTestFixture.cpp
```diff
@@ -28,6 +28,8 @@
 #include "AuthenticationTestUtils.h"
 #include "TestConstants.h"
 
+constexpr auto kDeleteUserWait = std::chrono::seconds(5);
```
r0busta: 5 seconds is too much, isn't it?

tests/functional/olp-cpp-sdk-authentication/AuthenticationCommonTestFixture.cpp
meravingen123: we do a signup check later but when server is not completed deleting user it sends us http status 404 which is not the right thing to do because we may miss broken url later.

tests/functional/olp-cpp-sdk-authentication/AuthenticationCommonTestFixture.cpp
meravingen123: we cannot be sure how much time must pass before sign in later. As discussed in email this is not UI user case to do it immediatly. So in my opinion 5seconds is not too fast for UI case and is not too long to block CI job too much.

PR #513, opened 2019-12-17
Add unit tests for the MessageFormatter class

r0busta:
Relates-To: OLPEDGE-1237

Signed-off-by: Kirill Zhuchkov <643378+r0busta@users.noreply.github.com>

### DIFF
--- a/olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
+++ b/olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
@@ -1,0 +1,504 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include <gtest/gtest.h>
+#include <olp/core/logging/MessageFormatter.h>
+
+namespace {
+
+using namespace olp::logging;
+
+TEST(MessageFormatterTest, ElementConstructors) {
+  MessageFormatter::Element string_element(
+      MessageFormatter::ElementType::String);
+  EXPECT_EQ(MessageFormatter::ElementType::String, string_element.type);
+  EXPECT_EQ("", string_element.format);
+  EXPECT_EQ(0, string_element.limit);
+
+  MessageFormatter::Element level_element(MessageFormatter::ElementType::Level);
+  EXPECT_EQ(MessageFormatter::ElementType::Level, level_element.type);
+  EXPECT_EQ("%s", level_element.format);
+  EXPECT_EQ(0, level_element.limit);
+
+  MessageFormatter::Element tag_element(MessageFormatter::ElementType::Tag);
+  EXPECT_EQ(MessageFormatter::ElementType::Tag, tag_element.type);
+  EXPECT_EQ("%s", tag_element.format);
+  EXPECT_EQ(0, tag_element.limit);
+
+  MessageFormatter::Element message_element(
+      MessageFormatter::ElementType::Message);
+  EXPECT_EQ(MessageFormatter::ElementType::Message, message_element.type);
+  EXPECT_EQ("%s", message_element.format);
+  EXPECT_EQ(0, message_element.limit);
+
+  MessageFormatter::Element file_element(MessageFormatter::ElementType::File);
+  EXPECT_EQ(MessageFormatter::ElementType::File, file_element.type);
+  EXPECT_EQ("%s", file_element.format);
+  EXPECT_EQ(0, file_element.limit);
+
+  MessageFormatter::Element line_element(MessageFormatter::ElementType::Line);
+  EXPECT_EQ(MessageFormatter::ElementType::Line, line_element.type);
+  EXPECT_EQ("%u", line_element.format);
+  EXPECT_EQ(0, line_element.limit);
+
+  MessageFormatter::Element function_element(
+      MessageFormatter::ElementType::Function);
+  EXPECT_EQ(MessageFormatter::ElementType::Function, function_element.type);
+  EXPECT_EQ("%s", function_element.format);
+  EXPECT_EQ(0, function_element.limit);
+
+  MessageFormatter::Element full_function_element(
+      MessageFormatter::ElementType::FullFunction);
+  EXPECT_EQ(MessageFormatter::ElementType::FullFunction,
+            full_function_element.type);
+  EXPECT_EQ("%s", full_function_element.format);
+  EXPECT_EQ(0, full_function_element.limit);
+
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+#include <gtest/gtest.h>
+#include <olp/core/logging/MessageFormatter.h>
+
+using namespace ::testing;
+
+
+namespace olp {
+namespace logging {
+
+TEST( MessageFormatterTest, ElementConstructors )
```
borzun:
```suggestion
TEST(MessageFormatterTest, ElementConstructors)
```
Please, run clang-format on this file.

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+ */
+
+#include <gtest/gtest.h>
+#include <olp/core/logging/MessageFormatter.h>
+
+using namespace ::testing;
+
+
+namespace olp {
+namespace logging {
```
borzun: Use just anonymous `namespace {`instead - refer to other tests.

The idea is to not expose the tests' symbols for other translation units.

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+
+TEST( MessageFormatterTest, ElementConstructors )
+{
+    MessageFormatter::Element stringElement( MessageFormatter::ElementType::String );
```
borzun:
```suggestion
    MessageFormatter::Element string_element( MessageFormatter::ElementType::String );
```
Please, change the naming of variables according to the google C++ code style: https://google.github.io/styleguide/cppguide.html#Variable_Names

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+    EXPECT_TRUE( element1 != element2 );
+
+    element2 = element1;
+    element2.format = "%u";
+    EXPECT_FALSE( element1 == element2 );
+    EXPECT_TRUE( element1 != element2 );
+
+    element2 = element1;
+    element2.limit = 20;
+    ;
```
borzun:
```suggestion
```

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+{
+    MessageFormatter formatter;
+    EXPECT_TRUE( formatter.getElements( ).empty( ) );
+    EXPECT_EQ( MessageFormatter::defaultLevelNameMap( ), formatter.getLevelNameMap( ) );
+    EXPECT_EQ( MessageFormatter::Timezone::Local, formatter.getTimezone( ) );
+}
+
+TEST( MessageFormatterTest, Constructor )
+{
+    std::vector< MessageFormatter::Element > elements
```
borzun:
```suggestion
static const  std::vector< MessageFormatter::Element > elements
```
No need to construct const vector each time test is run

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+        = {MessageFormatter::Element( MessageFormatter::ElementType::String, "LOG: " ),
+           MessageFormatter::Element( MessageFormatter::ElementType::Level, "%s " ),
+           MessageFormatter::Element( MessageFormatter::ElementType::Tag, "%s - " ),
+           MessageFormatter::Element( MessageFormatter::ElementType::File, "%30s:", -30 ),
+           MessageFormatter::Element( MessageFormatter::ElementType::Line, "%5u " ),
+           MessageFormatter::Element( MessageFormatter::ElementType::Time, "[%H:%M] " ),
+           MessageFormatter::Element( MessageFormatter::ElementType::Message )};
+
+    MessageFormatter::LevelNameMap levelNameMap
```
borzun:
```suggestion
    static const MessageFormatter::LevelNameMap levelNameMap
```
same here

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+    EXPECT_FALSE( formatter.getElements( ).empty( ) );
+    EXPECT_EQ( MessageFormatter::defaultLevelNameMap( ), formatter.getLevelNameMap( ) );
+    EXPECT_EQ( MessageFormatter::Timezone::Local, formatter.getTimezone( ) );
+}
+
+TEST( MessageFormatterTest, SetElements )
+{
+    MessageFormatter formatter;
+
+    std::vector< MessageFormatter::Element > elements
```
borzun: Same vector as with `MessageFormatterTest.Constructor` test. Consider to move this vector into common place (anonymous namespace).

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+    EXPECT_NE( elements, formatter.getElements( ) );
+    formatter.setElements( elements );
+    EXPECT_EQ( elements, formatter.getElements( ) );
+}
+
+TEST( MessageFormatterTest, SetLevelNameMap )
+{
+    MessageFormatter formatter;
+
+    MessageFormatter::LevelNameMap levelNameMap
```
borzun:
```suggestion
    MessageFormatter::LevelNameMap levelNameMap
```
Same here - consider to move this map into common place (anonymous namespace on top), as it is used in `MessageFormatterTest.Constructor` test.

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+    formatter.setElements( elements );
+    EXPECT_EQ( "ing", formatter.format( message ) );
+
+    elements[ 0 ].limit = 3;
+    formatter.setElements( elements );
+    EXPECT_EQ( "I a", formatter.format( message ) );
+}
+
+} // namespace logging
+} // namespace olp
```
borzun: missing newline at the end of file

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
r0busta: I don't mind changing but don't think that it's something to worry about. Such optimization gets a fraction of extra performance. And not to mention that compilers will probably optimize that anyway.

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
r0busta: Moving it away will decrease tests' readability.

r0busta:
> Please, add more details into description of the PR and commit message (according to our commit message template)

@borzun, what do you suggest adding extra that is not mentioned in the subject?

olp-cpp-sdk-core/tests/logging/MessageFormatterTest.cpp
borzun: But keeping it in several places will violate the DRY and you could potentially have misaligned containers.

r0busta: DRY is nice but tests readability is nicer :)
Misaligned containers are also not a problem because every container is different and belongs to an isolated test case.

This is the classical case of how to improve the code that looks identical but has a bit of difference. Moving it into a function will result in overgeneralization. Keeping it as it is will annoy with its duplication.

PR #518, opened 2019-12-17
Fix flaky tests in dataservice write versioned and volatile layer

mykhailo-kuchma:
Fix flaky tests in dataservice-write versioned and volatile layer clients.

* Added a sleep between retry.
* Reverted the assert back, to mark the right test as failed.

Resolves: OLPEDGE-708

### DIFF
--- a/tests/functional/olp-cpp-sdk-dataservice-write/DataserviceWriteVersionedLayerClientTest.cpp
+++ b/tests/functional/olp-cpp-sdk-dataservice-write/DataserviceWriteVersionedLayerClientTest.cpp
@@ -41,6 +41,10 @@ const std::string kLayer2 = "layer2";
 const std::string kLayerSdii = "layer_sdii";
 const std::string kVersionedLayer = "versioned_layer";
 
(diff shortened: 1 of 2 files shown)


tests/functional/olp-cpp-sdk-dataservice-write/DataserviceWriteVersionedLayerClientTest.cpp
```diff
@@ -40,6 +40,7 @@ const std::string kLayer = "layer";
 const std::string kLayer2 = "layer2";
 const std::string kLayerSdii = "layer_sdii";
 const std::string kVersionedLayer = "versioned_layer";
+const auto kWaitBeforeRetry = std::chrono::seconds(6);
```
borzun: Please, add comment on this constant about why it was

mykhailo-kuchma: Done

PR #520, opened 2019-12-17
Added protected cache usage in DefaultCache::Get method

dpoberezh:
Protected cache should be checked before mutable cache.
Added test to cover that.

Resolves: OLPEDGE-1040

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCache.cpp
@@ -195,21 +195,15 @@ boost::any DefaultCache::Get(const std::string& key, const Decoder& decoder) {
     }
   }
 
-  if (mutable_cache_) {
-    auto expiry = GetRemainingExpiryTime(key, *mutable_cache_);
-    if (expiry <= 0) {
-      PurgeDiskItem(key, *mutable_cache_);
-      return boost::any();
-    }
+  auto disc_cache = GetFromDiscCache(key);
 
-    auto value = mutable_cache_->Get(key);
-    if (value) {
-      auto decoded_item = decoder(value.get());
-      if (memory_cache_) {
-        memory_cache_->Put(key, decoded_item, expiry, value->size());
-      }
-      return decoded_item;
+  if (disc_cache) {
+    auto decoded_item = decoder(disc_cache->first);
+    if (memory_cache_) {
+      memory_cache_->Put(key, decoded_item, disc_cache->second,
+                         disc_cache->first.size());
     }
+    return decoded_item;
   }
 
   return boost::any();
@@ -231,22 +225,16 @@ std::shared_ptr<std::vector<unsigned char>> DefaultCache::Get(
     }
   }
 
-  if (mutable_cache_) {
-    auto expiry = GetRemainingExpiryTime(key, *mutable_cache_);
-    if (expiry <= 0) {
-      PurgeDiskItem(key, *mutable_cache_);
-      return nullptr;
-    }
+  auto disc_cache = GetFromDiscCache(key);
 
-    auto value = mutable_cache_->Get(key);
-    if (value) {
-      auto data = std::make_shared<std::vector<unsigned char>>(
-          value.get().begin(), value.get().end());
-      if (memory_cache_) {
-        memory_cache_->Put(key, data, expiry, value->size());
-      }
-      return data;
+  if (disc_cache) {
+    auto data = std::make_shared<std::vector<unsigned char>>(
+        disc_cache->first.begin(), disc_cache->first.end());
+
+    if (memory_cache_) {
+      memory_cache_->Put(key, data, disc_cache->second, data->size());
     }
+    return data;
   }
 
   return nullptr;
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
```diff
@@ -171,6 +171,88 @@ TEST(DefaultCacheTest, ExpiredDiskTest) {
...
+    ASSERT_EQ(olp::cache::DefaultCache::Success, cache.Open());
+
+    ASSERT_TRUE(cache.Remove(key1));
+
+    auto key1_data_read =
+        cache.Get(key1, [](const std::string& data) { return data; });
+    ASSERT_FALSE(key1_data_read.empty());
+    ASSERT_EQ(key1_data_string, boost::any_cast<std::string>(key1_data_read));
+  }
+}
```
mykhailo-kuchma: Can we also test Put to protected?

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -49,6 +51,21 @@ time_t GetRemainingExpiryTime(const std::string& key,
   return expiry;
 }
 
+/// Return success flag, cache data and remaing expiry time
+std::tuple<bool, DataPtr, time_t> GetFromCache(
```
mykhailo-kuchma: We should have something like optional<Result>, result = data + expiry

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -195,58 +212,41 @@ boost::any DefaultCache::Get(const std::string& key, const Decoder& decoder) {
...
 
-    auto value = mutable_cache_->Get(key);
-    if (value) {
-      auto decoded_item = decoder(value.get());
-      if (memory_cache_) {
-        memory_cache_->Put(key, decoded_item, expiry, value->size());
-      }
-      return decoded_item;
+  if (valid) {
+    auto decoded_item = decoder({result->begin(), result->end()});
```
andescu: Why do you copy result into the decoder? Seems like a design error to me. If the cost of having a common method is to copy every time one additional time, then i would say: please change back!

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -34,6 +34,8 @@ namespace {
 
 constexpr auto kLogTag = "DefaultCache";
 
+using DataPtr = std::shared_ptr<std::vector<unsigned char>>;
```
andescu: It is not `DataPtr` but rather `ValueType` as in key/value

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -49,6 +51,21 @@ time_t GetRemainingExpiryTime(const std::string& key,
   return expiry;
 }
 
+/// Return success flag, cache data and remaing expiry time
+std::tuple<bool, DataPtr, time_t> GetFromCache(
+    const std::string& key, olp::cache::DiskCache& disk_cache) {
+  auto expiry = GetRemainingExpiryTime(key, disk_cache);
+  if (expiry > 0) {
```
andescu: I would not count that we have expiry in the protected cache at all. With this we will not be able to find any key in the cache. Please change. Expiry should be optional here.

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
andescu: Why do you need the bool for? shared_ptr == nullptr is not enough?

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -195,58 +212,41 @@ boost::any DefaultCache::Get(const std::string& key, const Decoder& decoder) {
...
-    }
-  }
-
-  if (mutable_cache_) {
-    auto expiry = GetRemainingExpiryTime(key, *mutable_cache_);
-    if (expiry <= 0) {
-      PurgeDiskItem(key, *mutable_cache_);
-      return nullptr;
-    }
+  auto data =
```
andescu: Data copy is not good, just to avoid duplication! Please find a more effective way!

olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
dpoberezh: Added

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
dpoberezh: Redone.

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
dpoberezh: Redone

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
dpoberezh: Redone

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
dpoberezh: I've added a bool to ease the if conditions in Get method. Otherwise, we will need to checks both the data and the expiry. Though, it is not relevant anymore.

PR #521, opened 2019-12-18
Mark TokenRequest class as deprecated. 

LiubovDidkivska:
Disable warnings for all current usages of TokenRequest.

Relates-To: OLPEDGE-1263


borzun:
@LiubovDidkivska , @mykhailo-kuchma 
Do we really need to suppress the deprecated warnings?

PR #525, opened 2019-12-19
Mark TokenProvider class as deprecated.

LiubovDidkivska:
Relates-To: OLPEDGE-1258

### DIFF
--- a/olp-cpp-sdk-authentication/include/olp/authentication/TokenProvider.h
+++ b/olp-cpp-sdk-authentication/include/olp/authentication/TokenProvider.h
@@ -28,6 +28,8 @@
 #include "TokenEndpoint.h"
 #include "TokenResult.h"
 
+#include "olp/core/porting/deprecated.h"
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-authentication/include/olp/authentication/TokenProvider.h
```diff
@@ -39,7 +41,8 @@ namespace authentication {
  * @param minimumValidity the minimum token validity to use in seconds.
  */
 template <long long MinimumValidity>
-class TokenProvider {
+class OLP_SDK_DEPRECATED("Deprecated. Will be removed by 04.2020")
+TokenProvider {
```
r0busta: It seems that this line is not formatted, is it?

LiubovDidkivska: It is formatted with clang-format-6.0

PR #527, opened 2019-12-20
Implemented the sync version of PublishDataLessThanTwentyMib

borzun:
1. Implemented the sync version of PublishDataLessThanTwentyMib in dataservice-write::StreamLayerClientImpl class, which uses sync generated API (ConfigApi and IngestApi).
2. Also, added 'unit tests' to the StreamLayerClientTest.cpp regarding this function. I suggest to move those tests to `integration` tests component, once we will finish moving of `PublishData` method to the sync API.
3. Added StreamLayerClientImpl::FindContentTypeForLayerId with catalog as a parameter.

Relates-to: OLPEDGE-1181

### DIFF
--- a/olp-cpp-sdk-dataservice-write/tests/StreamLayerClientImplTest.cpp
+++ b/olp-cpp-sdk-dataservice-write/tests/StreamLayerClientImplTest.cpp
@@ -18,6 +18,7 @@
  */
 
 #include <gmock/gmock.h>
+#include <matchers/NetworkUrlMatchers.h>
 #include <mocks/CacheMock.h>
 #include <mocks/NetworkMock.h>
 #include <olp/core/client/OlpClientSettingsFactory.h>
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-write/tests/StreamLayerClientImplTest.cpp
```diff
@@ -41,6 +73,12 @@ class MockStreamLayerClientImpl : public StreamLayerClientImpl {
               (model::PublishSdiiRequest request,
                client::CancellationContext context),
               (override));
+
+  PublishDataResponse TestPublishDataLessThanTwentyMibSync(
+      const model::PublishDataRequest& request,
+      client::CancellationContext context) {
+    return PublishDataLessThanTwentyMibSync(request, context);
+  }
```
mykhailo-kuchma: can be replaced with using StreamLayerClientImpl::PublishDataLessThanTwentyMibSync

PR #538, opened 2019-12-31
Fix RapidJSON package version to v1.1.0

mykhailo-kuchma:
Additionally:
* Install rapidjson dependency (for future packing).
* Support both v1.1.0 tag version and master.
  They both have same version but different include variable.

Resolves: OLPEDGE-1290


CMakeLists.txt
```diff
@@ -70,7 +70,9 @@ include(coverage)
 
 if(OLP_SDK_BUILD_EXTERNAL_DEPS)
     # Add third-party dependencies
+    set(EXTERNAL_BINARY_INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/install)
     add_subdirectory(external)
+    set(CMAKE_PREFIX_PATH ${EXTERNAL_BINARY_INSTALL_DIR})
```
borzun: I think you don't need to overwrite the existing `CMAKE_PREFIX_PATH`, which can be set externally by user. Better to add it via `list(APPEND)`:
```suggestion
    list(APPEND CMAKE_PREFIX_PATH ${EXTERNAL_BINARY_INSTALL_DIR})
```

external/CMakeLists.txt
```diff
@@ -29,12 +26,29 @@ get_filename_component(compilername ${CMAKE_MAKE_PROGRAM} NAME)
...
+                            -DANDROID_STL=${ANDROID_STL}
+                            -DANDROID_ABI=${ANDROID_ABI})
+elseif(IOS)
+  set(COMMON_PLATFORM_FLAGS -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+                            -DPLATFORM=${PLATFORM}
+                            -DSIMULATOR=${SIMULATOR})
+endif()
+
+if(BUILD_SHARED_LIBS)
+  set(COMMON_PLATFORM_FLAGS ${COMMON_PLATFORM_FLAGS} -DCMAKE_POSITION_INDEPENDENT_CODE=ON)
```
borzun: Are you sure we need to enable it? Looks like by default, it should be turned on:
https://cmake.org/cmake/help/v3.4/prop_tgt/POSITION_INDEPENDENT_CODE.html#prop_tgt:POSITION_INDEPENDENT_CODE :

> This property is True by default for SHARED and MODULE library targets and False otherwise.

PR #541, opened 2019-12-31
Implemented StreamLayerClientImpl::PublishData that uses sync API

borzun:
1. Implemented the StreamLayerClientImpl::PublishData to use the sync API methods:
 - PublishDataLessThanTwentyMibSync;
 - PublishDataGreaterThanTwentyMibSync;
2. Switched the public PublishData API and tests to use those new methods.
3. Renamed old versions of PublishData into PublishDataOld

Relates-to: OLPEDGE-1183


borzun:
Code coverage will be back to the previous values after completion of OLPEDGE-1184 & OLPEDGE-1185 tasks.

PR #546, opened 2020-01-03
Added parsing of Client Id from Access Token

meravingen123:
Now ClientId that is associated with access token can be
retrieved using SignInResult::GetClientId() method

Resolves: OLPEDGE-1287

### DIFF
--- a/olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
+++ b/olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
@@ -19,17 +19,33 @@
 
 #include "SignInResultImpl.h"
 
+#include <vector>
+
 #include "Constants.h"
 #include "olp/core/http/HttpStatusCode.h"
+#include "olp/core/logging/Log.h"
+#include "olp/core/utils/Base64.h"
 
 namespace {
 constexpr auto kTokenType = "tokenType";
 constexpr auto kUserId = "userId";
 constexpr auto kScope = "scope";
+constexpr auto kLogTag = "SignIn";
+
 }  // namespace
 
 namespace olp {
 namespace authentication {
+
+namespace {
+bool IsJwtToken(const std::string& token) {
+  return token.front() == 'e';
+}
+
+bool IsHnToken(const std::string& token) {
+    return token.front() == 'h';
+}
+}  // namespace
 
 SignInResultImpl::SignInResultImpl() noexcept
     : SignInResultImpl(http::HttpStatusCode::SERVICE_UNAVAILABLE,
@@ -50,8 +66,10 @@ SignInResultImpl::SignInResultImpl(
       status_ = http::HttpStatusCode::SERVICE_UNAVAILABLE;
       error_.message = Constants::ERROR_HTTP_SERVICE_UNAVAILABLE;
     } else {
-      if (json_document->HasMember(Constants::ACCESS_TOKEN))
+      if (json_document->HasMember(Constants::ACCESS_TOKEN)) {
         access_token_ = (*json_document)[Constants::ACCESS_TOKEN].GetString();
+        client_id_ = ParseClientIdFromToken(access_token_.c_str());
+      }
       if (json_document->HasMember(kTokenType))
         token_type_ = (*json_document)[kTokenType].GetString();
       if (json_document->HasMember(Constants::REFRESH_TOKEN))
@@ -73,6 +91,8 @@ const std::string& SignInResultImpl::GetAccessToken() const {
   return access_token_;
 }
 
+const std::string& SignInResultImpl::GetClientId() const { return client_id_; }
+
 const std::string& SignInResultImpl::GetTokenType() const {
   return token_type_;
 }
@@ -90,6 +110,70 @@ const std::string& SignInResultImpl::GetUserIdentifier() const {
 const std::string& SignInResultImpl::GetScope() const { return scope_; }
 
 bool SignInResultImpl::IsValid() const { return is_valid_; }
+
+std::string SignInResultImpl::ParseJwtToken(const std::string& token) {
+  const std::string::size_type first_dot = token.find_first_of('.');
+  if (first_dot == std::string::npos) {
+    OLP_SDK_LOG_ERROR(kLogTag, "Cannot parse ClientId. Wrong token format.");
+    return std::string();
+  }
+
+  const std::string jws_header_encoded = token.substr(0, first_dot);
+
+  std::vector<std::uint8_t> decoded_bytes;
+  if (!utils::Base64Decode(jws_header_encoded, decoded_bytes)) {
+    OLP_SDK_LOG_ERROR(kLogTag,
+                      "Cannot parse ClientId. Non-decodable token format");
+    return std::string();
+  }
+
+  std::string jws_header_decoded;
+  std::copy(decoded_bytes.begin(), decoded_bytes.end(),
+            std::back_inserter(jws_header_decoded));
+
+  rapidjson::Document doc;
+  doc.Parse(jws_header_decoded.c_str());
+
+  if (doc.HasParseError() || !doc.IsObject()) {
+    OLP_SDK_LOG_ERROR(kLogTag, "Cannot parse ClientId. Defective token format");
+    return std::string();
+  }
+
+  auto client_id_field = doc.FindMember("aid");
+  if (client_id_field == doc.MemberEnd() ||
+      !client_id_field->value.IsString()) {
+    OLP_SDK_LOG_ERROR(kLogTag,
+                      "Cannot parse ClientId. Field does not exist or is not "
+                      "a string json value");
+    return std::string();
+  }
+
+  const std::string result = client_id_field->value.GetString();
+
+  if (result.empty()) {
+    OLP_SDK_LOG_ERROR(kLogTag,
+                      "Cannot parse ClientId. Incomplete token format");
+  }
+
+  return result;
+}
+
+std::string SignInResultImpl::ParseClientIdFromToken(const std::string& token) {
+  if (token.empty()) {
+    OLP_SDK_LOG_ERROR(kLogTag, "Token is empty");
+    return std::string();
+  }
+
+  if (IsJwtToken(token)) {
+    return ParseJwtToken(token);
+  } else if (IsHnToken(token)) {
+    OLP_SDK_LOG_ERROR(kLogTag, "hN Tokens are not supported!");
+    return std::string();
+  } else {
+    OLP_SDK_LOG_ERROR(kLogTag, "Unknown token format");
+    return std::string();
+  }
+}
 
 }  // namespace authentication
 }  // namespace olp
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +103,63 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
 
 bool SignInResultImpl::IsValid() const { return is_valid_; }
 
+std::string SignInResultImpl::ParseJwtToken(const std::string& token) {
+  std::string result;
```
borzun:
```suggestion
```
No need to initialize this here - better create this var when it is actually used - at line 136

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +103,63 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
...
+  rapidjson::Document doc;
+  doc.Parse(jws_header_decoded.c_str());
+
+  if (doc.HasParseError() || !doc.HasMember("aid")) {
+    OLP_SDK_LOG_ERROR_F(kLogTag,
+                        "Cannot parse ClientId. Defective token format");
+    return result;
+  }
+
+  result = doc["aid"].GetString();
```
borzun:
```suggestion
const std::string result = doc["aid"].GetString();
```
See my previous comment about result variable

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +103,63 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
 
 bool SignInResultImpl::IsValid() const { return is_valid_; }
 
+std::string SignInResultImpl::ParseJwtToken(const std::string& token) {
+  std::string result;
+  std::string::size_type first_dot = token.find_first_of('.');
+  if (std::string::npos == first_dot) {
+    OLP_SDK_LOG_ERROR_F(kLogTag, "Cannot parse ClientId. Wrong token format.");
+    return result;
```
borzun:
```suggestion
    return std::string{}; // or ""
```
The reason to return empty string is to protect ourself from some mistake when `result` variable might be initialized to some value before by mistake. Also, returning empty string directly improves readability and no need to check what value the `result` can have.

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +103,63 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
...
+  if (result.empty()) {
+    OLP_SDK_LOG_ERROR_F(kLogTag,
+                        "Cannot parse ClientId. Incomplete token format");
+  }
+
+  return result;
+}
+
+std::string SignInResultImpl::ParseClientIdFromToken(const std::string& token) {
+  std::string result;
```
borzun:
```suggestion
```
Same here - why not return `std::string{}` directly? This will protect us from some mistakes when result might be initialized by mistake to some value.

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +103,63 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
...
+  if (token.empty()) {
+    return result;
+  }
+
+  if (token.at(0) == 'e') {
```
borzun:
```suggestion
  if (token.front() == 'e') {
```
No need to call `at`, as we already verified the string's size at the top of this function and can safely access this string without any bounds checking.

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +103,63 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
...
+    // This is a "JWT Token"
+    return ParseJwtToken(token);
+  } else if (token.at(0) == 'h') {
```
borzun:
```suggestion
  } else if (token.front() == 'h') {
```

olp-cpp-sdk-authentication/src/SignInResultImpl.h
```diff
@@ -50,6 +50,12 @@ class SignInResultImpl : public BaseResult {
    */
   const std::string& GetTokenType() const;
 
+  /**
+   * @brief Client Id getter methid
```
borzun:
```suggestion
   * @brief Client Id getter method
```

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +103,63 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
 
 bool SignInResultImpl::IsValid() const { return is_valid_; }
 
+std::string SignInResultImpl::ParseJwtToken(const std::string& token) {
+  std::string result;
+  std::string::size_type first_dot = token.find_first_of('.');
```
borzun:
```suggestion
const std::string::size_type first_dot = token.find_first_of('.');
```
Prefer to declare the variables const by default.

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +103,63 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
...
+  if (std::string::npos == first_dot) {
+    OLP_SDK_LOG_ERROR_F(kLogTag, "Cannot parse ClientId. Wrong token format.");
+    return result;
+  }
+
+  std::string jws_header_encoded = token.substr(0, first_dot);
```
borzun:
```suggestion
  const std::string jws_header_encoded = token.substr(0, first_dot);
```

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
r0busta: I'd also add a line in the public docs that, if there is a parse error, the returned client ID is empty. That will help users to avoid unpleasant surprises with empty client IDs.

olp-cpp-sdk-authentication/src/SignInResultImpl.h
```diff
@@ -50,6 +50,12 @@ class SignInResultImpl : public BaseResult {
...
+   * @return string containing Client Id that is associated with access token
```
r0busta:
```suggestion
   * @return string containing the client ID that is associated with the access token
```

olp-cpp-sdk-authentication/src/SignInResultImpl.h
r0busta:
```suggestion
   * @brief The client ID getter method
```

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +101,70 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
 
 bool SignInResultImpl::IsValid() const { return is_valid_; }
 
+std::string SignInResultImpl::ParseJwtToken(const std::string& token) {
+  const std::string::size_type first_dot = token.find_first_of('.');
+  if (std::string::npos == first_dot) {
```
r0busta: That's an interesting expression order :) I couldn't find anything on that in the guidelines we follow but it feels unnatural to see a binary expression ordered as `constant == variable` vs the more common `variable == constant`, e. g. `nullptr != my_ptr` vs. `my_ptr != nullptr`. The latter is more common than the former, isn't it?.

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +101,70 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
...
+                        "Cannot parse ClientId. Incomplete token format");
+  }
+
+  return result;
+}
+
+std::string SignInResultImpl::ParseClientIdFromToken(const std::string& token) {
+  if (token.empty()) {
+    return std::string();
+  }
```
r0busta: Log an error or a debug message? This case shouldn't happen because we parse the access token at line 60.

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +101,79 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
...
+}
+
+std::string SignInResultImpl::ParseClientIdFromToken(const std::string& token) {
+  if (token.empty()) {
+    OLP_SDK_LOG_ERROR_F(kLogTag, "Token is empty");
+    return std::string();
+  }
+
+  if (IsJwtToken(token)) {
+    // This is a "JWT Token"
```
r0busta:
```suggestion
```
Redundant comment

olp-cpp-sdk-authentication/src/SignInResultImpl.h
```diff
@@ -79,11 +86,17 @@ class SignInResultImpl : public BaseResult {
   const std::string& GetScope() const;
 
  private:
+  static std::string ParseJwtToken(const std::string&);
+  static std::string ParseClientIdFromToken(const std::string& token);
+  static bool IsJwtToken(const std::string& token);
+  static bool IsHnToken(const std::string& token);
```
r0busta: Better to make these as free inline functions and move them in the cpp?

olp-cpp-sdk-authentication/tests/SignInResultImplTest.cpp
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include <gtest/gtest.h>
+#include <olp/core/http/HttpStatusCode.h>
+#include "../src/SignInResultImpl.h"
+
+constexpr auto kClientId = "4GaUh8X9CpOax4CCC3of";
```
borzun: Put everything into anonymous namespace

olp-cpp-sdk-authentication/src/SignInResultImpl.cpp
```diff
@@ -91,5 +111,70 @@ const std::string& SignInResultImpl::GetScope() const { return scope_; }
 
 bool SignInResultImpl::IsValid() const { return is_valid_; }
 
+std::string SignInResultImpl::ParseJwtToken(const std::string& token) {
+  const std::string::size_type first_dot = token.find_first_of('.');
+  if (first_dot == std::string::npos) {
+    OLP_SDK_LOG_ERROR_F(kLogTag, "Cannot parse ClientId. Wrong token format.");
```
borzun:
```suggestion
    OLP_SDK_LOG_ERROR(kLogTag, "Cannot parse ClientId. Wrong token format.");
```
Not critical, but no need to add formatted logging. Same with other OLP_SDK_LOG... in this file

PR #548, opened 2020-01-03
Models for Stream Layer added

mykolamalik:
Models were generated using Swagger from OLP Stream API v2 and adapted
to be consistent with existing models for other layers.

All Doxygen documentation corresponds to OLP Stream API v2 descriptions.

Original OLP Stream API v2 models CommitOffsetsRequest and
SeekOffsetsRequest are represented in the code by single OffsetsRequest
model to reduce code duplication because they are completely the same.

Relates-To: OLPEDGE-1196

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/generated/model/ConsumeDataResponse.h
+++ b/olp-cpp-sdk-dataservice-read/src/generated/model/ConsumeDataResponse.h
@@ -1,0 +1,155 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-dataservice-read/src/generated/model/ConsumeDataResponse.h
```diff
@@ -0,0 +1,155 @@
...
+#include <boost/optional.hpp>
+#include <olp/dataservice/read/model/Data.h>
+#include "StreamOffset.h"
+
+namespace olp {
+namespace dataservice {
+namespace read {
+namespace model {
+
+class Metadata {
```
mykhailo-kuchma: DATASERVICE_READ_API required, and other classes below.

r0busta:
Would it make sense to keep with the widely-accepted convention of using imperative in commit messages, e.g. "Fix bug" and not "Fixed bug"? This convention matches up with commit messages
generated by commands like `git merge` and `git revert`.

mykolamalik:
Closed due to splitting PR into couple smaller ones - one model per one PR.

PR #552, opened 2020-01-06
Add unit tests for olp::logging::Configuration class

r0busta:
# Add unit tests for olp::logging::Configuration class

Introducing MockAppender that will be used for testing
this and other logging classes.

Relates-To: OLPEDGE-1237

Signed-off-by: Kirill Zhuchkov <643378+r0busta@users.noreply.github.com>

### DIFF
--- a/olp-cpp-sdk-core/tests/logging/ConfigurationTest.cpp
+++ b/olp-cpp-sdk-core/tests/logging/ConfigurationTest.cpp
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include <gtest/gtest.h>
+#include <olp/core/logging/Configuration.h>
+#include <olp/core/logging/Log.h>
+#include <memory>
+#include "MockAppender.h"
+
+namespace {
+
+using namespace olp::logging;
+using namespace testing;
+
+TEST(ConfigurationTest, InvalidConfiguration) {
+  Configuration configuration;
+  EXPECT_FALSE(configuration.isValid());
+  EXPECT_FALSE(Log::configure(configuration));
+  OLP_SDK_LOG_TRACE("My tag", "My message");
+}
+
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-core/tests/logging/ConfigurationTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+TEST(ConfigurationTest, MockAppender) {
+  Configuration configuration;
+  configuration.addAppender(std::make_shared<MockAppender>());
+  EXPECT_TRUE(configuration.isValid());
+  EXPECT_TRUE(Log::configure(configuration));
+}
+
+}  // namespace
```
andescu: Fix LF issue. For below files also.

olp-cpp-sdk-core/tests/logging/ConfigurationTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+TEST(ConfigurationTest, DefaultConfiguration) {
+  Configuration configuration = Configuration::createDefault();
+  EXPECT_TRUE(configuration.isValid());
+  EXPECT_TRUE(Log::configure(configuration));
+}
+
+TEST(ConfigurationTest, MockAppender) {
+  Configuration configuration;
+  configuration.addAppender(std::make_shared<MockAppender>());
```
andescu: Would it not bet better to mock appender with gmock EXPECT_CALLs and realy check here that the Append() method is called when log message is written?

olp-cpp-sdk-core/tests/logging/ConfigurationTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace {
+
+using namespace olp::logging;
+using namespace testing;
+
+TEST(ConfigurationTest, InvalidConfiguration) {
+  Configuration configuration;
+  EXPECT_FALSE(configuration.isValid());
+  EXPECT_FALSE(Log::configure(configuration));
```
andescu: Maybe also trigger a log so that we make sure nothing crashes.

olp-cpp-sdk-core/tests/logging/ConfigurationTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+TEST(ConfigurationTest, DefaultConfiguration) {
+  Configuration configuration = Configuration::createDefault();
+  EXPECT_TRUE(configuration.isValid());
+  EXPECT_TRUE(Log::configure(configuration));
```
andescu: Same here, trigger a log so that we make sure it is working.

olp-cpp-sdk-core/tests/logging/ConfigurationTest.cpp
r0busta: The approach is rather simple - to access and to check appended messages directly after invoking logging. That should be simpler/cleaner than the gmock approach.

r0busta: Also +1 to this, all the tests, that are imported, use the MockAppender

r0busta: _Resolved offline._
Agreed that we stick with the MockAppender class and in the future, if the benefits will overweight the effort, we'll switch to using `gmock` in the `logging` tests

PR #553, opened 2020-01-06
Improve android HttpClient error handling

tfalencarhere:
- Provide consistency between Android and iOS error handling:
  In iOS, offline error code is already correctly translated; For Android, the most similar error given in such situation is java.net.UnknownHostException.

Relates-To: IOTSDK-3924
Signed-off-by: Thiago F. Alencar <thiago.fariaalencar@here.com>

### DIFF
--- a/olp-cpp-sdk-core/src/http/android/HttpClient.java
+++ b/olp-cpp-sdk-core/src/http/android/HttpClient.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.InetSocketAddress;
+import java.net.MalformedURLException;
 import java.net.ProtocolException;
 import java.net.Proxy;
 import java.net.URL;
@@ -40,13 +41,16 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
+import javax.net.ssl.SSLException;
+
 public class HttpClient {
   private static final String LOGTAG = "HttpClient";
 
   // The replica of http/NetworkType.h error codes
-  // TODO: retrieve them via native method
   public static final int IO_ERROR = -1;
+  public static final int AUTHORIZATION_ERROR = -2;
   public static final int INVALID_URL_ERROR = -3;
+  public static final int OFFLINE_ERROR = -4;
   public static final int CANCELLED_ERROR = -5;
   public static final int TIMEOUT_ERROR = -7;
   public static final int THREAD_POOL_SIZE = 8;
@@ -438,18 +442,22 @@ public class HttpClient {
             isDone = true;
             completeRequest(request.requestId(), status, error, contentType);
           } while (!isDone);
+        } catch (SSLException e) {
+          completeRequest(request.requestId(), AUTHORIZATION_ERROR, "SSL connection failed.", "");
+        } catch (MalformedURLException e) {
+          completeRequest(
+              request.requestId(), INVALID_URL_ERROR, "The provided URL is not valid.", "");
         } catch (OperationCanceledException e) {
           completeRequest(request.requestId(), CANCELLED_ERROR, "Cancelled", "");
         } catch (SocketTimeoutException e) {
           completeRequest(request.requestId(), TIMEOUT_ERROR, "Timed out", "");
+        } catch (java.net.UnknownHostException e) {
+          completeRequest(
+              request.requestId(), OFFLINE_ERROR, "The device has no internet connectivity", "");
         } catch (Exception e) {
           Log.e(LOGTAG, "HttpClient::HttpTask::run exception: " + e);
           e.printStackTrace();
-          if (e instanceof UnknownHostException) {
-            completeRequest(request.requestId(), INVALID_URL_ERROR, "Unknown Host", "");
-          } else {
-            completeRequest(request.requestId(), IO_ERROR, e.toString(), "");
-          }
+          completeRequest(request.requestId(), IO_ERROR, e.toString(), "");
         } finally {
           cleanup(httpConn);
         }


r0busta:
Could you please split the PR into two? `Applied clang-format to the whole file` and `Provide consistency between Android and iOS error handling`?

Even though the PR is not large, it's still better not to mix the bulk code re-format changes with the logic changes.

Thanks

olp-cpp-sdk-core/src/http/android/HttpClient.java
```diff
@@ -329,12 +333,12 @@ protected Void doInBackground(Request... requests) {
...
-                        || (status >= HttpURLConnection.HTTP_SERVER_ERROR))) {
-                  if (retryCount++ < request.maxRetries()) {
-                    continue;
-                  }
+                if ((status == HttpURLConnection.HTTP_GATEWAY_TIMEOUT
+                        || status == HttpURLConnection.HTTP_BAD_GATEWAY
+                        || status == HttpURLConnection.HTTP_INTERNAL_ERROR
+                        || status == HttpURLConnection.HTTP_UNAVAILABLE)
+                    && retryCount++ < request.maxRetries()) {
+                  continue;
```
borzun: What is the reason about changing the retry mechanism? I see that in `NetworkCurl` we also use the old retry mechanism, thus, your changes will make the `NetworkAndroid` retry logic inconsistent with other implementations.

olp-cpp-sdk-core/src/http/android/HttpClient.java
```diff
@@ -40,13 +41,16 @@
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
+import javax.net.ssl.SSLException;
+
 public class HttpClient {
   private static final String LOGTAG = "HttpClient";
 
   // The replica of http/NetworkType.h error codes
-  // TODO: retrieve them via native method
```
borzun: What is the reason about removing this TODO comment?

tfalencarhere: There is no ticket or issue number attached to it. I also think it doesn't make sense because this class needs to know about the error codes available upfront to handle each situation properly, so I see  no value in loading them 'dynamically' if thats what the comment was about. Or did I miss something?

olp-cpp-sdk-core/src/http/android/HttpClient.java
tfalencarhere: What do you mean with 'old retry mechanism'? As far as I see I didn't change the mechanism, as it will still do the "retrying". Here's what I did and my intention behind it:

- Removed the higher than zero check because it is redundant (all http codes are positive)
- Changed the status codes to replace the deprecated one (HTTP_SERVER_ERROR "should have never existed" according to android's own documentation: https://developer.android.com/reference/java/net/HttpURLConnection.html#HTTP_SERVER_ERROR) - so its replacement is HTTP_INTERNAL_ERROR.
- Finally, from my understanding when reading this code, the author's original intention was to retry the connection for failures that are potentially temporary and on the server side (as it makes no sense for example to retry if the device has no internet connection), so I made it more explicit by adding such error codes.

Though to be honest, I would vote to remove the retry mechanism altogether in a separate patch since:
- iOS implementation does not contain it.
- IMO retrying requests should not be done at this level at all.
- It could be that there is a specific case, for curl-based clients , that retrying could make sense, but then I would like to know when that is the case, if possible.

Otherwise, please let me know if I'm missing anything.

borzun: Thanks for the detailed comment.

> What do you mean with 'old retry mechanism'?

As for the 'old retry mechanism' I meant the retry mechanism that is currently in SDK (i.e. previous version of the retry logic relative to your PR).

My concern regarding your changes is that you slightly changes current retry logic - as far as I see, you omitted 2 HTTP error codes:
- https://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html#HTTP_NOT_IMPLEMENTED ;
- https://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html#HTTP_VERSION ;

Even though, a client most probably will not be interested in retrying on these error codes, but still your change is not 'backward-compatible' with current retry mechanism. So, if we agree to apply your changes in HttpClient.java, we have to change retry logic in other implementations, like `NetworkCurl` (https://github.com/heremaps/here-olp-sdk-cpp/blob/master/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp#L937). So, it would be perfect if you apply those changes in this PR as well, otherwise we will handle the `NetworkCurl` by ourselves.

> IMO retrying requests should not be done at this level at all.

We have some SPIKE about reconsider current retry mechanism, so, the retry logic might change in the future release(s).

olp-cpp-sdk-core/src/http/android/HttpClient.java
borzun: > so I see no value in loading them 'dynamically' 

The idea about this TODO was to avoid duplication of error codes in 2 places. Thus, to avoid a problem when we change something in C++ and forget to do this in Java side.

tfalencarhere: Sorry, I still don't understand: Is the TODO here to signal that one should keep the error codes in sync, or is it as a future feature where one would "fetch" from either C++ or Java?
For the former, we already have the following line: "The replica of http/NetworkType.h error codes". For the second, I really don't see a way to avoid the duplication, can you please elaborate? If you'd like to keep this line we should at least rephrase it because its hard to understand as is.
Also, I recommend having tickets assigned to the TODO, otherwise they will just stay there forever.

olp-cpp-sdk-core/src/http/android/HttpClient.java
tfalencarhere: Okay, I see your point. So when you say backward compatibility you mean consistent behaviour across different implementations, and not backward compatibility in android (a change here (android) should have no impact whatsoever for users of the curl implementation - though, I see value in making the behaviour consistent across all implementations).

I will remove these changes, since it is not critical to the improvement I need on this patch.

Please note though that the following (current implementation):

((status > 0) && ((status < 200)|| (status >= 500)))

doesn't provide much value since:
Http status codes at 1XX range are "informational", and other error codes such as 501 (not implemented), 511, etc, cannot be recovered by simply retrying.

borzun: Ok, we will address your suggested changes about retry logic in a separate ticket.

olp-cpp-sdk-core/src/http/android/HttpClient.java
borzun: Agree, let's remove for now

PR #563, opened 2020-01-10
Fix the doxygen for DataRequest, added preconditions for GetData method

mykhailo-kuchma:
Resolves: OLPEDGE-1332

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
@@ -33,11 +33,11 @@ namespace read {
 
 /**
  * @brief The DataRequest class encapsulates the fields required to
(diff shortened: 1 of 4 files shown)


r0busta:
I suggest reflecting the major purpose of the change in the commit message by changing the subject to:
```
Add preconditions for GetData method
```

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
```diff
@@ -33,11 +33,11 @@ namespace read {
```
andescu: Can you maybe also adapt DataRequest to coding style and remove layer_id_ as it is not used anymore.

mykhailo-kuchma: Done

PR #573, opened 2020-01-14
Add SubscribeResponse model for Stream API

mykolamalik:
Holds subscription info to use to pass further requests to the OLP
stream layer.

Relates-To: OLPEDGE-1373

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
+++ b/olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#include <string>
+
+namespace olp {
+namespace dataservice {
+namespace read {
+namespace model {
+
+class SubscribeResponse final {
+ public:
+  const std::string& GetNodeBaseURL() const {
+    return node_base_url_;
+  }
+  void SetNodeBaseURL(std::string value) { node_base_url_ = std::move(value); }
+
+  const std::string& GetSubscriptionId() const {
+    return subscription_id_;
+  }
+  void SetSubscriptionId(std::string value) {
+    subscription_id_ = std::move(value);
+  }
+
+ private:
+  /// A subscription is created in a specific server instance. This base URL
+  /// represents the specific server node on which this subscription was
+  /// created. This base URL should be used instead of the base URL returned by
+  /// the `lookup` service when making further calls to StreamApi methods.
+  std::string node_base_url_;
+
+  /// A unique identifier of the subscription. It is used to make further calls
+  /// to StreamApi methods.
+  std::string subscription_id_;
+};
+
+}  // namespace model
+}  // namespace read
+}  // namespace dataservice
+}  // namespace olp
--- a/olp-cpp-sdk-dataservice-read/src/generated/parser/SubscribeResponseParser.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/generated/parser/SubscribeResponseParser.cpp
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
```diff
@@ -0,0 +1,73 @@
...
+    x_correlation_id = std::move(value);
+  }
+
+ private:
+  /// A subscription/Kafka Consumer is created in a specific server instance.
+  /// This base URL represents the specific server node on which this
+  /// subscription/Kafka Consumer was created. This base URL should be used
+  /// instead of the base URL returned by the `lookup` service when making
+  /// requests to these endpoints: /partitions, /offsets, /seek and /subscribe
+  /// (DELETE).
```
andescu: Please remove Kafka from the comment.
Also remove ```(DELETE)``` as it is very strange and doesn't relate much to the C++ API.

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
```diff
@@ -0,0 +1,73 @@
...
+  boost::optional<std::string> node_base_url_;
```
andescu: It doesn't seem optional to me. Can we do anything without this node_base_url_?
I think not. If it is empty which URL should we use?

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
```diff
@@ -0,0 +1,73 @@
...
+
+  /// A unique identifier of the subscription. It is used to make requests to
+  /// these endpoints: /partitions, /offsets, /seek and /subscribe (DELETE).
+  boost::optional<std::string> subscription_id_;
```
andescu: Also probably not optional.

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
```diff
@@ -0,0 +1,73 @@
...
+
+  /// A tracing ID (correlation-id) to use in subsequent requests. This ID
+  /// should be obtained from HTTP response header.
+  boost::optional<std::string> x_correlation_id;
```
andescu: I dont think we need to add this here. This represent more the body response and should not contain anything else. The X-Correlation-Id should be a separate output to the StreamApi::Subscribe() method.

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
mykolamalik: (DELETE) means the HTTP methods here.

We have two /subscribe endpoints:
1. For initiate subscription /subscribe (POST)
2. For delete subscription /subscribe (DELETE)

And it has a lot of sense, because we really need SubscribeResponse only for one version of /subscribe endpoint (DELETE).

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
mykolamalik: Logically you're right. But this field is not marked as mandatory in OLP API.
Should I remove optional for it?

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
mykolamalik: The same case as for node_base_url_

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
mykolamalik: We discussed it during regarding Stream Layer POC. I thought we agreed to add this field here.
Will add one more parameter to StreamApi::Subscribe() for X-Correlation-Id output.

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
andescu: Yes, please.

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
andescu: Not sure we agreed on this. It doesn't look to me like it is part of the response.
I would rather prefer to put X-Correlation-Id separate as it is more work needed to extract it.

olp-cpp-sdk-dataservice-read/src/generated/model/SubscribeResponse.h
andescu: If you need it then you need to reference the StreamApi methods so that users of the SDK know what to do with this information. If I read your comment as is, my next question would be how can I access /seek or /partitions?

mykolamalik: This is private model, so users of the SDK has no access to it. We will use this model internally in StreamLayerClientImpl.
So only SDK developers can see these comments. And they should have no problems with access to /seek or /partitions.

andescu: Except it is not /seek but StreamApi::Seek() and not /partitions but StreamApi::Partitions() or StreamApi::Consume() rather. So even for developers it is not so straight fwd to figure this out. Especially for those that do not have any experience with the StreamLayerClient or StreamApi.

PR #585, opened 2020-01-17
Create stale.yml

r0busta:
Add automation workflow that helps to mark issues and PRs as stale.

Doesn't need a ticket as it's purely a housekeeping feature that doesn't affect the software 🤓


r0busta:
> Does it not require some kind of app to be enabled for this? E.g. https://github.com/marketplace/stale

Indeed.
Closing the PR as it will take ages to approve the app installed in our repository🐌

PR #586, opened 2020-01-20
Adapts DiskCache class to coding style.

andescu:
DiskCache was not fully compliant with the coding style.
This commit applies the coding style on this class with several
paramters name chaning and methods renaming, including public API.
Class is private and only visible to DefaultCache so no API break expected.
Also improved the doxygen comments.

Relates-to: OLPEDGE-566

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DiskCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/DiskCache.cpp
@@ -40,108 +40,105 @@ namespace olp {
 namespace cache {
 
 namespace {
-
 constexpr auto kLogTag = "Storage.LevelDB";
 
-static inline leveldb::Slice toLeveldbSlice(const std::string& slice) {
+leveldb::Slice ToLeveldbSlice(const std::string& slice) {
   return leveldb::Slice(slice);
 }
 
-std::vector<std::string> tokenizePath(const std::string& dirName,
+std::vector<std::string> TokenizePath(const std::string& directory_name,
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -153,112 +150,116 @@ bool DiskCache::Clear() {
...
-  if (environment_ && (max_size_ != std::uint64_t(-1))) {
-    if (environment_->Size() >= max_size_) return false;
+  if (environment_ && (max_size_ != kSizeMax) &&
+      (environment_->Size() >= max_size_)) {
+    return false;
   }
 
-  const leveldb::Status& status = database_->Put(
-      leveldb::WriteOptions(), toLeveldbSlice(key), toLeveldbSlice(value));
+  const auto status = database_->Put(
```
r0busta:
```suggestion
  const auto& status = database_->Put(
```

andescu: Not sure why it was like that as the leveldb::Put() method has the following signature:

```cpp
Status DB::Put(const WriteOptions& opt, const Slice& key, const Slice& value)
```

So theoretically we used a ref to a temp variable before if the compiler did not optimized it out.

r0busta: Ok. Either should be fine.

PR #587, opened 2020-01-20
Adapt DefaultCache and other cache files to coding style.

andescu:
KeyValueCache is now optimized with some aliases to be also used above
by parent class and users. DefaultCache is now marking overriden methods
from KeyValueCache with the override keyword. DiskCacheSizeLimitEnv and
DiskCacheSizeLimitWritableFile are also adapted where needed to coding style.

Relates-to: OLPEDGE-566

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCache.cpp
@@ -39,7 +39,7 @@ std::string CreateExpiryKey(const std::string& key) { return key + "::expiry"; }
 time_t GetRemainingExpiryTime(const std::string& key,
                               olp::cache::DiskCache& disk_cache) {
   auto expiry_key = CreateExpiryKey(key);
-  auto expiry = (std::numeric_limits<time_t>::max)();
+  auto expiry = olp::cache::KeyValueCache::kDefaultExpiry;
   auto expiry_value = disk_cache.Get(expiry_key);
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -332,10 +332,9 @@ boost::optional<std::pair<std::string, time_t>> DefaultCache::GetFromDiscCache(
     const std::string& key) {
   if (protected_cache_) {
     auto result = protected_cache_->Get(key);
-
     if (result) {
-      return std::make_pair(std::move(result.value()),
-                            std::numeric_limits<time_t>::max());
+      auto default_expiry = KeyValueCache::kDefaultExpiry;
+      return std::make_pair(std::move(result.value()), default_expiry);
```
r0busta: Why not `return std::make_pair(std::move(result.value()), KeyValueCache::kDefaultExpiry);`?

andescu: As discussed, this is not possible with std::make_pair() taking rvalue parameters. Because of the `std::move(result.value())` the second parameter is also expected to be an rvalue that cannot apply to constexpr.

r0busta: Thanks for the explanation.

PR #588, opened 2020-01-21
Add CommitOffsets and SeekToOffset public methods to the read StreamApi

mykolamalik:
Will be used by StreamLayerClient to interact with OLP Stream Layer.

The only difference of these methods is the URI endpoint. So the logic
is extracted to one private method.

Relates-To: OLPEDGE-1196

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/generated/api/StreamApi.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/generated/api/StreamApi.cpp
@@ -30,6 +30,7 @@
 #include "generated/parser/SubscribeResponseParser.h"
 #include <olp/core/generated/parser/JsonParser.h>
 #include "generated/serializer/ConsumerPropertiesSerializer.h"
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-dataservice-read/src/generated/api/StreamApi.cpp
```diff
@@ -92,26 +92,46 @@ StreamApi::SubscribeApiResponse StreamApi::Subscribe(
...
+  query_params.emplace("subscriptionId", subscription_id);
 
-  query_params.insert(std::make_pair("mode", mode));
+  query_params.emplace("mode", mode);
 
   std::multimap<std::string, std::string> header_params;
-  header_params.insert(std::make_pair("Accept", "application/json"));
+  header_params.emplace("Accept", "application/json");
+
   if (x_correlation_id) {
```
andescu: This check is not valid -> remove.

PR #594, opened 2020-01-23
Adapts HRN, CancellationContext and Condition to coding style.

andescu:
This commit adapt the above mentioned classes the the coding style,
also marks the public members of HRN as deprecated and to be
scoped to private by 05.2020 as it is not recommended by HERE Platform
to interpret the HRN in any way. Users should only use the public methods
available and not care about the composition of the HRN string.

Relates-to: OLPEDGE-567

### DIFF
--- a/olp-cpp-sdk-core/src/client/HRN.cpp
+++ b/olp-cpp-sdk-core/src/client/HRN.cpp
@@ -19,84 +19,102 @@
 
 #include "olp/core/client/HRN.h"
 
-#include <cassert>
 #include <sstream>
 
+#include <olp/core/logging/Log.h>
+#include <olp/core/porting/make_unique.h>
 #include "Tokenizer.h"
+
+namespace {
+constexpr auto kLogTag = "HRN";
+static const std::string kDataTag = "data";
+static const std::string kPipelineTag = "pipeline";
+static const std::string kSchemaTag = "schema";
+static const std::string kHRNTag = "hrn:";
+static constexpr char kSeparator = ':';
+}  // namespace
 
 namespace olp {
 namespace client {
 
 std::string HRN::ToString() const {
   std::ostringstream ret;
-  std::string generic_part = ":" + region + ":" + account + ":";
-  ret << "hrn:" << partition << ":";
+  const auto generic_part =
+      kSeparator + region + kSeparator + account + kSeparator;
+  ret << kHRNTag << partition << kSeparator;
+
   switch (service) {
     case ServiceType::Data: {
-      ret << "data" << generic_part << catalogId;
+      ret << kDataTag << generic_part << catalogId;
       if (!layerId.empty()) {
-        ret << ":" << layerId;
+        ret << kSeparator << layerId;
       }
       break;
     }
-    case ServiceType::Schema:
-      ret << "schema" << generic_part << groupId << ":" << schemaName << ":"
-          << version;
+    case ServiceType::Schema: {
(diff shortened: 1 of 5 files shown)


olp-cpp-sdk-core/src/client/HRN.cpp
```diff
@@ -19,84 +19,101 @@
...
+    }
   }
+
   return ret.str();
 }
 
 std::string HRN::ToCatalogHRNString() const {
   if (service != ServiceType::Data) {
-    return std::string();
+    OLP_SDK_LOG_WARNING_F(kLogTag, "ToCatalogHRNString: ServiceType != Data");
```
borzun:
```suggestion
    OLP_SDK_LOG_WARNING(kLogTag, "ToCatalogHRNString: ServiceType != Data");
```
Minor - you can use logging without format

olp-cpp-sdk-core/src/client/HRN.cpp
```diff
@@ -19,84 +19,101 @@
...
+}  // namespace
+
 namespace olp {
 namespace client {
 
 std::string HRN::ToString() const {
   std::ostringstream ret;
-  std::string generic_part = ":" + region + ":" + account + ":";
-  ret << "hrn:" << partition << ":";
+  auto generic_part = kSeparator + region + kSeparator + account + kSeparator;
```
borzun:
```suggestion
  const auto generic_part = kSeparator + region + kSeparator + account + kSeparator;
```
Prefer declaring the variable as const by default, unless there is a reason to change it (Con1: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-immutable)

olp-cpp-sdk-core/src/client/HRN.cpp
```diff
@@ -19,84 +19,101 @@
...
 
-  // hrn must start with "hrn:"
+  // Must start with "hrn:"
   if (!tokenizer.HasNext()) {
     return;
   }
 
-  std::string protocol = tokenizer.Next();
-
+  auto protocol = tokenizer.Next();
```
borzun:
```suggestion
  const auto protocol = tokenizer.Next();
```
Prefer immutable objects by default

olp-cpp-sdk-core/src/client/HRN.cpp
```diff
@@ -19,84 +19,101 @@
...
 std::string HRN::ToCatalogHRNString() const {
   if (service != ServiceType::Data) {
-    return std::string();
+    OLP_SDK_LOG_WARNING_F(kLogTag, "ToCatalogHRNString: ServiceType != Data");
+    return {};
   }
 
-  return "hrn:" + partition + ":data:" + region + ":" + account + ":" +
-         catalogId;
+  return kHRNTag + partition + ":data:" + region + kSeparator + account +
```
borzun:
```suggestion
  return kHRNTag + partition + kSeparator + kDataTag + kSeparator + region + kSeparator + account +
```
You can use `kDataTag` here.

olp-cpp-sdk-core/src/client/HRN.cpp
andescu: If it is ok I would like to keep consistent within one file as I use printf style below in constructor also.

borzun: ok, I'm fine with that approach as well.

PR #596, opened 2020-01-23
Move the OlpClientTest closer to its relatives

r0busta:
# Move the OlpClientTest closer to its relatives

The test file should be in the folder where it's expected to be found.

Relates-To: OLPEDGE-1398


r0busta:
> The test folders match the class folders and it will create more confusion if we change that.

The thing is that the OlpClient is in the `client` folder, not `olpclient`. So it's a fix actually to the `test folders match the class folders` ;)

PR #598, opened 2020-01-24
Add a cache-dumper utility

haraldF:
leveldbutil isn't really up to the task to introspect our cache (too
low-level), so I wrote a little tool that helps to understand the
persistent cache. It either dumps all keys or a specific value.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,5 +95,9 @@ if(OLP_SDK_BUILD_EXAMPLES)
     add_subdirectory(examples)
 endif()
 
+if(NOT CMAKE_CROSSCOMPILING)
+    add_subdirectory(tools/cache-dumper)
+endif()
+
 # Add uninstall script
 add_custom_target(uninstall "${CMAKE_COMMAND}" -P "${CMAKE_MODULE_PATH}/uninstall.cmake")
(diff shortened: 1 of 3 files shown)


r0busta:
@haraldF, thank you for your contribution 👍

I'm not sure though, that the SDK repository is the best place for such a tool. And it's not to mention that across GitHub, there are a few similar tools written in various languages:

 * https://github.com/tgulacsi/leveldb-tools (Go)
 * https://github.com/thlorenz/level-dump (JavaScript)
 * https://github.com/sel-fish/leveldb-dump (C++)

Do you think it'd make sense to continue with this PR?

haraldF:
> Do you think it'd make sense to continue with this PR?

The problem with those tools is that they don't know our DB structure, e.g. they dump things as hex which makes it tricky to debug. Our tool can use our DB structure to show e.g. tile ranges if we choose to extend it.

The JavaScript tool I couldn't get to work (compilation errors, probably due to me using a newer nodejs). Go one I didn't try (no Go compiler installed).

Since it's tightly coupled to the SDK releases (and it's not installed, it's just an internal developer tool), I would leave it in this repository. However, it's you call, feel free to abandon my request and I'll keep the tool for myself for playing around :)

PR #612, opened 2020-01-29
Add integration tests for StreamLayerClient GetData

dpoberezh:
Resolves: OLPEDGE-1493


tests/integration/olp-cpp-sdk-dataservice-read/StreamLayerClientTest.cpp
```diff
@@ -33,9 +33,15 @@ using namespace olp::dataservice::read;
 using namespace olp::tests::common;
 using namespace testing;
 
+const std::string kBlobData =
```
andescu: Why do you need an additional variable?
Are you not able to create a temp one from HTTP_RESPONSE_BLOB_DATA_STREAM_MESSAGE?

dpoberezh: Fixed. I just thought to not mix the blob data and the response together. But it is probably not an issue.

PR #616, opened 2020-01-30
Use system clock time as a fallback

meravingen123:
* Use system clock time as a fallback in AuthenticationClient::SignInClient()
   request when GetTimeFromServer() call fails.

Resolves: OLPEDGE-1499

### DIFF
--- a/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
+++ b/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
@@ -278,19 +278,16 @@ client::CancellationToken AuthenticationClient::Impl::SignInClient(
   client::CancellationContext context;
 
   auto time_callback = [=](TimeResponse response) mutable {
-    if (!response.IsSuccessful()) {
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -278,19 +278,16 @@ client::CancellationToken AuthenticationClient::Impl::SignInClient(
...
   auto time_callback = [=](TimeResponse response) mutable {
-    if (!response.IsSuccessful()) {
-      callback(AuthenticationError(
-          static_cast<int>(response.GetError().GetErrorCode()),
-          response.GetError().GetMessage()));
-      return;
-    }
+    // As a fallback use local system time as a timestamp
+    time_t timestamp =
+        response.IsSuccessful() ? response.GetResult() : std::time(nullptr);
```
r0busta: Are you going to add a test for this logic in a follow-up commit?

PR #620, opened 2020-01-30
Add intergation tests for StreamApi::CommitOffsets method

mykolamalik:
Resolves: OLPEDGE-1473

### DIFF
--- a/olp-cpp-sdk-dataservice-read/tests/StreamApiTest.cpp
+++ b/olp-cpp-sdk-dataservice-read/tests/StreamApiTest.cpp
@@ -40,18 +40,14 @@ std::string ApiErrorToString(const ApiError& error) {
 }
 
 model::StreamOffsets GetStreamOffsets() {
-  static model::StreamOffsets offsets;
-
-  if (offsets.GetOffsets().empty()) {
-    model::StreamOffset offset1;
-    offset1.SetPartition(7);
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-dataservice-read/tests/StreamApiTest.cpp
```diff
@@ -39,6 +39,22 @@ std::string ApiErrorToString(const ApiError& error) {
   return result_stream.str();
 }
 
+model::StreamOffsets GetStreamOffsets() {
+  static model::StreamOffsets offsets;
```
andescu: Why does it need to be static?

mykolamalik: To be initialized and filled only once. It will be used for SeekToOffset tests also

andescu: This is just complicating the test. You make static something that needs to have a certain lifecycle or has some limitations that force you to use the singleton pattern. In this scenario, I don't see any need to make it static and this only adds additional confusion.
Is there any limitation that forces you to make it static or do you aim for obscurity?

mykolamalik: Static variable is removed.

PR #623, opened 2020-01-31
Retry unsuccessful HTTP requests on 500 and 503 errors by default.

r0busta:
### Retry unsuccessful HTTP requests on 500 and 503 errors by default.

The error code 500 Internal Error indicates that a server is unable
to handle the request at that time. The error code 503 Slow Down
typically means that the request rate to the server bucket is
too high. Hence, it makes sense to resolve such errors by retrying
the request.

Related-To: OLPEDGE-1398

### DIFF
--- a/olp-cpp-sdk-core/src/client/OlpClientSettings.cpp
+++ b/olp-cpp-sdk-core/src/client/OlpClientSettings.cpp
@@ -19,6 +19,7 @@
 
 #include "olp/core/client/OlpClientSettings.h"
 #include "olp/core/client/HttpResponse.h"
+#include "olp/core/http/HttpStatusCode.h"
 
 namespace olp {
 namespace client {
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/src/client/OlpClientSettings.cpp
```diff
@@ -26,7 +26,15 @@ unsigned int DefaultBackdownPolicy(unsigned int milliseconds) {
   return milliseconds;
 }
 
-bool DefaultRetryCondition(const HttpResponse&) { return false; }
+bool DefaultRetryCondition(const HttpResponse& response) {
+  switch (response.status) {
+    case 500:
+    case 503:
```
mykhailo-kuchma: Please use HttpStatusCode's

PR #626, opened 2020-01-31
Modify documentation in core/include/context

HalynaDumych:
Relates-to: OLPEDGE-1442

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/context/Context.h
+++ b/olp-cpp-sdk-core/include/olp/core/context/Context.h
@@ -33,57 +33,109 @@
 namespace olp {
 namespace context {
 /**
- * @brief The Context class represents the application context
+ * @brief Represents the application context.
  *
- * Client application must initialize its Context before using any other
- * functionality of the library.
+ * Before your application uses any other functionality of the library,
+ * initialize the `Context` class.
  */
-
 struct ContextData;
 
+/**
+ * @brief Used only for the Android environment to correctly initialize
+ * the `NetworkAndroid` class.
+ *
+ * Initialize the `Context` class before you send network requests in
+ * the Android environment.
+ * To use the `Context` class, cerate the `Scope` object first.
+ */
 class CORE_API Context {
  public:
+  /**
+   * @brief Called when the `Context` object is initialized.
+   *
+   * Controlled by the `Scope` class.
+   *
+   * @see `Scope` for more information.
+   */
   using InitializedCallback = std::function<void()>;
+
+  /**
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/include/olp/core/context/Context.h
```diff
@@ -33,57 +33,108 @@
...
  public:
+  /**
+   * @brief Called when the `Context` object is initialized.
+   *
+   * Controlled by the `Scope` class.
+   *
+   * @see `Scope` for more information.
+   */
   using InitializedCallback = std::function<void()>;
+  /**
```
r0busta:
```suggestion

  /**
```

olp-cpp-sdk-core/include/olp/core/context/Context.h
```diff
@@ -33,57 +33,108 @@
...
+   * Instead of calling `Context::init()` and `Context::deinit()` manually,
+   * instantiate the `Context::Scope()` object.
    */
   class CORE_API Scope {
    public:
-    /// see Context::init()
+    /**
+     * @brief Creates the `Scope` instance.
+     *
+     * Used to initialize the `Context` class.
```
r0busta: `<rant>` I see such sentences starting with `Used to ...` all across the SDK and while they might make sense in the library context. They sound weird, and I'm not even sure that such a sentence is grammatically correct.`</rant>`
What do you think should fix it some other day?

HalynaDumych: In the descriptions, it is a common practice not to repeat the class/method/function name and not to use phrases like 'this method", "this function", and so on. This sentence is a shorter version of  "This constructor is used to initialize the `Context` class". 

For more information, please see the [API Reference Code Comments](https://developers.google.com/style/api-reference-comments) section in the 
Developer Documentation Style Guide.

HalynaDumych: However, we can put it in the following way: "The `Scope` instance is used to initialize the `Context` class."

olp-cpp-sdk-core/include/olp/core/context/Context.h
HalynaDumych: Added

olp-cpp-sdk-core/include/olp/core/context/Context.h
r0busta: I was just asking a question, but thanks for changing that quick and pro-actively anyway :)

PR #635, opened 2020-02-06
Add unit tests for StreamLayerClient::Poll method

mykolamalik:
Resolves: OLPEDGE-1500

### DIFF
--- a/olp-cpp-sdk-dataservice-read/tests/StreamLayerClientImplTest.cpp
+++ b/olp-cpp-sdk-dataservice-read/tests/StreamLayerClientImplTest.cpp
@@ -93,8 +93,6 @@ constexpr auto kHttpResponsePollCommitConflict =
 constexpr auto kHttpResponseUnsubscribeNotFound =
     R"jsonString({"title":"Subscription not found","status":404,"code":"E213003","cause":"SubscriptionId 12345 not found","action":"Subscribe again","correlationId":"123"})jsonString";
 
-const std::string kCatalog =
-    "hrn:here:data::olp-here-test:hereos-internal-test-v2";
 const std::string kConsumerID = "consumer_id_1234";
 const ConsumerProperties kConsumerProperties = {
     {"key1", "value1"}, {"key2", 10}, {"key3", true}};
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-read/tests/StreamLayerClientImplTest.cpp
```diff
@@ -104,11 +135,57 @@ void SetupNetworkExpectation(NetworkMock& network_mock, T url, T response,
...
+          .WillOnce(ReturnHttpResponse(
+              http::NetworkResponse().WithStatus(status), response));
+      return;
+    }
     default:
       return;
   }
 }
 
+HRN GetTestCatalog() { return HRN::FromString(kCatalog); }
```
andescu: One-liners are usually creating more confusion than helping.

olp-cpp-sdk-dataservice-read/tests/StreamLayerClientImplTest.cpp
```diff
@@ -139,17 +216,16 @@ TEST(StreamLayerClientImplTest, Subscribe) {
     ASSERT_EQ(future.wait_for(kTimeout), std::future_status::ready);
 
     const auto& response = future.get();
-    ASSERT_TRUE(response.IsSuccessful());
+    EXPECT_TRUE(response.IsSuccessful());
```
andescu: This was for a reason so that you don't continue and possibly crash as there is no result to get.
Why did you change to expect?

olp-cpp-sdk-dataservice-read/tests/StreamLayerClientImplTest.cpp
```diff
@@ -718,38 +703,309 @@ TEST(StreamLayerClientImplTest, GetDataCancel) {
   EXPECT_EQ(response.GetError().GetErrorCode(), ErrorCode::Cancelled);
 }
 
+TEST(StreamLayerClientImplTest, Poll) {
+  auto network_mock = std::make_shared<NetworkMock>();
+  auto cache_mock = std::make_shared<CacheMock>();
+  OlpClientSettings settings;
+  settings.network_request_handler = network_mock;
+  settings.cache = cache_mock;
```
andescu: Now after 10+ test cases I don't think it was worth it to use TEST instead of TEST_F with setting up this entire stuff once?
Please change and create a simple fixture that sets up this entire stuff once and possibly also creates a StreamLayerClient for you.

olp-cpp-sdk-dataservice-read/tests/StreamLayerClientImplTest.cpp
mykolamalik: It would be great to use assert here, but the problem is we write one test per one API method and use Scoped Traces for actually different tests. And that is why we should use ASSERT only where we can not use EXPECT, because in other case next scoped parts of the test will not run.

In this particular and similar cases there is no problem with EXPECT, because ApiResponse has member of ResultType in any case, even if IsSuccessful()==false. If ApiResponse contains an error, GetResult() will return default constructed ResultType value. And further checks will be focused on this default constructed result value, and this is ok.

So we can use EXPECT here and so we should us it to increase chances other sub-tests to run.

andescu: Usually, this should not be so black and white on unit tests as you should only be able to break this during development if you really work on the tested method. I would assume this to be true on integration test.

PR #640, opened 2020-02-06
Modify CallApi to return HTTP headers.

LiubovDidkivska:
Add headers to http responce. 

Resolves: OLPEDGE-1348

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
@@ -20,7 +20,6 @@
 #pragma once
 
 #include <sstream>
-#include <vector>
 
 #include <olp/core/CoreApi.h>
 #include <olp/core/http/NetworkTypes.h>
@@ -33,10 +32,6 @@ namespace client {
  */
 class CORE_API HttpResponse {
  public:
-  /**
-   * @brief Type alias for HTTP headers.
-   */
-  using HeadersType = std::vector<std::pair<std::string, std::string>>;
 
   HttpResponse() = default;
   /**
@@ -60,10 +55,10 @@ class CORE_API HttpResponse {
    *
    * @param status The HTTP status.
    * @param response The response body.
-   * @param headers The headers of the responce.
+   * @param headers Response headers.
    */
   HttpResponse(int status, std::stringstream&& response,
-               HeadersType&& headers)
+               http::HeadersType&& headers)
       : status(status),
         response(std::move(response)),
         headers(std::move(headers)) {}
@@ -79,11 +74,12 @@ class CORE_API HttpResponse {
   /**
    * @brief The HTTP response.
    */
+
   std::stringstream response;
   /**
    * @brief HTTP headers.
    */
-  HeadersType headers;
+  http::HeadersType headers;
 };
 
 }  // namespace client
--- a/olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
@@ -20,6 +20,7 @@
 #pragma once
 
 #include <sstream>
+#include <vector>
 
 #include <olp/core/CoreApi.h>
 #include <olp/core/http/NetworkTypes.h>
@@ -32,6 +33,11 @@ namespace client {
  */
 class CORE_API HttpResponse {
  public:
+  /**
+   * @brief Type alias for HTTP headers.
+   */
+  using HeadersType = std::vector<std::pair<std::string, std::string>>;
+
   HttpResponse() = default;
   /**
    * @brief Creates the `HttpResponse` instance.
@@ -49,7 +55,18 @@ class CORE_API HttpResponse {
    */
   HttpResponse(int status, std::stringstream&& response)
       : status(status), response(std::move(response)) {}
-
+  /**
+   * @brief Creates the `HttpResponse` instance.
+   *
+   * @param status The HTTP status.
+   * @param response The response body.
+   * @param headers The headers of the responce.
+   */
+  HttpResponse(int status, std::stringstream&& response,
+               HeadersType&& headers)
+      : status(status),
+        response(std::move(response)),
+        headers(std::move(headers)) {}
   /**
    * @brief The HTTP status.
    *
@@ -63,6 +80,10 @@ class CORE_API HttpResponse {
    * @brief The HTTP response.
    */
   std::stringstream response;
+  /**
+   * @brief HTTP headers.
+   */
+  HeadersType headers;
 };
 
 }  // namespace client
--- a/olp-cpp-sdk-core/include/olp/core/http/NetworkRequest.h
+++ b/olp-cpp-sdk-core/include/olp/core/http/NetworkRequest.h
@@ -37,8 +37,6 @@ class CORE_API NetworkRequest final {
  public:
   /// Short type alias for HTTP request body.
   using RequestBodyType = std::shared_ptr<const std::vector<std::uint8_t>>;
-  /// Short type alias for HTTP request header.
-  using RequestHeadersType = std::vector<std::pair<std::string, std::string>>;
 
   /// HTTP method as in https://tools.ietf.org/html/rfc2616.
   enum class HttpVerb {
@@ -60,7 +58,7 @@ class CORE_API NetworkRequest final {
    * @brief Get all HTTP headers.
    * @return vector of HTTP headers.
    */
-  const RequestHeadersType& GetHeaders() const;
+  const HeadersType& GetHeaders() const;
 
   /**
    * @brief Add extra HTTP header.
@@ -128,7 +126,7 @@ class CORE_API NetworkRequest final {
   /// Request URL.
   std::string url_;
   /// HTTP headers.
-  RequestHeadersType headers_;
+  HeadersType headers_;
   /// Body of HTTP request.
   RequestBodyType body_;
   /// Network settings for this request.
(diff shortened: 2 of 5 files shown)


olp-cpp-sdk-core/include/olp/core/http/NetworkRequest.h
```diff
@@ -37,8 +37,6 @@ class CORE_API NetworkRequest final {
  public:
   /// Short type alias for HTTP request body.
   using RequestBodyType = std::shared_ptr<const std::vector<std::uint8_t>>;
-  /// Short type alias for HTTP request header.
-  using RequestHeadersType = std::vector<std::pair<std::string, std::string>>;
```
andescu: This is an API break. Please avoid changing public defines.
Better just provide an additional define in HttpResponse.

olp-cpp-sdk-dataservice-read/src/generated/api/StreamApi.cpp
```diff
@@ -87,9 +87,10 @@ StreamApi::SubscribeApiResponse StreamApi::Subscribe(
 
   OLP_SDK_LOG_DEBUG_F(kLogTag, "subscribe, uri=%s, status=%d",
                       metadata_uri.c_str(), http_response.status);
-
-  // TODO: Set x_correlation_id to the value received in http_response.header
-  // when http_response.header will be implemented.
+  if (!HandleCorrelationId(http_response.headers, x_correlation_id)) {
+    OLP_SDK_LOG_WARNING(kLogTag,
```
andescu: I dont think this is so tragically to be handles as warning.

LiubovDidkivska: X-Correlation should be in a separate pull request.

olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
```diff
@@ -49,7 +55,18 @@ class CORE_API HttpResponse {
    */
   HttpResponse(int status, std::stringstream&& response)
       : status(status), response(std::move(response)) {}
-
+  /**
+   * @brief Creates the `HttpResponse` instance.
+   *
+   * @param status The HTTP status.
+   * @param response The response body.
+   * @param headers The headers of the responce.
```
r0busta:
```suggestion
   * @param headers Response headers.
```

olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
```diff
@@ -63,6 +80,10 @@ class CORE_API HttpResponse {
    * @brief The HTTP response.
    */
   std::stringstream response;
+  /**
```
r0busta:
```suggestion

  /**
```

olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
```diff
@@ -32,6 +33,11 @@ namespace client {
  */
 class CORE_API HttpResponse {
  public:
+  /**
+   * @brief Type alias for HTTP headers.
+   */
+  using HeadersType = std::vector<std::pair<std::string, std::string>>;
```
r0busta: Would a `map` (or and `unordered_map`) be more helpful and easier to use downstream?

r0busta: Or even better – an `std::multimap` that would reflect the way headers are passed to `OlpClient::CallApi`

r0busta: Apparently, it's a big mess :) 
The `http::NetworkRequest` class has a type `RequestHeadersType` that is a vector.
So I propose to deprecate the `http::NetworkRequest::RequestHeadersType`, add  a common `http:: HeadersType` as `using HeadersType = std:: multimap <std::string, std::string>` instead, and use it consistently.

andescu: I would prefer we use `std::vector<std::pair<std::string, std::string>>` all the way, as it is handier then a multimap and has a better memory footprint.

LiubovDidkivska: HeadersType was updated to multimap on Friday. Please take a look, what commit I should leave, previous with vector, or last one with multimap.

olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
```diff
@@ -128,5 +129,9 @@ class SendOutcome final {
  */
 CORE_API std::string ErrorCodeToString(ErrorCode code);
 
+/**
+ * @brief Type alias for HTTP headers.
+ */
+using HeadersType = std::multimap<std::string, std::string>;
```
andescu: Why multimap? It is basically the same as `vector<pair<string, string>>`, except it consumes additional memory and it keeps elements sorted. Do we need this? Is there any benefit? Is it worth it to change API over it?

olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
```diff
@@ -49,7 +50,18 @@ class CORE_API HttpResponse {
...
-
+  /**
+   * @brief Creates the `HttpResponse` instance.
+   *
+   * @param status The HTTP status.
+   * @param response The response body.
+   * @param headers The headers of the responce.
+   */
+  HttpResponse(int status, std::stringstream&& response,
+               http::HeadersType&& headers)
```
andescu: Maybe we can start moving with this also towards named parameter idiom same as for NetworkRequest.
So that we do not have to put everything in constructor anymore (which btw is an API break!).

olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
r0busta: **TL;DR**
Even though the `vector` of pairs is not as usable as a map, it also okay with me.

**Detailed answer.**
That's my fault. Working with `OlpClient::CallApi` got me thinking that `multipmaps` are the way to do the headers thing around here. So I confused Liubov by asking her to be consistent with the `CallApi` and extract a separate header type using `multimap`.

I also consider maps to be the best choice for storing headers as it's easier to do things such as:
```
// Pseudo-code starts
headers["Authorization"] = "Bearer h1.blah";
client.Get("http://example.com", headers);
// Pseudo-code ends
```

andescu: multimap misses the operator[] as far as I know as it does not have unique keys.

r0busta: That's sad. But on the bright side, it has `find(const Key& key)` which is one step closer to the `std::map::[]` than a vector.

LiubovDidkivska: Even if we leave std::vector, it seems to be good to have one type alias in NetworkTypes.h. What do you think?

r0busta: +1 to having `using HeadersType = std::vector<std::pair<std::string, std::string>>`

olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
LiubovDidkivska: HttpResponse currently has all parameters public, without any setters. In this case I need to change this class completely to use named parameter idiom.

olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
LiubovDidkivska: done

olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
andescu: Ok, let's leave it for now and do it later.

olp-cpp-sdk-core/include/olp/core/http/NetworkRequest.h
```diff
@@ -37,8 +38,8 @@ class CORE_API NetworkRequest final {
...
+  /// Short type alias for HTTP request header, use HeadersType instead
+  using RequestHeadersType  OLP_SDK_DEPRECATED("Use HeadersType instead") = HeadersType;
```
andescu: You don't need to deprecate. Having an additional alias didn't hurt anybody.

olp-cpp-sdk-core/include/olp/core/http/NetworkRequest.h
```diff
@@ -26,6 +26,7 @@
 
 #include <olp/core/CoreApi.h>
 #include <olp/core/http/NetworkSettings.h>
+#include <olp/core/porting/deprecated.h>
```
andescu: Not needed anymore.

olp-cpp-sdk-core/include/olp/core/http/NetworkRequest.h
andescu: I mean we should keep it but use the HeadersType as base:

```cpp
using RequestHeadersType = HeadersType;
```

and then leave the rest of the class untouched.

olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
```diff
@@ -128,5 +129,9 @@ class SendOutcome final {
  */
 CORE_API std::string ErrorCodeToString(ErrorCode code);
 
+/**
+ * @brief Type alias for HTTP headers.
+ */
+using HeadersType = std::vector<std::pair<std::string, std::string>>;
```
r0busta:
```suggestion
using HeadersType = std::vector<std::pair<std::string, std::string>>;

```

PR #646, opened 2020-02-07
Review documentation for `GetExpiresIn`

HalynaDumych:
Relates-to: OLPEDGE-1495

### DIFF
--- a/olp-cpp-sdk-authentication/include/olp/authentication/SignInResult.h
+++ b/olp-cpp-sdk-authentication/include/olp/authentication/SignInResult.h
@@ -103,9 +103,9 @@ class AUTHENTICATION_API SignInResult {
   time_t GetExpiryTime() const;
 
   /**
-   * @brief Gets the access token expiry time in seconds.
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-authentication/include/olp/authentication/SignInResult.h
```diff
@@ -105,7 +105,7 @@ class AUTHENTICATION_API SignInResult {
   /**
    * @brief Gets the access token expiry time in seconds.
    *
-   * @return Duration for which token stays valid.
+   * @return The expiry time of the access token (in seconds).
```
andescu: There was a vivid discussion regarding this in some other PR. The bottom line was that because of the method returning `std::chrono::seconds` no explicit mention of the unit should be necessary for the documentation.
I agree with that.
Same applies for methods that take in `std::chrono::seconds`.

HalynaDumych: I made the necessary changes.

PR #634, opened 2020-02-06
Add new ctor for VersionedLayerClient which accepts catalog version

borzun:
Add new constructor for dataserivce-read::VersionedLayerClient, which in addition to previous ctor's parameters, also accepts catalog version. Mark previous ctor as deprecated.

Relates-to: OLPEDGE-1513

Signed-off-by: Bohdan Kurylovych <ext-bohdan.kurylovych@here.com>

PR #651, opened 2020-02-10
Add version lock to VersionedLayerClient

meravingen123:
Now version of request will be locked on first
call to any of VersionedLayerClient methods that
accept version field in request. Any further
requests with version specified will be ignored.
In case if first request will not contain version
parameter, SDK will request latest and lock to it.

Resolves: OLPEDGE-1515, OLPEDGE-1513

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -39,15 +39,19 @@ namespace read {
 
 namespace {
 constexpr auto kLogTag = "VersionedLayerClientImpl";
+constexpr int64_t kInvalidVersion = -1;
 }  // namespace
 
 VersionedLayerClientImpl::VersionedLayerClientImpl(
     client::HRN catalog, std::string layer_id,
+    boost::optional<int64_t> catalog_version,
     client::OlpClientSettings settings)
     : catalog_(std::move(catalog)),
       layer_id_(std::move(layer_id)),
       settings_(std::move(settings)),
-      pending_requests_(std::make_shared<client::PendingRequests>()) {
+      pending_requests_(std::make_shared<client::PendingRequests>()),
+      catalog_version_(catalog_version ? catalog_version.get()
+                                       : kInvalidVersion) {
   if (!settings_.cache) {
     settings_.cache = client::OlpClientSettingsFactory::CreateDefaultCache({});
   }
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -320,6 +327,40 @@ VersionedLayerClientImpl::PrefetchTiles(PrefetchTilesRequest request) {
                                                           promise);
 }
 
+VersionedLayerClientImpl::VersionResponse VersionedLayerClientImpl::GetVersion(
+    boost::optional<int64_t> requested_version,
```
andescu: Not quite, but close :)
The main idea is that the version is locked either by the user in the constructor or by the first call to GetLatestVersion() triggered by either one of the API calls from VersionedLayerClient.
This means that once the user can specify the version or boost::none in the constructor we will deprecate all the WithVersion() from the request classes and we should ignore them.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
```diff
@@ -73,10 +73,19 @@ class VersionedLayerClientImpl {
       PrefetchTilesRequest request);
 
  private:
+  using VersionResponse = client::ApiResponse<int64_t, client::ApiError>;
+  VersionResponse GetVersion(boost::optional<int64_t> requested_version,
```
andescu:
```suggestion
  VersionResponse GetVersion(boost::optional<std::string> billing_tag,
```

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
```diff
@@ -73,10 +73,19 @@ class VersionedLayerClientImpl {
       PrefetchTilesRequest request);
 
  private:
+  using VersionResponse = client::ApiResponse<int64_t, client::ApiError>;
```
andescu: Why don't you use directly `CatalogVersionResponse` from Types.h?

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
```diff
@@ -73,10 +73,19 @@ class VersionedLayerClientImpl {
       PrefetchTilesRequest request);
```
andescu: Add new constructor taking in a `boost::optional<int32_t> version`.
See #634.

PR #652, opened 2020-02-11
Deprecate WithVersion methods in request classes.

mykhailo-kuchma:
VersionedLayerClient needs to serve consistent data for all requests
which are triggered on a particular instance. If user do not know the
catalog version up front and always use boost::none for version, the
risk of returning different catalog versions data is high. Therefore
we lock the catalog version in the constructor of the VersionLayerClient
instance and all requests are therefore served with that particular catalog
version. Hence we do not need the WithVersion() methods in the requests
structures anymore and they will be ignore during requests and the constructor
passed catalog version will be taken instead.

Relates-To: OLPEDGE-1514

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
@@ -22,6 +22,7 @@
 #include <sstream>
 #include <string>
 
+#include <olp/core/porting/deprecated.h>
 #include <olp/dataservice/read/DataServiceReadApi.h>
 #include <olp/dataservice/read/FetchOptions.h>
 #include <boost/optional.hpp>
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
```diff
@@ -49,16 +51,19 @@ class DATASERVICE_READ_API DataRequest final {
    *
    * @return A reference to the updated `DataRequest` instance.
    */
+  OLP_SDK_DEPRECATED("Deprecated, to be removed in 05.2020")
```
andescu:
```suggestion
  OLP_SDK_DEPRECATED("Deprecated, to be removed in 06.2020")
```
Same for below once.

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
```diff
@@ -41,6 +42,7 @@ namespace read {
 class DATASERVICE_READ_API DataRequest final {
  public:
   /**
+   * @deprecated
```
andescu: Please move down after return and write why it is deprecated and what users should use instead, which means basically lock version in constructor.

PR #658, opened 2020-02-14
Create AuthenticationSettings

kzvieriev:
This is a part of authentication refactoring. Added
AuthenticationSettings struct is same as Settings but without
credentials. AuthenticationClient can replace setters to NetworkProxy,
Network and TaskScheduler by the new struct.

Resolves: OLPEDGE-1261

### DIFF
--- a/olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
+++ b/olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include <boost/optional.hpp>
+#include <olp/core/http/NetworkProxySettings.h>
+#include <olp/authentication/AuthenticationApi.h>
+
+namespace olp {
+namespace http {
+class Network;
+}
+
+namespace thread {
+class TaskScheduler;
+}
+
+namespace authentication {
+/// The default token endpoint url.
+static constexpr auto kHereAccountProductionTokenEndpointUrl =
+    "https://account.api.here.com/oauth2/token";
+
+/**
+ * @brief Configures the `TokenEndpoint` instance.
+ *
+ * Contains settings that customize the `TokenEndpoint` class.
+ */
+struct AUTHENTICATION_API AuthenticationSettings {
+  /**
+   * @brief The configuration settings for the network layer.
+   *
+   * To remove any existing proxy settings, set to boost::none.
+   */
+  boost::optional<http::NetworkProxySettings> network_proxy_settings;
+
+  /**
+   * @brief The network instance that is used to internally operate with the OLP
+   * services.
+   */
+  std::shared_ptr<http::Network> network_request_handler = nullptr;
+
+  /**
+   * @brief The `TaskScheduler` class that is used to manage
+   * the callbacks enqueue.
+   *
+   * If nullptr, all request calls are performed synchronously.
+   */
+  std::shared_ptr<thread::TaskScheduler> task_scheduler;
+
+  /**
+   * @brief The server URL of the token endpoint.
+   *
+   * @note Only standard OAuth2 Token URLs (those ending in `oauth2/token`) are
+   * supported.
+   */
+  std::string token_endpoint_url{kHereAccountProductionTokenEndpointUrl};
+};
+
+}  // namespace authentication
+}  // namespace olp


olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (C) 2019 HERE Europe B.V.
```
r0busta:
```suggestion
 * Copyright (C) 2020 HERE Europe B.V.
```

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,77 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#include <boost/optional.hpp>
```
r0busta: Not a system header; belongs closer to line 26

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,77 @@
...
+}
+
+namespace thread {
+class TaskScheduler;
+}
+
+
+namespace authentication {
+static const std::string kHereAccountProductionTokenUrl =
+    "https://account.api.here.com/oauth2/token";
```
r0busta: Correct me if I'm wrong, but defining static const strings in headers might cause issues with linkers, unspecified global initialization errors, as well as, sometimes it causes even double-free segfaults. I suggest moving initialization in a cpp file.

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,77 @@
...
+class Network;
+}
+
+namespace thread {
+class TaskScheduler;
+}
+
+
+namespace authentication {
+static const std::string kHereAccountProductionTokenUrl =
```
r0busta: Documentation for the constant is missing

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,77 @@
...
+struct AUTHENTICATION_API AuthenticationSettings {
+  /**
+   * @brief The network instance that is used to internally operate with the OLP
+   * services.
+   */
+  std::shared_ptr<http::Network> network_request_handler;
+
+  /**
+   * @brief (Optional) The `TaskScheduler` class that is used to manage
+   * the callbacks enqueue.
```
r0busta: I suggest being consistent with the rest of settings classes and instead of saying the scheduler is optional, I'd use the same wording: "If `nullptr`, all request calls are performed synchronously."

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,77 @@
...
+   */
+  std::shared_ptr<thread::TaskScheduler> task_scheduler;
+
+  /**
+   * @brief (Optional) The configuration settings for the network layer.
```
r0busta:
```suggestion
   * @brief The configuration settings for the network layer.
```

It's redundant with the property type. I suggest being consistent with the rest of settings classes and add a following line instead of `(Optional)`: "To remove any existing proxy settings, set to `boost::none`."

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,77 @@
...
+   */
+  boost::optional<http::NetworkProxySettings> network_proxy_settings;
+
+  /**
+   * @brief (Optional) The server URL of the token endpoint.
```
r0busta:
```suggestion
   * @brief The server URL of the token endpoint.
```
Would it make sense to use the `boost::optional<std::string>` type for the property instead?

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
kzvieriev: Agree.
Settings.h using the same url `https://account.api.here.com/oauth2/token`. Would be good to move both to one place. Do we have such place?
Or make token_endpoint_url `boost::optional` as you mention below. AuthenticationClient will use `https://account.api.here.com/oauth2/token` instead of `boost::none`.

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include <boost/optional.hpp>
+#include <olp/core/http/NetworkProxySettings.h>
+#include "AuthenticationApi.h"
```
andescu: All public APIs should include with global scope:

```cpp
#include <olp/authentication/AuthenticationApi.h>
```

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
andescu: Use `static constexpr auto kHereAccountProductionTokenUrl = "...."`
This is a compile time define.
If this also does not work then let it empty and write comment that if empty we will use "https://account.api.here.com/oauth2/token" although this is not my favorite solution and is pretty ambiguous.

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,79 @@
...
+ * @brief Configures the `TokenEndpoint` instance.
+ *
+ * Contains settings that customize the `TokenEndpoint` class.
+ */
+struct AUTHENTICATION_API AuthenticationSettings {
+  /**
+   * @brief The network instance that is used to internally operate with the OLP
+   * services.
+   */
+  std::shared_ptr<http::Network> network_request_handler;
```
andescu:
```suggestion
  std::shared_ptr<http::Network> network_request_handler = nullptr;
```
or 

```cpp
std::shared_ptr<http::Network> network_request_handler{nullptr};
```

If you want to keep similarity.

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,79 @@
...
+   * If nullptr, all request calls are performed synchronously.
+   */
+  std::shared_ptr<thread::TaskScheduler> task_scheduler;
+
+  /**
+   * @brief The configuration settings for the network layer.
+   *
+   * To remove any existing proxy settings, set to boost::none.
+   */
+  boost::optional<http::NetworkProxySettings> network_proxy_settings;
```
andescu: Please move this before `network_request_handler`.

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
kzvieriev: Not sure about this one.
`token_endpoint_url` has default value. What is expected for `boost::none` then?

r0busta: If it has a default value, it's not an optional property anymore (as it always has a value), is it?

andescu: I don't think we should create the impression that this value is optional. Actually, without the user providing this value, we cannot do authentication. I would say a non-optional value with a default assigned URL is a good strong message :)

kzvieriev: Yeah, sorry, I misunderstood the original comment.

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -0,0 +1,80 @@
...
+}
+
+namespace thread {
+class TaskScheduler;
+}
+
+namespace authentication {
+/// The default token endpoint url.
+static constexpr auto kHereAccountProductionTokenEndpointUrl =
+    "https://account.api.here.com/oauth2/token";
```
kzvieriev:
``` suggestion
static constexpr auto kHereAccountProductionTokenEndpointUrl =
    "https://account.api.here.com";
```
In current implementation, AuthenticationClient has url input which is Settings::token_endpoint_url but without "/oauth2/token" ending. Maybe this default value should be without the ending too, so it could be used instead of the url?

andescu: Yeah, this is a long story but currently, it is still necessary but we can adapt later on during the refactoring of the TokenEndpoint.

PR #662, opened 2020-02-14
Add a cpp client for MockServer

r0busta:
With the client, we'll be able to mock HTTP responses of various OLP services.

The MockServer CPP client helps to mock such responses directly from `gtest` code

Relates-To: OLPEDGE-1252

### DIFF
--- a/tests/utils/mock-server-client/Client.h
+++ b/tests/utils/mock-server-client/Client.h
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <olp/core/client/OlpClient.h>
+#include <olp/core/client/OlpClientFactory.h>
+#include <olp/core/client/OlpClientSettings.h>
+#include <olp/core/generated/parser/JsonParser.h>
+#include "Expectation.h"
+#include "Status.h"
+
+namespace mockserver {
+
+const auto kBaseUrl = "https://localhost:1080";
+const auto kExpectationPath = "/mockserver/expectation";
+const auto kStatusPath = "/mockserver/status";
+const auto kResetPath = "/mockserver/reset";
+const auto kTimeout = std::chrono::seconds{10};
+
+class Client {
+ public:
+  explicit Client(olp::client::OlpClientSettings settings) {
+    http_client_ = olp::client::OlpClientFactory::Create(settings);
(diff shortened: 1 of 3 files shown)


mykhailo-kuchma:
I assume a mock server should not be used for performance/memory tests.

andescu:
> I assume a mock server should not be used for performance/memory tests.

Why not? Why should we use different servers for different purposes? As far as I see we can configure the mock_server to have the default actions as currently defined in the olp_server.

tools/mock-server-client/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+# Copyright (C) 2020 HERE Europe B.V.
```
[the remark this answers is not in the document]
r0busta: Yes, I'd say we need it as an isolated library. In this or any other repository. 
So we can use the mock server in other projects.

r0busta:
> Let us do it step by step, please.
> First, let's add the mock server with necessary helpers (set expectation, latency, error code, etc.) to the repo by removing the unused mock_server in `tests/utils/mock_server`.
> Then we can gradually connect to functional and performance tests.

The `tests/utils/mock_server` is used. By perf tests. So we either have two tools and then adjust the perf tests, or we do the switch in one go, as in this pull request.

I chose the second approach because the MockServer and its client have a pretty slim code, and everything can be reviewed in one go.

But I'm okay with splitting.

r0busta:
Created a separate PR #664 that adds the mock-server runner

tools/mock-server-client/src/Client.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  boost::optional<Expectation::ResponseAction> action =
+      Expectation::ResponseAction{};
+  action->body = response_body;
+  expectation.action = action;
+
+  CreateExpectation(expectation);
+};
+
+std::vector<int32_t> Client::Ports() const {
```
andescu: Use OlpClient instead?

tools/mock-server-client/CMakeLists.txt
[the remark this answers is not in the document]
andescu: I dont think we need an extra library for this. Aim for headers only and then we can include in any upstream project by path.

tests/utils/mock-server-client/Client.h
```diff
@@ -0,0 +1,120 @@
...
+        std::multimap<std::string, std::string>(), nullptr, "",
+        olp::client::CancellationContext{});
+
+    if (response.status != olp::http::HttpStatusCode::OK) {
+      return {};
+    }
+
+    auto response_body = std::string{};
+    response.GetResponse(response_body);
+    const auto status = olp::parser::parse<Status>(response_body);
```
andescu: The parser also accepts std::stringstream.

tests/utils/mock-server-client/Client.h
```diff
@@ -0,0 +1,120 @@
...
+        std::multimap<std::string, std::string>(), nullptr, "",
+        olp::client::CancellationContext{});
+
+    return;
+  }
+
+ private:
+  void CreateExpectation(const Expectation& expectation) {
+    const auto data = serialize(expectation);
+    std::cout << data << std::endl;
```
andescu: Needed?

tests/utils/mock-server-client/Client.h
```diff
@@ -0,0 +1,120 @@
...
+};
+
+static std::unique_ptr<mockserver::Client> CreateDefaultMockServerClient(){
+  auto settings = olp::client::OlpClientSettings{};
+  settings.network_request_handler = olp::client::OlpClientSettingsFactory::
+      CreateDefaultNetworkRequestHandler();
+  settings.task_scheduler =
+      olp::client::OlpClientSettingsFactory::CreateDefaultTaskScheduler();
+  return std::make_unique<mockserver::Client>(settings);
+}
```
andescu: I would expect we share the Network layer with the clients. Maybe take the OlpClientSettings as input?

PR #671, opened 2020-02-17
Add support for X-Correlation-Id

LiubovDidkivska:
Get x-corelation-id from http responce  headers.
Modify mock network to return headers.

Relates-To: OLPEDGE-1355

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
+++ b/olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
@@ -132,12 +132,12 @@ CORE_API std::string ErrorCodeToString(ErrorCode code);
 /**
  * @brief Type alias for HTTP header.
  */
-using HeaderPair = std::pair<std::string, std::string>;
+using Header = std::pair<std::string, std::string>;
 
 /**
  * @brief Type alias for vector of HTTP headers.
  */
-using HeadersType = std::vector<HeaderPair>;
+using Headers = std::vector<Header>;
 
 }  // namespace http
 }  // namespace olp
--- a/olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
+++ b/olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
@@ -130,9 +130,14 @@ class SendOutcome final {
 CORE_API std::string ErrorCodeToString(ErrorCode code);
 
 /**
- * @brief Type alias for HTTP headers.
+ * @brief Type alias for HTTP header.
  */
-using HeadersType = std::vector<std::pair<std::string, std::string>>;
+using HeaderPair = std::pair<std::string, std::string>;
+
+/**
+ * @brief Type alias for vector of HTTP headers.
+ */
+using HeadersType = std::vector<HeaderPair>;
 
 }  // namespace http
 }  // namespace olp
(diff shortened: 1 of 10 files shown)


olp-cpp-sdk-dataservice-read/src/generated/api/StreamApi.cpp
```diff
@@ -87,9 +87,10 @@ StreamApi::SubscribeApiResponse StreamApi::Subscribe(
 
   OLP_SDK_LOG_DEBUG_F(kLogTag, "subscribe, uri=%s, status=%d",
                       metadata_uri.c_str(), http_response.status);
-
-  // TODO: Set x_correlation_id to the value received in http_response.header
-  // when http_response.header will be implemented.
+  if (!HandleCorrelationId(http_response.headers, x_correlation_id)) {
+    OLP_SDK_LOG_INFO(kLogTag,
```
andescu: I don't think we need any logs for this. Covered by integration tests and this should be enough.

olp-cpp-sdk-dataservice-read/src/generated/api/StreamApi.cpp
```diff
@@ -217,12 +220,30 @@ Response<int> StreamApi::HandleOffsets(
...
+                     "X-Correlation-Id was not found in responce headers");
+  }
 
   return http_response.status;
 }
 
+bool StreamApi::HandleCorrelationId(const http::HeadersType& headers,
+                                    std::string& x_correlation_id) {
+  auto it = std::find_if(std::begin(headers), std::end(headers),
+                         [](const std::pair<std::string, std::string>& header) {
```
andescu:
```suggestion
                         [](const headers::value_type& header) {
```
Maybe also a alias for the value_type should be added to NetworkTypes.h

olp-cpp-sdk-dataservice-read/src/generated/api/StreamApi.cpp
```diff
@@ -217,12 +220,30 @@ Response<int> StreamApi::HandleOffsets(
...
+                           if (header.first.compare("X-Correlation-Id") == 0) {
+                             return true;
+                           }
+                           return false;
```
andescu:
```cpp
return (header.first.compare("X-Correlation-Id") == 0);
```

olp-cpp-sdk-dataservice-read/src/generated/api/StreamApi.h
```diff
@@ -256,6 +256,9 @@ class StreamApi {
       const boost::optional<std::string>& mode,
       const client::CancellationContext& context, const std::string& endpoint,
       std::string& x_correlation_id);
+
+  static bool HandleCorrelationId(const http::HeadersType& headers,
```
andescu: Private static methods feel more comfortable of they are anonymous functions :) Keeps the headers clean.

olp-cpp-sdk-core/include/olp/core/http/NetworkTypes.h
```diff
@@ -130,9 +130,14 @@ class SendOutcome final {
 CORE_API std::string ErrorCodeToString(ErrorCode code);
 
 /**
- * @brief Type alias for HTTP headers.
+ * @brief Type alias for HTTP header.
  */
-using HeadersType = std::vector<std::pair<std::string, std::string>>;
+using HeaderPair = std::pair<std::string, std::string>;
```
andescu:
```suggestion
using HeaderType = std::pair<std::string, std::string>;
```

LiubovDidkivska: Renamed to Header and Headers. It is harder for humans distinguish differences in one letter inside word, than at the ends.

PR #673, opened 2020-02-19
Fix the warning message

mykhailo-kuchma:
Fix the condition when to print the warning about memory cache cant fit a value.

Resolves: OLPEDGE-1616

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCache.cpp
@@ -132,12 +132,14 @@ bool DefaultCache::Put(const std::string& key, const boost::any& value,
     return false;
   }
 
-  auto encodedItem = encoder();
+  auto encoded_item = encoder();
   if (memory_cache_) {
-    if (!memory_cache_->Put(key, value, expiry, encodedItem.size())) {
+    const auto size = encoded_item.size();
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -134,7 +134,8 @@ bool DefaultCache::Put(const std::string& key, const boost::any& value,
 
   auto encodedItem = encoder();
   if (memory_cache_) {
-    if (!memory_cache_->Put(key, value, expiry, encodedItem.size())) {
+    memory_cache_->Put(key, value, expiry, encodedItem.size());
+    if (encodedItem.size() > settings_.max_memory_cache_size) {
```
andescu: Maybe it is a good time to also adapt to the coding style.

olp-cpp-sdk-core/src/cache/DefaultCache.cpp
```diff
@@ -134,7 +134,8 @@ bool DefaultCache::Put(const std::string& key, const boost::any& value,
 
   auto encodedItem = encoder();
   if (memory_cache_) {
-    if (!memory_cache_->Put(key, value, expiry, encodedItem.size())) {
+    memory_cache_->Put(key, value, expiry, encodedItem.size());
```
andescu: Maybe check return together with the size?

PR #674, opened 2020-02-20
Fix crash in OlpClient

meravingen123:
Fix of thread concurrency problem in OlpClient
which in some situations resulted in crash

Resolves: OLPEDGE-1594

### DIFF
--- a/olp-cpp-sdk-core/src/client/OlpClient.cpp
+++ b/olp-cpp-sdk-core/src/client/OlpClient.cpp
@@ -130,61 +130,56 @@ HttpResponse SendRequest(const http::NetworkRequest& request,
                          const olp::client::OlpClientSettings& settings,
                          const olp::client::RetrySettings& retry_settings,
                          client::CancellationContext context) {
-  http::NetworkResponse network_response = kCancelledErrorResponse;
-  auto interest_flag = std::make_shared<std::atomic_bool>(true);
-  Condition condition{};
+  struct ResponseData {
+    olp::client::Condition condition_;
+    olp::http::NetworkResponse response_{kCancelledErrorResponse};
+    olp::http::Headers headers_;
+  };
+
+  auto response_data = std::make_shared<ResponseData>();
   auto response_body = std::make_shared<std::stringstream>();
   http::SendOutcome outcome{http::ErrorCode::CANCELLED_ERROR};
-  http::Headers headers;
 
   context.ExecuteOrCancelled(
       [&]() {
         outcome = settings.network_request_handler->Send(
             request, response_body,
-            [&, interest_flag](http::NetworkResponse response) {
-              if (interest_flag->exchange(false)) {
-                network_response = std::move(response);
-                condition.Notify();
-              }
+            [response_data](http::NetworkResponse response) {
+              response_data->response_ = std::move(response);
+              response_data->condition_.Notify();
             },
-            [&headers, interest_flag](std::string key, std::string value) {
-              if (interest_flag->load()) {
-                headers.emplace_back(std::move(key), std::move(value));
-              }
+            [response_data](std::string key, std::string value) {
+              response_data->headers_.emplace_back(std::move(key),
+                                                   std::move(value));
             });
 
-        return CancellationToken([&, interest_flag]() {
-          if (interest_flag->exchange(false)) {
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -172,13 +199,16 @@ HttpResponse SendRequest(const http::NetworkRequest& request,
   if (!condition.Wait(std::chrono::seconds(retry_settings.timeout))) {
     OLP_SDK_LOG_INFO_F(kLogTag, "Timeout");
     context.CancelOperation();
+    while (interest_flag->load() == State::Processing) {
```
andescu: What if the state is Waiting or Done? This will be an endless loop.
Also, it seems to me that the context.CancelOperation() called one line before the check is setting the flag to State::Done actually, right? Except of course it is in the state Processing.

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -142,19 +148,40 @@ HttpResponse SendRequest(const http::NetworkRequest& request,
...
               }
+
+              network_response = std::move(response);
+              interest_flag->store(State::Done);
+              condition.Notify();
             },
             [&headers, interest_flag](std::string key, std::string value) {
-              if (interest_flag->load()) {
+              State state = State::Waiting;
+              if (interest_flag->compare_exchange_weak(state,
```
andescu: This is not correct. This header callback is called multiple times, for each header once. The second time it is called the value of the interest_flag is already State::Processing so the compare_exchange_weak is going to return false so you will never reach to add any other header to the headers list.

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -126,12 +126,41 @@ http::NetworkRequest::HttpVerb GetHttpVerb(const std::string& verb) {
   return http_verb;
 }
 
+enum class State {
+  Waiting,
+  Processing,
+  Cancelling,
```
r0busta:
```suggestion
  Canceling,
```

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -172,13 +204,22 @@ HttpResponse SendRequest(const http::NetworkRequest& request,
   if (!condition.Wait(std::chrono::seconds(retry_settings.timeout))) {
     OLP_SDK_LOG_INFO_F(kLogTag, "Timeout");
     context.CancelOperation();
+
+    // We still can be in Processing or Cancelling state that was called
```
r0busta:
```suggestion
    // We still can be in Processing or Canceling state that was called
```

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -172,13 +204,22 @@ HttpResponse SendRequest(const http::NetworkRequest& request,
...
+    // from other thread so we must wait until that operation complete
+    // as they rely on stack variables allocated in this function.
+    auto state = shared_state->load();
+    while (state == State::Processing || state == State::Cancelling) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
mykhailo-kuchma: Do we need to sleep? might be too long

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -141,24 +170,27 @@ HttpResponse SendRequest(const http::NetworkRequest& request,
...
+              if (CheckState(shared_state)) {
                 headers.emplace_back(std::move(key), std::move(value));
               }
             });
 
-        return CancellationToken([&, interest_flag]() {
-          if (interest_flag->exchange(false)) {
+        return CancellationToken([&, shared_state]() {
+          State state = State::Waiting;
+          if (shared_state->compare_exchange_weak(state, State::Cancelling)) {
```
andescu: Be aware that this task can be cancelled anytime as it is attached to the CancellationContext given to the user. So it means we can also land here when state is `State::Processing` or `State::Done`.

meravingen123: And we cannot cancel it in processing state or done. This is the same behavior as it was before.

PR #678, opened 2020-02-20
Add implementation for get data using tile key.

LiubovDidkivska:
Get QuadTreeIndex for requested tile with debth equal 4, convert to partitions.
Put partitions to cache. Get blob data using data handle for requested tile key. Return data to user.

Relates-To: OLPEDGE-1529

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
@@ -34,9 +34,7 @@
 #include "generated/api/VolatileBlobApi.h"
 #include "olp/dataservice/read/CatalogRequest.h"
 #include "olp/dataservice/read/CatalogVersionRequest.h"
-#include "olp/dataservice/read/DataRequest.h"
 #include "olp/dataservice/read/PartitionsRequest.h"
-#include "olp/dataservice/read/TileRequest.h"
 
 namespace olp {
 namespace dataservice {
@@ -50,7 +48,7 @@ constexpr auto kBlobService = "blob";
 constexpr auto kVolatileBlobService = "volatile-blob";
 }  // namespace
 
-DataResponse DataRepository::GetVersionedDataTileQuadTree(
+DataResponse DataRepository::GetVersionedTile(
     const client::HRN& catalog, const std::string& layer_id,
     TileRequest request, int64_t version, client::CancellationContext context,
     client::OlpClientSettings settings) {
@@ -78,21 +76,6 @@ DataResponse DataRepository::GetVersionedDataTileQuadTree(
     return ApiError(ErrorCode::NotFound,
                     "Cache only resource not found in cache (data).");
   } else {
-    auto parent_tile = request;
-    parent_tile.WithTileKey(request.GetTileKey().Parent());
-    auto cached_partitions = PartitionsRepository::GetTileFromCache(
-        catalog, layer_id, parent_tile, version, settings);
-    if (cached_partitions.GetPartitions().size() != 0) {
-      DataRequest data_request =
-          DataRequest()
-              .WithPartitionId(request.GetTileKey().ToHereTile())
-              .WithVersion(version)
-              .WithFetchOption(request.GetFetchOption());
-      // get the requested tile data, if parent tile was found in cache
-      return repository::DataRepository::GetVersionedData(
-          catalog, layer_id, data_request, context, settings);
-    }
-
     auto response = PartitionsRepository::QueryPartitionsAndGetDataHandle(
         catalog, layer_id, request, version, context, settings,
         requested_tile_data_handle);
@@ -103,8 +86,7 @@ DataResponse DataRepository::GetVersionedDataTileQuadTree(
 
   if (requested_tile_data_handle.empty()) {
     OLP_SDK_LOG_ERROR(
-        kLogTag,
-        "GetVersionedDataTileQuadTree: requested tile handle was not found");
+        kLogTag, "GetVersionedDataTile: requested tile handle was not found");
     return client::ApiError(client::ErrorCode::NotFound,
                             "Requested tile handle was not found.");
   }
--- a/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
@@ -31,7 +31,6 @@
 #include "PartitionsCacheRepository.h"
 #include "PartitionsRepository.h"
 #include "generated/api/BlobApi.h"
-#include "generated/api/QueryApi.h"
 #include "generated/api/VolatileBlobApi.h"
 #include "olp/dataservice/read/CatalogRequest.h"
 #include "olp/dataservice/read/CatalogVersionRequest.h"
@@ -49,7 +48,6 @@ namespace {
 constexpr auto kLogTag = "DataRepository";
 constexpr auto kBlobService = "blob";
 constexpr auto kVolatileBlobService = "volatile-blob";
-constexpr std::uint32_t kMaxQuadTreeIndexDepth = 4u;
 }  // namespace
 
 model::Partitions DataRepository::GetPartitionsFromCache(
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.h
```diff
@@ -29,10 +29,16 @@ namespace olp {
 namespace dataservice {
 namespace read {
 class DataRequest;
+class TileRequest;
 namespace repository {
 
 class DataRepository final {
  public:
+  static DataResponse GetVersionedDataTileQuadTree(
```
andescu:
```suggestion
  static DataResponse GetVersionedTile(
```
How we do it internally is our business. Doesn't need to be extra marked in API.

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.h
```diff
@@ -29,10 +29,16 @@ namespace olp {
 namespace dataservice {
 namespace read {
 class DataRequest;
+class TileRequest;
```
andescu: This doesn't work. Actually you can remove both fwd defines.
We already include Types.h and fwd defines only work for references or ptrs.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -40,6 +41,7 @@ using namespace olp::client;
 
 namespace {
 constexpr auto kLogTag = "PartitionsRepository";
+constexpr std::uint32_t kMaxQuadTreeIndexDepth = 4u;
```
andescu:
```suggestion
constexpr uint32_t kMaxQuadTreeIndexDepth = 4u;
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
```diff
@@ -56,6 +58,21 @@ class PartitionsRepository final {
...
+      TileRequest request, int64_t version, client::CancellationContext context,
+      client::OlpClientSettings settings,
+      std::string& requested_tile_data_handle);
+
+  static model::Partitions GetTileFromCache(const client::HRN& catalog,
+                                            const std::string& layer_id,
+                                            TileRequest request,
+                                            int64_t version,
+                                            client::OlpClientSettings settings);
+
```
andescu: Not sure I get it.
I assumed we have something like `GetVersionedTile(...)` which looks in the cache if the tiles partition is there, if not goes to the `QueryApi::QuadTreeIndex(...)` with depth 4 which downloads quad-tree and caches each subquad as partition metadata same as `PrefetchTilesRepository::GetSubQuads()`. At the end it returns the requested partition to the user.
Additional improvements can be done later on, like:
- If requested tile key not found i cache check if parent is found and if yes then assume there was a quad tree index query before but it maybe got LRU evicted in the meantime so we maybe only downlaod that specific tile in the end.
-Return nearest tile if the tile is not in the response (optional, old hype behaviour)

LiubovDidkivska: 1.Updated, please check. Added check if parent tile in cache. If true, get requested tile. Added test GetVersionedDataTileQuadTreeParentTileInCache.
2. Return nearest tile if the tile is not in the response. Is this applicable only for CacheOnly option?

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.h
LiubovDidkivska: There is no DataRequest in Types.h. Forward declarations changed to includes.

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.h
andescu: My point is we don't need to add implementation details to the method name. Just call it `GetVersionedTile`. How we get the tile or the metadata internally is not important for the API name.

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
```diff
@@ -49,6 +50,74 @@ constexpr auto kBlobService = "blob";
...
+          catalog, layer_id, data_request, context, settings);
+    }
+
+    auto response = PartitionsRepository::QueryPartitionsAndGetDataHandle(
+        catalog, layer_id, request, version, context, settings,
+        requested_tile_data_handle);
+    if (!response.IsSuccessful()) {
+      return response.GetError();
+    }
+  }
```
andescu: All this should not be here, it should be implementation details of the `PartitionsRepository::GetVersionedTile()`.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
andescu: Those were only theoretical scenarios for the future. At least the 2. point I would omit for now as I don't see any requirement.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
```diff
@@ -56,6 +58,21 @@ class PartitionsRepository final {
...
+      const client::HRN& catalog, const std::string& layer_id,
+      TileRequest request, int64_t version, client::CancellationContext context,
+      client::OlpClientSettings settings,
+      std::string& requested_tile_data_handle);
+
+  static model::Partitions GetTileFromCache(const client::HRN& catalog,
+                                            const std::string& layer_id,
+                                            TileRequest request,
+                                            int64_t version,
+                                            client::OlpClientSettings settings);
```
andescu: Why not only `GetVersionedTile()`?
Why do you need three public methods if all three serve actually an internal implementation purpose and are not in any point related to what DataRepository needs?
DataRepository::GetVersionedTile() is only interested in the partition metadata for his request. How you get this partition metadata (by using quad tree index, by requesting single partition) is implementation details of PartitionsRepository and should not be handled by DataRepository.

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
mykhailo-kuchma: Please check the other methods in this class, partition repo works with the metadata only.

andescu: This is an implementation detail of PartitionsRepo which is easily fixed, considering that the query REST API wrapper is static and can be used everywhere.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
LiubovDidkivska: I can move this methods to DataRepository and make them private. It was spitted to do not have very big method, and theoretically this methods could be useful for getting metadata for partitions in future. Also it easier to test and cover all functionality separately. Should I move everything to one function?

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.h
LiubovDidkivska: Renamed

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
LiubovDidkivska: Removed checking for parent tile in cache.

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
LiubovDidkivska: Moved to PartitionRepository

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -233,6 +235,132 @@ PartitionsResponse PartitionsRepository::GetPartitionById(
...
+    TileRequest request, int64_t version, client::CancellationContext context,
+    client::OlpClientSettings settings) {
+  auto fetch_option = request.GetFetchOption();
+  auto tile = request.GetTileKey().ToHereTile();
+  auto query_api = ApiClientLookup::LookupApi(
+      catalog, context, "query", "v1", request.GetFetchOption(), settings);
+
+  if (!query_api.IsSuccessful()) {
+    OLP_SDK_LOG_ERROR(kLogTag,
+                      "QueryPartitionsAndGetDataHandle: LookupApi failed.");
```
andescu:
```suggestion
                      "QueryPartitionForVersionedTile: LookupApi failed.");
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -233,6 +235,132 @@ PartitionsResponse PartitionsRepository::GetPartitionById(
...
+                        "QuadTreeIndex failed (%s, %" PRId64 ", %" PRId32 ")",
+                        tile.c_str(), version, kMaxQuadTreeIndexDepth);
+    return quad_tree.GetError();
+  }
+  model::Partitions result;
+  model::Partitions partitions_for_caching;
+  const auto& subquads = quad_tree.GetResult().GetSubQuads();
+  partitions_for_caching.GetMutablePartitions().reserve(subquads.size());
+
+  OLP_SDK_LOG_TRACE_F(kLogTag, "Requested tile subquads size %lu.",
```
andescu:
```suggestion
  OLP_SDK_LOG_DEBUG_F(kLogTag, "Requested tile subquads size %lu.",
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -233,6 +235,132 @@ PartitionsResponse PartitionsRepository::GetPartitionById(
...
+  if (!quad_tree.IsSuccessful()) {
+    OLP_SDK_LOG_ERROR_F(kLogTag,
+                        "QuadTreeIndex failed (%s, %" PRId64 ", %" PRId32 ")",
+                        tile.c_str(), version, kMaxQuadTreeIndexDepth);
+    return quad_tree.GetError();
+  }
+  model::Partitions result;
+  model::Partitions partitions_for_caching;
+  const auto& subquads = quad_tree.GetResult().GetSubQuads();
+  partitions_for_caching.GetMutablePartitions().reserve(subquads.size());
```
andescu: Please extract to a local variable partitions_for_caching.GetMutablePartitions()

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -233,6 +235,132 @@ PartitionsResponse PartitionsRepository::GetPartitionById(
...
+
+  OLP_SDK_LOG_TRACE_F(kLogTag, "Requested tile subquads size %lu.",
+                      subquads.size());
+
+  for (const auto& subquad : subquads) {
+    auto subtile =
+        request.GetTileKey().AddedSubHereTile(subquad->GetSubQuadKey());
+    // add partitions for caching
+    partitions_for_caching.GetMutablePartitions().emplace_back(
+        PartitionsRepository::PartitionFromSubQuad(*subquad,
```
andescu:
```suggestion
       PartitionFromSubQuad(*subquad, subtile.ToHereTile()));
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -233,6 +235,132 @@ PartitionsResponse PartitionsRepository::GetPartitionById(
...
+                                                   subtile.ToHereTile()));
+    // find data handle for requested tile
+    if (subtile.ToHereTile().compare(tile) == 0) {
```
andescu: You compare tiles as string, but you can compare them as TileKeys. Why?

`subtile.ToHereTile().compare(tile.ToHereTile())` is same as `subtile == tile`.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -233,6 +235,132 @@ PartitionsResponse PartitionsRepository::GetPartitionById(
...
+      boost::none, request.GetBillingTag(), context);
+
+  if (!quad_tree.IsSuccessful()) {
+    OLP_SDK_LOG_ERROR_F(kLogTag,
+                        "QuadTreeIndex failed (%s, %" PRId64 ", %" PRId32 ")",
+                        tile.c_str(), version, kMaxQuadTreeIndexDepth);
+    return quad_tree.GetError();
+  }
+  model::Partitions result;
+  model::Partitions partitions_for_caching;
```
andescu:
```suggestion
  model::Partitions partitions;
```
Would simplify a lot the formatting, I presume.

PR #685, opened 2020-02-24
Fix Cache keys.

LiubovDidkivska:
Append 'partition' and 'partitions' for key values.

Relates-To: OLPEDGE-1568

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
@@ -40,12 +40,12 @@ std::string CreateKey(const std::string& hrn, const std::string& layer_id,
                       const std::string& partitionId,
                       const boost::optional<int64_t>& version) {
   return hrn + "::" + layer_id + "::" + partitionId +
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
```diff
@@ -40,7 +40,7 @@ std::string CreateKey(const std::string& hrn, const std::string& layer_id,
                       const std::string& partitionId,
                       const boost::optional<int64_t>& version) {
   return hrn + "::" + layer_id + "::" + partitionId +
-         "::" + (version ? std::to_string(*version) + "::partition" : "");
+         "::" + (version ? std::to_string(*version) + "::" : "") + "partition";
 }
 std::string CreateKey(const std::string& hrn, const std::string& layer_id,
```
andescu: Same here. The layer metadata should also include `partitions` at the end.

LiubovDidkivska: added partitions and fixed Clear function.

PR #689, opened 2020-02-26
Add support for a list of partitions to GetPartition API

mykhailo-kuchma:
When the user specifies the list of partitions the query API is used instead of
metadata API. This brings the user functionality to access partitions
metadata of individual partitions.
The implementation checks the cache if the number of requested partitions is
less then the number of partitions found in the cache, the check fails and SDK
performs the online request.

Resolves: OLPEDGE-1017

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
@@ -121,7 +121,6 @@ PartitionsResponse PartitionsRepository::GetPartitions(
     client::CancellationContext cancellation_context, PartitionsRequest request,
     const client::OlpClientSettings& settings, boost::optional<time_t> expiry) {
   auto fetch_option = request.GetFetchOption();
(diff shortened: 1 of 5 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -139,32 +138,54 @@ PartitionsResponse PartitionsRepository::GetPartitions(
...
+    response = QueryApi::GetPartitionsbyId(
+        query_api.GetResult(), layer, partition_ids, request.GetVersion(),
+        boost::none, request.GetBillingTag(), cancellation_context);
   }
 
-  auto metadata_response = MetadataApi::GetPartitions(
-      query_api.GetResult(), layer, request.GetVersion(), boost::none,
-      boost::none, request.GetBillingTag(), cancellation_context);
+  // Save all partitions only when downloaded via metadata API
+  const bool save_all_layer = partition_ids.empty();
```
andescu:
```suggestion
  const bool is_layer_metadata = partition_ids.empty();
```

PR #691, opened 2020-02-26
Fix iifinite loop in PrefetchTiles. Add tests with wrong levels for PrefetchTiles.

LiubovDidkivska:
Add tests for PrefetchTiles with different min/max levels. 
Fix infinite recursion by adding checks for valid TileKey value. 
Reduce min/max level to current level, current level+depth(depth is 4).

Relates-To: OLPEDGE-1624

### DIFF
--- a/tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVersionedLayerClientTest.cpp
+++ b/tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVersionedLayerClientTest.cpp
@@ -280,8 +280,8 @@ TEST_F(DataserviceReadVersionedLayerClientTest, PrefetchWrongLevels) {
     std::vector<olp::geo::TileKey> tile_keys = {
         olp::geo::TileKey::FromHereTile(kTileId)};
 
-    SCOPED_TRACE("min/max levels are 0");
     {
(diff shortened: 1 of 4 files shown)


tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVersionedLayerClientTest.cpp
```diff
@@ -263,6 +263,98 @@ TEST_F(DataserviceReadVersionedLayerClientTest, Prefetch) {
...
+
+  auto client = std::make_unique<olp::dataservice::read::VersionedLayerClient>(
+      catalog, kLayerId, *settings_);
+
+  {
+    SCOPED_TRACE("Prefetch tiles online and store them in memory cache");
+    std::vector<olp::geo::TileKey> tile_keys = {
+        olp::geo::TileKey::FromHereTile(kTileId)};
+
+    SCOPED_TRACE("min/max levels are 0");
```
andescu: The SCOPED_TRACE needs to be under the scope else you will get it when some of the tests below fail also.

```cpp
{
    SCOPED_TRACE("123");
}
{
    SCOPED_TRACE("456");
}
```

PR #693, opened 2020-02-28
Update VersionedLayerClient example and doxygen.

kzvieriev:
VersionedLayerClient API is changed, version is specified in constructor
instead of DataRequest now. So the goal of this change is to adapt
example and doxygen to the new API. Added new argument for catalog
version.

Resolves: OLPEDGE-1566

### DIFF
--- a/examples/main.cpp
+++ b/examples/main.cpp
@@ -22,9 +22,10 @@
 #include "ReadExample.h"
 #include "WriteExample.h"
 
-#include <iostream>
 #include <stdio.h>
 #include <string.h>
+#include <iostream>
+#include <sstream>
 
 bool IsMatch(const std::string& name, const tools::Option& option) {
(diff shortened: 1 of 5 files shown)


examples/main.cpp
```diff
@@ -104,6 +108,17 @@ int ParseArguments(const int argc, char** argv, AccessKey& access_key,
...
+      ++it;
+      try {
+        catalog_version = std::stoll(*it);
+      } catch (const std::invalid_argument&) {
+        std::cout << "invalid catalog version value -- '" << *it
+                  << "', but int64 is expected." << std::endl;
+      } catch (const std::out_of_range&) {
+        std::cout << "invalid catalog version value -- '" << *it
+                  << "', its too big." << std::endl;
+      }
```
andescu: We don't/shouldn't use exceptions.

PR #694, opened 2020-02-28
Expose introspect app API

kzvieriev:
Added new IntrospectApp method for AuthenticationClient. Now it's
possible to obtain application info like name, description, realm etc.
using access token.
Functional and integration tests added also.

Resolves: OLPEDGE-1489

### DIFF
--- a/olp-cpp-sdk-authentication/include/olp/authentication/Types.h
+++ b/olp-cpp-sdk-authentication/include/olp/authentication/Types.h
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#include <memory>
+
+#include <olp/authentication/AuthenticationError.h>
+#include <olp/authentication/IntrospectAppResult.h>
+#include <olp/core/client/ApiError.h>
+#include <olp/core/client/ApiResponse.h>
+
+namespace olp {
+namespace authentication {
+
+/// The response template type.
+template <typename ResultType>
+using Response = client::ApiResponse<ResultType, AuthenticationError>;
+
+/// The callback template type.
+template <typename ResultType>
+using Callback = std::function<void(Response<ResultType>)>;
+
+/// Defines the callback signature when the user request is completed.
+using IntrospectAppResponse = Response<IntrospectAppResult>;
+
+/// Called when the user introspect app request is completed.
+using IntrospectAppCallback = Callback<IntrospectAppResult>;
+
+}  // namespace authentication
+}  // namespace olp
--- a/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
+++ b/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
@@ -22,6 +22,7 @@
 #include <chrono>
 #include <sstream>
 
+#include <olp/core/client/PendingRequests.h>
 #include <rapidjson/document.h>
 #include <rapidjson/istreamwrapper.h>
 #include <rapidjson/stringbuffer.h>
@@ -29,6 +30,7 @@
 #include <boost/uuid/uuid.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
+
 #include "Constants.h"
 #include "Crypto.h"
 #include "SignInResultImpl.h"
@@ -39,6 +41,7 @@
 #include "olp/core/client/ApiError.h"
 #include "olp/core/client/CancellationToken.h"
 #include "olp/core/client/ErrorCode.h"
+#include "olp/core/client/OlpClient.h"
 #include "olp/core/http/HttpStatusCode.h"
 #include "olp/core/http/Network.h"
 #include "olp/core/http/NetworkConstants.h"
@@ -55,6 +58,8 @@ using namespace olp::thread;
 using namespace olp::utils;
 
 namespace {
+using namespace olp::authentication;
+
 // Helper characters
 constexpr auto kParamAdd = "&";
 constexpr auto kParamComma = ",";
@@ -76,6 +81,7 @@ const std::string kSignoutEndpoint = "/logout";
 const std::string kTermsEndpoint = "/terms";
 const std::string kUserEndpoint = "/user";
 const std::string kTimestampEndpoint = "/timestamp";
+const std::string kIntrospectAppEndpoint = "/app/me";
 
 // JSON fields
 constexpr auto kCountryCode = "countryCode";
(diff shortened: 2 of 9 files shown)


olp-cpp-sdk-authentication/include/olp/authentication/Types.h
```diff
@@ -0,0 +1,46 @@
...
+/// The callback template type.
+template <typename ResultType>
+using Callback = std::function<void(Response<ResultType>)>;
+
+/// Defines the callback signature when the user request is completed.
+using IntrospectAppResponse = Response<IntrospectAppResult>;
+
+/// Called when the user introspect app request is completed.
+using IntrospectAppCallback = Callback<IntrospectAppResult>;
+}  // namespace authentication
```
andescu: Add an empty line before namespace closing.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -190,13 +227,97 @@ class AuthenticationClient::Impl final {
...
+                                          IntrospectAppCallback callback);
+
   void SetNetworkProxySettings(
       const http::NetworkProxySettings& proxy_settings);
 
   void SetNetwork(std::shared_ptr<http::Network> network);
 
   void SetTaskScheduler(std::shared_ptr<TaskScheduler> task_scheduler);
 
+  static olp::authentication::IntrospectAppResult parseIntrospectAppResult(
```
andescu:
```suggestion
  static olp::authentication::IntrospectAppResult GetIntrospectAppResult(
```

Also, why is it a static method and not an anonymous function?

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -190,13 +227,97 @@ class AuthenticationClient::Impl final {
...
+      const std::shared_ptr<rapidjson::Document>& doc) {
```
andescu: If you do not need to copy or work on the shared_ptr then pass instead reference to rapidjson::Document or to const rapidjson::Document.

```suggestion
      const rapidjson::Document& doc) {
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -190,13 +227,97 @@ class AuthenticationClient::Impl final {
...
+      result.SetName((*doc)[Constants::NAME].GetString());
+    }
+    if (doc->HasMember(Constants::DESCRIPTION)) {
+      result.SetDescription((*doc)[Constants::DESCRIPTION].GetString());
+    }
+    if (doc->HasMember(Constants::REDIRECT_URIS)) {
+      auto uris = (*doc)[Constants::REDIRECT_URIS].GetArray();
+      std::vector<std::string> value_array;
+      value_array.reserve(uris.Size());
+      for (auto& value : uris) value_array.push_back(value.GetString());
```
andescu: Always add `{}`
Same below.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -190,13 +227,97 @@ class AuthenticationClient::Impl final {
...
+      value_array.reserve(scopes.Size());
+      for (auto& value : scopes) value_array.push_back(value.GetString());
+      result.SetAllowedScopes(std::move(value_array));
+    }
+    if (doc->HasMember(Constants::TOKEN_ENDPOINT_AUTH_METHOD)) {
+      result.SetTokenEndpointAuthMethod(
+          (*doc)[Constants::TOKEN_ENDPOINT_AUTH_METHOD].GetString());
+    } else {
+      result.SetTokenEndpointAuthMethod(
+          Constants::DEFAULT_TOKEN_ENDPOINT_AUTH_METHOD);
```
andescu: Why do we need to set a default one?

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -769,6 +894,77 @@ client::CancellationToken AuthenticationClient::Impl::SignOut(
...
+          auto error_msg = network_response.GetError();
+
+          if (response_status != 200) {
+            // Network error response not available
+            AuthenticationError result(response_status, error_msg);
+            callback(std::move(result));
+            return;
+          }
+
+          auto document = std::make_shared<rapidjson::Document>();
```
andescu: I wonder, do we still need to use shared_ptr for this? It is not passed anymore to the result class.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -769,6 +894,77 @@ client::CancellationToken AuthenticationClient::Impl::SignOut(
...
+        });
+
+    if (!send_outcome.IsSuccessful()) {
+      std::string error_message =
+          ErrorCodeToString(send_outcome.GetErrorCode());
+      client::ApiError result({static_cast<int>(send_outcome.GetErrorCode()),
+                               std::move(error_message)});
+      return std::move(result);
+    }
+    return IntrospectAppResult();
```
andescu: Why dont you return the real result and call the callback from the below wrap_callback? Why do you bind the callback to two lambdas?

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -769,6 +894,77 @@ client::CancellationToken AuthenticationClient::Impl::SignOut(
...
+  };
+
+  // wrap_callback needed to handle introspect_app_task errors occured before
+  // network request and to convert client::ApiError->AuthenticationError.
+  auto wrap_callback =
+      [callback](
+          client::ApiResponse<IntrospectAppResult, client::ApiError> response) {
```
andescu: I would make a local alias for this, as it is used in two places already:

```cpp
using ResponseType = client::ApiResponse<IntrospectAppResult, client::ApiError>;
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
kzvieriev: REST API says `The default value will be "client_secret_jwt" if the field is not included in the request body`, so an empty auth method field actually means `client_secret_jwt`.
Is this user responsibility to handle such case, so we can left it empty?

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
kzvieriev: Inside of wrap_callback we can't distinguish valid result from network->Send() response and empty result returned from introspect_app_task completion (`return IntrospectAppResult();`).

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
andescu: This `The default value will be "client_secret_jwt" if the field is not included in the request body` doesn't mean that you will not receive this value. Or?
So this all depends on the type of token passed by the user. I would not set any default values. Let the backend set what he thinks is necessary.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
andescu: But why are you returning an empty `IntrospectAppResult()` back? Is it because your wrap_callback has the assumption that only errors are sent to it. So if you also send the parsed result to wrap_callback, do you still need to return the default constructed `IntrospectAppResult`?

kzvieriev: network->Send() is async, parsed result is not ready when finishing introspect_app_task.
So I can only return some error or an empty result.
Well, I can use some special error and handle this case in wrap_callback.

andescu: Maybe it is then better to use OlpClient::CallApi(), as this is sync. If it is not possible then ignore my comment, although it looks very confusing.

kzvieriev: Done, should be better now.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -769,6 +901,74 @@ client::CancellationToken AuthenticationClient::Impl::SignOut(
...
+  auto introspect_app_task =
+      [=](client::CancellationContext context) -> ResponseType {
+    if (!settings_.network_request_handler) {
+      client::ApiError result({static_cast<int>(http::ErrorCode::IO_ERROR),
+                               "Cannot introspect app while offline"});
+      return std::move(result);
+    }
+
+    client::AuthenticationSettings auth_settings;
+    auth_settings.provider = [&access_token]() { return access_token; };
```
andescu: Do you need this? Probably not. If you leave this empty then the bearer token is not added to the headers.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -769,6 +901,74 @@ client::CancellationToken AuthenticationClient::Impl::SignOut(
...
+    settings.proxy_settings = settings_.network_proxy_settings;
+    settings.authentication_settings = auth_settings;
+
+    client::OlpClient client;
+    client.SetBaseUrl(settings_.token_endpoint_url);
+    client.SetSettings(settings);
+
+    std::multimap<std::string, std::string> query_params;
+    std::multimap<std::string, std::string> header_params;
+    std::multimap<std::string, std::string> forms_params;
```
andescu: These are input parameters, so if you don't need them then just pass default constructed values instead:

```suggestion
   auto http_result = client.CallApi(kIntrospectAppEndpoint, "GET",
                                      {}, {}, {}, nullptr, content_type, context);
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -769,6 +901,74 @@ client::CancellationToken AuthenticationClient::Impl::SignOut(
...
+                                     error.GetMessage()};
+      callback(std::move(auth_error));
+      return;
+    }
+
+    callback(response.MoveResult());
+  };
+
+  return AddTask(settings_.task_scheduler, pending_requests_,
+                 std::move(introspect_app_task), wrap_callback);
```
andescu:
```suggestion
                 std::move(introspect_app_task), std::move(wrap_callback));
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -769,6 +901,74 @@ client::CancellationToken AuthenticationClient::Impl::SignOut(
...
+    }
+
+    IntrospectAppResult response_result = GetIntrospectAppResult(document);
+    return std::move(response_result);
+  };
+
+  // wrap_callback needed to convert client::ApiError into AuthenticationError.
+  auto wrap_callback = [callback](ResponseType response) {
+    if (!response.IsSuccessful()) {
+      auto error = response.GetError();
```
andescu:
```suggestion
      const auto& error = response.GetError();
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -769,6 +901,74 @@ client::CancellationToken AuthenticationClient::Impl::SignOut(
...
+      AuthenticationError auth_error{error.GetHttpStatusCode(),
+                                     error.GetMessage()};
+      callback(std::move(auth_error));
```
andescu: Instead, probably you can merge and save the extra move:

```suggestion
      callback({error.GetHttpStatusCode(), error.GetMessage()});
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
kzvieriev: Yes, I need to authenticate  using access token somehow. I assume this is a valid way to do it for OlpClient::CallApi().

andescu: At least the app/me is supposed to work without bearer. So if you remove it it is not working?

mykhailo-kuchma: app/me requires access token

kzvieriev: Yeah, you will get authetication error(401 I think).

andescu: Ok. Then let's leave it like this.

PR #699, opened 2020-03-04
Make CacheWithUpdate option unsuported for VersionedLayerClient.

LiubovDidkivska:
Return InvalidArgument for CacheWithUpdate option.
Update tests.

Relates-To: OLPEDGE-1658

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
@@ -164,9 +164,7 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    *
    * @param data_request The `DataRequest` instance that contains a complete set
    * of request parameters.
-   * @note The `GetLayerId` value of the \c DataRequest object is ignored, and
-   * the parameter from the constructor is used instead. CacheWithUpdate fetch
-   * option is not suported.
+   * @note CacheWithUpdate fetch option is not suported.
    * @param callback The `DataResponseCallback` object that is invoked if
    * the `DataResult` object is available or an error is encountered.
    *
@@ -186,9 +184,7 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    *
    * @param data_request The `DataRequest` instance that contains a complete set
    * of request parameters.
-   * @note The `GetLayerId` value of the \c DataRequest object is ignored, and
-   * the parameter from the constructor is used instead. CacheWithUpdate fetch
-   * option is not suported.
+   * @note CacheWithUpdate fetch option is not suported.
    *
    * @return `CancellableFuture` that contains the `DataResponse` instance
    * or an error. You can also use `CancellableFuture` to cancel this request.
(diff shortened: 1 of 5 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -97,6 +97,22 @@ client::CancellationToken VersionedLayerClientImpl::GetPartitions(
                    std::move(partitions_task), std::move(callback));
   };
 
+  if (request.GetFetchOption() == CacheWithUpdate) {
+    request.WithFetchOption(FetchOptions::OnlineIfNotFound);
```
andescu: Why do you need to change FetchOptions to OnlineIfNotFound?

LiubovDidkivska: ScheduleFetch function in case of   if (request.GetFetchOption() == FetchOptions::CacheWithUpdate)  will call schedule_task two times. For returning error is is not needed.

`  if (request.GetFetchOption() == FetchOptions::CacheWithUpdate) {
    auto cache_token =
        schedule_task(request.WithFetchOption(FetchOptions::CacheOnly),
                      std::forward<Callback>(callback));
    auto online_token = schedule_task(
        std::move(request.WithFetchOption(FetchOptions::OnlineOnly)), nullptr);
`

andescu: But then we maybe should not use `ScheduleFetch()` if we need to trick it so that it behaves correctly. Just use AddTask.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -97,6 +97,16 @@ client::CancellationToken VersionedLayerClientImpl::GetPartitions(
                    std::move(partitions_task), std::move(callback));
   };
 
+  if (request.GetFetchOption() == CacheWithUpdate) {
```
andescu: Why don't you move this up top? Why extra create the schedule_get_partitions closure just to not execute it?
Same for below methods.

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
```diff
@@ -186,7 +187,8 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    * @param data_request The `DataRequest` instance that contains a complete set
    * of request parameters.
    * @note The `GetLayerId` value of the \c DataRequest object is ignored, and
-   * the parameter from the constructor is used instead.
+   * the parameter from the constructor is used instead. CacheWithUpdate fetch
+   * option is not suported.
```
andescu: The GetLayerId line is not valid anymore, we removed those APIs.
Please remove that and write only about the fetch option:
"@note FetchOptions::CacheWithUpdate is not supported and will result in an error if used for VersionedLayerClient."

Same for below.

LiubovDidkivska:
> Please also improve the commit message. Write why we do not support it (e.g. version locked in the constructor and cannot be updated, o.s.)
Commit message will be updated using squash with following message. 

Return error in case CacheWithUpdate prefetch option used for VersionуdLayerClient.
CacheWithUpdate fetch option are not supported  for versioned layer. Version is locked in constructor for VersionуdLayerClient, so if data are already in the cache, it makes no sense to update online. Update tests.

PR #702, opened 2020-03-06
Fix flaky test GetVolatilePartitionsExpiry

meravingen123:
VolatileLayerClientCacheTest.GetVolatilePartitionsExpiry test could sometimes
fail due to low expiry time specified for data while performing check that
data is in cache. Expiry time has been increased to 30 seconds. Now after checking
that data is in cache test will loop until data is expired. Check for expiriation
is performed every second for 35 seconds.

Resolves: OLPEDGE-1113


tests/integration/olp-cpp-sdk-dataservice-read/VolatileLayerClientCacheTest.cpp
```diff
@@ -135,13 +122,19 @@ TEST_P(VolatileLayerClientCacheTest, GetVolatilePartitionsExpiry) {
...
   // wait for the layer to expire in cache
-  std::this_thread::sleep_for(std::chrono::seconds(2));
-  request.WithFetchOption(olp::dataservice::read::OnlineIfNotFound);
-  future = catalog_client->GetPartitions(request);
-  partitions_response = future.GetFuture().get();
-
-  ASSERT_TRUE(partitions_response.IsSuccessful())
-      << ApiErrorToString(partitions_response.GetError());
+  bool expired = false;
+  for (;;) {
```
r0busta:
```suggestion
  while (true) {
```
Prefer a `while-statement` to a `for-statement` when there is no obvious loop variable

tests/integration/olp-cpp-sdk-dataservice-read/VolatileLayerClientCacheTest.cpp
```diff
@@ -135,13 +122,19 @@ TEST_P(VolatileLayerClientCacheTest, GetVolatilePartitionsExpiry) {
...
+    std::this_thread::sleep_for(std::chrono::seconds(32));
```
r0busta: Just curious, why 32 and not 42? 😉

r0busta:
Please also elaborate on the fix details in the commit message. (What was happening and why the change is a proper fix to the issue.)

tests/integration/olp-cpp-sdk-dataservice-read/VolatileLayerClientCacheTest.cpp
```diff
@@ -135,13 +122,18 @@ TEST_P(VolatileLayerClientCacheTest, GetVolatilePartitionsExpiry) {
...
+    future = catalog_client->GetPartitions(request);
+    partitions_response = future.GetFuture().get();
+    if (!partitions_response.IsSuccessful()) {
+      expired = true;
+      break;
+    }
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+  }
+  ASSERT_TRUE(expired);
   ASSERT_EQ(0u, partitions_response.GetResult().GetPartitions().size());
```
andescu: I guess you can safely remove this as this is not part of the test itself. You already test in line 130 that the requests was not successful so I would expect that the result is empty.

r0busta:
The commit message's still missing even a basic description of what has happened, how it's fixed, etc.

r0busta:
Thanks for editing the commit message.

PR #705, opened 2020-03-10
SDK should not cache data with FetchOption::OnlineOnly requests.

LiubovDidkivska:
Add Check if fetch_option is not OnlineOnly, and cache only in that case.

Resolves:OLPEDGE-1659

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
@@ -107,7 +107,9 @@ ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApi(
   }
 
   const auto& service_url = api_result.at(0).GetBaseUrl();
-  repository.Put(service, service_version, service_url);
+  if (options != OnlineOnly) {
+    repository.Put(service, service_version, service_url);
+  }
   OLP_SDK_LOG_INFO_F(kLogTag, "LookupApi(%s/%s): %s - OK, service_url=%s",
                      service.c_str(), service_version.c_str(),
                      catalog.partition.c_str(), service_url.c_str());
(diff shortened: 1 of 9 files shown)


olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
```diff
@@ -57,7 +57,7 @@ ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApi(
     client::OlpClientSettings settings) {
   repository::ApiCacheRepository repository(catalog, settings.cache);
 
-  if (options != OnlineOnly) {
+  if (options != OnlineOnly && options != CacheWithUpdate) {
```
andescu: Why also CacheWithUpdate?

LiubovDidkivska: Previously for CacheWithUpdate was splitted to CacheOnly and OnlineOnly. Now with OnlineOnly we are not updating cache. So for updating cache and getting data only online, even if we have valid stored cache we need additional fetch option. And to not brake api, I used CacheWithUpdate for this case.

olp-cpp-sdk-dataservice-read/tests/CatalogRepositoryTest.cpp
```diff
@@ -183,9 +183,9 @@ TEST_F(CatalogRepositoryTest, GetLatestVersionOnlineOnlyFoundAndCacheWritten) {
         return boost::any{};
```
andescu: Test name is then wrong, it says -> GetLatestVersionOnlineOnlyFoundAndCacheWritten
But you are not testing that the cache is written, you actually test that the cache is not written.

olp-cpp-sdk-dataservice-read/tests/CatalogRepositoryTest.cpp
```diff
@@ -351,9 +351,9 @@ TEST_F(CatalogRepositoryTest, GetCatalogOnlineOnlyFoundAndCacheWritten) {
...
```
andescu: Same here. Change test name.

PR #714, opened 2020-03-13
Add custom headers for network.

mykhailo-kuchma:
User can set headers that will be added for each request made by network.
Additionally, it concatenates the user agents set to network and passed with
request.

Relates-To: OLPEDGE-1669

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
@@ -48,11 +48,12 @@
 #ifdef NETWORK_USE_TIMEPROVIDER
 #include "timeprovider/TimeProvider.h"
 #endif
-#include "olp/core/http/NetworkUtils.h"
-#include "olp/core/logging/Log.h"
-#include "olp/core/porting/make_unique.h"
-#include "olp/core/porting/platform.h"
-#include "olp/core/utils/Dir.h"
+#include <olp/core/http/NetworkUtils.h>
+#include <olp/core/http/NetworkConstants.h>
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -218,6 +219,39 @@ static curl_code SslctxFunction(CURL* curl, void* sslctx, void*) {
...
+                                          const std::string& user_agent) {
+  struct curl_slist* headers = nullptr;
+
+  std::ostringstream sstrm;
+
+  for (const auto& header : default_headers) {
+    sstrm.str("");
+    sstrm << header.first;
+    sstrm << ": ";
+    sstrm << header.second;
```
andescu: This seems to me a little overengineered. Why not use just a normal temp string with operator+:

```cpp
std::string curl_header = header.first + ": " + header.second. 
headers = curl_slist_append(headers, curl_header.c_str());
```

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -218,6 +219,39 @@ static curl_code SslctxFunction(CURL* curl, void* sslctx, void*) {
...
+    headers = curl_slist_append(headers, sstrm.str().c_str());
+  }
+
+  for (const auto& header : request_headers) {
+    sstrm.str("");
+    sstrm << header.first;
+    sstrm << ": ";
+    sstrm << header.second;
```
andescu: Same here.

PR #715, opened 2020-03-13
Reduce compiler warnings about deprecated methods/classes.

andescu:
As we deprecated some classes that are public but only to be used
internaly there are a lot of deprecated warnings when building the SDK.
This adds pragma compiler flag disables for the places where we actually
use these classes internaly until we can finaly move these classes to private.
Also this commit removes the version usage from the request key creation and
adds again pragma disables to the repositories which are still using the version
setters/getters internaly. This will be changed in the future but until then
we need to lower the warnings count while building.

Resolves: OLPSUP-9858

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
@@ -245,8 +245,8 @@ class DATASERVICE_READ_API DataRequest final {
       out << GetDataHandle().get();
     }
     out << "]";
-    if (GetVersion()) {
-      out << "@" << GetVersion().get();
+    if (catalog_version_) {
+      out << "@" << catalog_version_.get();
     }
     if (GetBillingTag()) {
       out << "$" << GetBillingTag().get();
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PartitionsRequest.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PartitionsRequest.h
@@ -220,8 +220,8 @@ class DATASERVICE_READ_API PartitionsRequest final {
   std::string CreateKey(const std::string& layer_id) const {
     std::stringstream out;
     out << layer_id;
-    if (GetVersion()) {
-      out << "@" << GetVersion().get();
+    if (catalog_version_) {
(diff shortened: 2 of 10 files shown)


olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/DataRequest.h
```diff
@@ -245,9 +245,6 @@ class DATASERVICE_READ_API DataRequest final {
       out << GetDataHandle().get();
     }
     out << "]";
-    if (GetVersion()) {
-      out << "@" << GetVersion().get();
-    }
```
mykhailo-kuchma: We can't do it, we will lose the information about the version.

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PartitionsRequest.h
```diff
@@ -220,9 +220,6 @@ class DATASERVICE_READ_API PartitionsRequest final {
   std::string CreateKey(const std::string& layer_id) const {
     std::stringstream out;
     out << layer_id;
-    if (GetVersion()) {
-      out << "@" << GetVersion().get();
-    }
```
mykhailo-kuchma: Same here.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -33,6 +33,11 @@
 #include "repositories/PartitionsRepository.h"
 #include "repositories/PrefetchTilesRepository.h"
 
+// Needed to avoid endless warnings from GetVersion/WithVersion
+#include <olp/core/porting/warning_disable.h>
+PORTING_PUSH_WARNINGS()
+PORTING_CLANG_GCC_DISABLE_WARNING("-Wdeprecated-declarations")
+
```
mykhailo-kuchma: We should rather avoid calling GetVersion, and the warning will not be triggered.

andescu: Yes, definitely, but this is a bigger change and is not so easy. We basically need to change all internal classes to not use the version from the request structures anymore. This needs to be done once we remove the deprecated accessors from the request structures.

PR #720, opened 2020-03-17
Add method that removes an expectation by a path-method matcher.

r0busta:
**Add method that removes an expectation by a path-method matcher.**

The method is essential for cleaning expectations on the fly so that
the mock server doesn't eat up all available RAM.

Relates-To: OAM-188

### DIFF
--- a/tests/utils/mock-server-client/Client.h
+++ b/tests/utils/mock-server-client/Client.h
@@ -33,11 +33,14 @@
 
 namespace mockserver {
 
+namespace {
 const auto kBaseUrl = "https://localhost:1080";
 const auto kExpectationPath = "/mockserver/expectation";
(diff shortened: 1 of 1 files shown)


tests/utils/mock-server-client/Client.h
```diff
@@ -127,8 +133,28 @@ inline Status::Ports Client::Ports() const {
...
+  writer.Key("method");
+  writer.String(method_matcher.c_str());
+  writer.Key("path");
+  writer.String(path_matcher.c_str());
+  writer.EndObject();
+
+  const auto data = std::string{buffer.GetString()};
 
-  return;
+  const std::shared_ptr<std::vector<unsigned char>> request_body =
```
andescu:
```suggestion
  const auto request_body =
```

PR #728, opened 2020-03-20
Unit tests for PrefetchRepository.

LiubovDidkivska:
Add tests to check splitting subtree on slices with depth 4.
Add tests to calculate subtree with min/max levels default, or specified.
Test with sibling tiles on the same levels and different.

Resolves:OLPSUP-9847

### DIFF
--- a/olp-cpp-sdk-dataservice-read/tests/PrefetchRepositoryTest.cpp
+++ b/olp-cpp-sdk-dataservice-read/tests/PrefetchRepositoryTest.cpp
@@ -24,93 +24,96 @@
 namespace {
 using namespace olp::dataservice::read;
 
-class PrefetchRepositoryTest : public ::testing::Test,
-                               protected repository::PrefetchTilesRepository {};
+class PrefetchRepositoryTestable
+    : protected repository::PrefetchTilesRepository {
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-read/tests/PrefetchRepositoryTest.cpp
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+#include <gtest/gtest.h>
+
+#include <repositories/PrefetchTilesRepository.h>
+
+namespace {
+using namespace olp::dataservice::read;
+
+class PrefetchRepositoryTest : public ::testing::Test,
```
andescu: Not necessary. Seems most of the method are static either way so just use without fixture.
For this you can easily create a testable:

```cpp
class PrefetchRepositoryTestable : public repository::PrefetchTilesRepository {
  // Expose protected methods to public scope
  using repository::PrefetchTilesRepository::GetSubQuads;
};

TEST(PrefetchRepositoryTest, SplitTreeLevel5) {
  PrefetchRepositoryTestable repository;
  ....
}
```

PR #748, opened 2020-03-30
Add log in case DiskCache open failed

meravingen123:
On Creating Default cache through OlpClientSettingsFactory
there was not check for failure on opening DiskCache. Now in case of
error proper log message is printed.

Resolves: OLPSUP-9914

### DIFF
--- a/olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
+++ b/olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
@@ -22,11 +22,14 @@
 #include "olp/core/cache/CacheSettings.h"
 #include "olp/core/cache/DefaultCache.h"
 #include "olp/core/http/Network.h"
+#include "olp/core/logging/Log.h"
 #include "olp/core/porting/make_unique.h"
 #include "olp/core/thread/ThreadPoolTaskScheduler.h"
 
 namespace olp {
 namespace client {
+
+auto constexpr kLogTag = "OlpClientSettingsFactory";
 
 std::unique_ptr<thread::TaskScheduler>
 OlpClientSettingsFactory::CreateDefaultTaskScheduler(size_t thread_count) {
@@ -42,7 +45,16 @@ OlpClientSettingsFactory::CreateDefaultNetworkRequestHandler(
 std::unique_ptr<cache::KeyValueCache>
 OlpClientSettingsFactory::CreateDefaultCache(cache::CacheSettings settings) {
   auto cache = std::make_unique<cache::DefaultCache>(std::move(settings));
-  cache->Open();
+  auto error = cache->Open();
+  if (error == cache::DefaultCache::StorageOpenResult::OpenDiskPathFailure) {
+    OLP_SDK_LOG_FATAL_F(
+        kLogTag,
+        "Error opening disk cache, disk_path_mutable=%s, "
+        "disk_path_protected=%s",
+        settings.disk_path_mutable.get_value_or("(empty)").c_str(),
+        settings.disk_path_protected.get_value_or("(empty)").c_str());
+    return nullptr;
+  }
   return std::move(cache);
 }
 


olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
```diff
@@ -22,12 +22,15 @@
...
 #include "olp/core/cache/DefaultCache.h"
 #include "olp/core/http/Network.h"
+#include "olp/core/logging/Log.h"
 #include "olp/core/porting/make_unique.h"
 #include "olp/core/thread/ThreadPoolTaskScheduler.h"
 
 namespace olp {
 namespace client {
 
+auto constexpr kLogTag = "Cache";
```
andescu:
```suggestion
auto constexpr kLogTag = "OlpClientSettingsFactory";
```

olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
```diff
@@ -42,7 +45,11 @@ OlpClientSettingsFactory::CreateDefaultNetworkRequestHandler(
 std::unique_ptr<cache::KeyValueCache>
 OlpClientSettingsFactory::CreateDefaultCache(cache::CacheSettings settings) {
   auto cache = std::make_unique<cache::DefaultCache>(std::move(settings));
-  cache->Open();
+  auto error = cache->Open();
+  if (error == cache::DefaultCache::StorageOpenResult::OpenDiskPathFailure)
+  {
+      OLP_SDK_LOG_ERROR(kLogTag, "Error opening disk cache");
+  }
```
andescu: Seems you missed two very important points from the ticket:

"In case the DefaultCache fails to open the mutable/protected cache we need to print a _fatal error_ and return _nullptr_ back so that the user knows something bad happened."

olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
```diff
@@ -42,8 +45,12 @@ OlpClientSettingsFactory::CreateDefaultNetworkRequestHandler(
 std::unique_ptr<cache::KeyValueCache>
 OlpClientSettingsFactory::CreateDefaultCache(cache::CacheSettings settings) {
   auto cache = std::make_unique<cache::DefaultCache>(std::move(settings));
-  cache->Open();
-  return std::move(cache);
+  auto error = cache->Open();
+  if (error == cache::DefaultCache::StorageOpenResult::OpenDiskPathFailure)
+  {
+      OLP_SDK_LOG_FATAL(kLogTag, "Error opening disk cache");
```
andescu: Please print cache path(s) and the open options.

olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
```diff
@@ -42,8 +45,12 @@ OlpClientSettingsFactory::CreateDefaultNetworkRequestHandler(
...
+  }
+  return nullptr;
```
andescu: In case of an open error you should return nullptr only! In case open is successfully you return as before the cache instance.

olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
```diff
@@ -42,7 +45,21 @@ OlpClientSettingsFactory::CreateDefaultNetworkRequestHandler(
...
   auto cache = std::make_unique<cache::DefaultCache>(std::move(settings));
-  cache->Open();
+  auto error = cache->Open();
+  if (error == cache::DefaultCache::StorageOpenResult::OpenDiskPathFailure) {
+    std::string disk_path_mutable = settings.disk_path_mutable
+                                        ? settings.disk_path_mutable.get()
+                                        : "(empty)";
+    std::string disk_path_protected = settings.disk_path_protected
+                                          ? settings.disk_path_protected.get()
+                                          : "(empty)";
```
andescu: Optionally you can also use the exactly for this case exposed boost::optional::get_value_or(), means:

```cpp
auto disk_path_mutable = settings.disk_path_mutable.get_value_or("(empty)");
```

olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
```diff
@@ -42,7 +45,21 @@ OlpClientSettingsFactory::CreateDefaultNetworkRequestHandler(
...
+
+    OLP_SDK_LOG_FATAL_F(
+        kLogTag,
+        "Error opening disk cache. Mutable Path: %s Protected Path: %s",
```
andescu:
```suggestion
        "Error opening disk cache, disk_path_mutable=%s, disk_path_protected=%s",
```

To follow the same pattern as everywhere else.

PR #763, opened 2020-04-06
LRU cache for the mutable disk cache.

kzvieriev:
Added LRU cache is going to be used to control mutable cache eviction.
This change is only preparation and no eviction strategy implemented
yet. New unit tests check DefaultCache functionality is not broken by
different eviction policies.

Resolves: OLPEDGE-1591

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -18,14 +18,26 @@
  */
 
 #include "DefaultCacheImpl.h"
+
+#include <algorithm>
+
 #include "olp/core/logging/Log.h"
 #include "olp/core/porting/make_unique.h"
 
 namespace {
 
 constexpr auto kLogTag = "DefaultCache";
+constexpr auto kExpirySuffix = "::expiry";
+constexpr auto kMaxDiskSize = std::uint64_t(-1);
 
-std::string CreateExpiryKey(const std::string& key) { return key + "::expiry"; }
+std::string CreateExpiryKey(const std::string& key) {
+  return key + kExpirySuffix;
+}
+
+bool IsExpiryKey(const std::string& key) {
+  auto pos = key.rfind(kExpirySuffix);
+  return pos == 0;
+}
 
 time_t GetRemainingExpiryTime(const std::string& key,
                               olp::cache::DiskCache& disk_cache) {
@@ -65,6 +77,7 @@ DefaultCacheImpl::DefaultCacheImpl(CacheSettings settings)
       is_open_(false),
       memory_cache_(nullptr),
       mutable_cache_(nullptr),
+      mutable_cache_lru_(nullptr),
       protected_cache_(nullptr) {}
 
 DefaultCache::StorageOpenResult DefaultCacheImpl::Open() {
@@ -81,6 +94,7 @@ void DefaultCacheImpl::Close() {
 
   memory_cache_.reset();
   mutable_cache_.reset();
+  mutable_cache_lru_.reset();
   protected_cache_.reset();
   is_open_ = false;
 }
@@ -93,6 +107,10 @@ bool DefaultCacheImpl::Clear() {
 
   if (memory_cache_) {
     memory_cache_->Clear();
+  }
+
+  if (mutable_cache_lru_) {
+    mutable_cache_lru_->Clear();
   }
 
   if (mutable_cache_) {
@@ -132,6 +150,17 @@ bool DefaultCacheImpl::Put(const std::string& key, const boost::any& value,
     if (!mutable_cache_->Put(key, encoded_item)) {
       return false;
     }
+
+    if (mutable_cache_lru_) {
+      const auto value = encoded_item.size();
+      auto result = mutable_cache_lru_->InsertOrAssign(key, value);
+      if (result.first == mutable_cache_lru_->end() && !result.second) {
+        OLP_SDK_LOG_WARNING_F(
+            kLogTag, "Failed to store value in mutable LRU cache, key %s",
+            key.c_str());
+        return false;
+      }
+    }
   }
 
   return true;
@@ -167,6 +196,16 @@ bool DefaultCacheImpl::Put(const std::string& key,
     if (!mutable_cache_->Put(key, slice)) {
       return false;
     }
+
+    if (mutable_cache_lru_) {
+      auto result = mutable_cache_lru_->InsertOrAssign(key, value->size());
+      if (result.first == mutable_cache_lru_->end() && !result.second) {
+        OLP_SDK_LOG_WARNING_F(
+            kLogTag, "Failed to store value inmutable LRU cache, key %s",
+            key.c_str());
+        return false;
+      }
+    }
   }
 
   return true;
@@ -178,6 +217,8 @@ boost::any DefaultCacheImpl::Get(const std::string& key,
   if (!is_open_) {
     return boost::any();
   }
+
+  PromoteKeyLru(key);
 
   if (memory_cache_) {
     auto value = memory_cache_->Get(key);
@@ -205,6 +246,8 @@ KeyValueCache::ValueTypePtr DefaultCacheImpl::Get(const std::string& key) {
   if (!is_open_) {
     return nullptr;
   }
+
+  PromoteKeyLru(key);
 
   if (memory_cache_) {
     auto value = memory_cache_->Get(key);
@@ -240,6 +283,10 @@ bool DefaultCacheImpl::Remove(const std::string& key) {
     memory_cache_->Remove(key);
   }
 
+  if (mutable_cache_lru_) {
+    mutable_cache_lru_->Erase(key);
+  }
+
   if (mutable_cache_) {
     if (!mutable_cache_->Remove(key)) {
       return false;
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
```diff
@@ -369,4 +369,266 @@ TEST(DefaultCacheTest, ValueGreaterThanMemCacheLimit) {
...
+  olp::cache::CacheSettings settings;
+  settings.disk_path_mutable = olp::utils::Dir::TempDirectory() + "/unittest";
+  settings.max_memory_cache_size = 0;
+  settings.eviction_policy = olp::cache::EvictionPolicy::kNone;
+  {
+    SCOPED_TRACE("Put/Get decode.");
+    olp::cache::DefaultCache cache(settings);
+    ASSERT_EQ(olp::cache::DefaultCache::Success, cache.Open());
+    ASSERT_TRUE(cache.Clear());
+    std::string key1DataString{"this is key1's data"};
```
andescu, line 384: Here camel-case, below snake-case. Please follow the coding style!
Additionally, I don't see any key2 and key3 and so on. Why do you need to call the var key1?

olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
```diff
@@ -369,4 +369,266 @@ TEST(DefaultCacheTest, ValueGreaterThanMemCacheLimit) {
...
+    bool put_result = cache.Put(
+        "key1", key1DataString, [=]() { return key1DataString; },
+        (std::numeric_limits<time_t>::max)());
```
andescu, line 387: Extract to a constexpr.

olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
```diff
@@ -369,4 +369,266 @@ TEST(DefaultCacheTest, ValueGreaterThanMemCacheLimit) {
...
+    cache.Close();
+
+    ASSERT_EQ(olp::cache::DefaultCache::Success, cache.Open());
+    key1DataRead = cache.Get("key1");
+    ASSERT_TRUE(key1DataRead);
+    ASSERT_EQ(*key1DataRead, binary_data);
+  }
+}
+
+TEST(DefaultCacheTest, EvictionPolicykLeastRecentlyUsed) {
```
andescu, line 505: What exactly are you testing here? I don't see any difference?
I don't see any real unit test here. Unit test should test that the LRU cache is created on open, that it contains all key, that adding new key to the cache also adds it to the LRU, that removing one key from cache also adds it to the LRU. Please implement such a test.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
```diff
@@ -50,6 +50,20 @@ class DefaultCacheImpl {
   bool RemoveKeysWithPrefix(const std::string& key);
 
  private:
+  /// Weight operator assigning the value of 1 to every entry.
+  struct UnitWeightFunction {
+    size_t operator()(const size_t&) const { return 1; }
+  };
```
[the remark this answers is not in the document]
mykhailo-kuchma: Looks like we need a different cache cost function that will return size directly instead of "one".

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -18,15 +18,28 @@
...
 
+bool IsExpiryKey(const std::string& key) {
+  auto expiry_ending = CreateExpiryKey({});
+  if (key.size() < expiry_ending.size()) {
+    return false;
+  }
+  return std::equal(expiry_ending.rbegin(), expiry_ending.rend(),
+                    expiry_ending.rbegin());
+}
+
```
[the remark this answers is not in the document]
kzvieriev: Its used only when initializing, to be sure LRU cache doesn't contain expiry keys. I was following design document for this.

olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
kzvieriev: Oops, my bad.

olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
kzvieriev: Eviction policy is different:-) I have moved tests into separate function.
Existing tests are not covering well mutable cache use, so the idea was to make  couple of tests for each DefaultCache method and check them with different eviction policies.
I can't check is LRU cache actually created and used through DefaultCache interface, only validate nothing is broken with different options.
I think, it can be tested only by eviction strategy tests, but this is out of scope of the PR.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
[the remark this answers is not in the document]
kzvieriev: done, using size as weight

olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
andescu: Unit tests is not an API test. It is a class test. So if you need to access the internals of the DefaultCache in unit tests, then you should do it. Prepare the DefaultCache class so that a derived class can access needed internals, e.g. trough protected section helpers.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -93,6 +93,7 @@ class VolatileLayerClientTest : public ::testing::Test {
 };
 
 TEST_F(VolatileLayerClientTest, GetPartitions) {
+  auto catalog = GetTestCatalog();
```
andescu, line 96: What is this used for?

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -331,9 +416,17 @@ DefaultCacheImpl::GetFromDiscCache(const std::string& key) {
     auto expiry = GetRemainingExpiryTime(key, *mutable_cache_);
     if (expiry <= 0) {
       PurgeDiskItem(key, *mutable_cache_);
+      if (mutable_cache_lru_) {
+        mutable_cache_lru_->Erase(key);
+      }
     } else {
-      auto result = mutable_cache_->Get(key);
+      if (mutable_cache_lru_ &&
```
andescu, line 423: Please move to separate method, e.g. `PromoteKey()`or similar.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
```diff
@@ -50,6 +50,21 @@ class DefaultCacheImpl {
   bool RemoveKeysWithPrefix(const std::string& key);
 
  private:
+  /// Weight operator assigning the value of size to every entry.
+  struct UnitWeightFunction {
+    size_t operator()(const size_t& size) const { return size; }
```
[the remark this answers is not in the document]
kzvieriev: done :-)

olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
kzvieriev: Added some tests to check lru cache directly.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -259,17 +305,67 @@ bool DefaultCacheImpl::RemoveKeysWithPrefix(const std::string& key) {
...
+    if (IsExpiryKey(key.ToString())) {
+      continue;
+    }
+
+    mutable_cache_lru_->Insert(key.ToString(), value.size());
+  }
+}
+
+void DefaultCacheImpl::RemoveKeysWithPrefixLru(const std::string& key) {
+  for (auto it = mutable_cache_lru_->begin();
```
andescu, line 341: In `PromoteKeyLru()` you check if the `mutable_cache_lru` is null, here not. Why?

olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
```diff
@@ -0,0 +1,291 @@
...
+ */
+
+#include <gtest/gtest.h>
+
+#include <cache/DefaultCacheImpl.h>
+#include <olp/core/utils/Dir.h>
+
+using namespace olp::cache;
+
+class DefaultCacheImplHelper : public DefaultCacheImpl {
```
andescu, line 27: Wrap in anonymous namespace.

olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
```diff
@@ -0,0 +1,291 @@
...
+ public:
+  DefaultCacheImplHelper(const CacheSettings& settings)
+      : DefaultCacheImpl(settings){};
+
+  bool HasLruCache() const { return GetMutableCacheLru().operator bool(); }
```
andescu, line 32: This is very strange.
```suggestion
  bool HasLruCache() const { return GetMutableCacheLru() != nullptr; }
```
is probably what you want to do.

PR #765, opened 2020-04-06
Add use_local_time parameter to AuthenticationSettings

meravingen123:
Add use_local_time parameter to AuthenticationSettings to
control if AuthenticationClient will try to get time from server
or use local machine time instead.

Resolves: OLPEDGE-1762

### DIFF
--- a/olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
+++ b/olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
@@ -77,6 +77,12 @@ struct AUTHENTICATION_API AuthenticationSettings {
    * in-memory cache.
    */
   size_t token_cache_limit{100u};
+
+  /**
+   * @brief Uses system system time in authentication requests rather than
+   * requesting time from authentication server.
+   */
+  bool use_system_time{false};
 };
 
 }  // namespace authentication
--- a/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
+++ b/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
@@ -329,6 +329,7 @@ class AuthenticationClient::Impl final {
   using TimeResponse = client::ApiResponse<time_t, client::ApiError>;
   using TimeCallback = std::function<void(TimeResponse)>;
 
+  client::CancellationToken GetTime(TimeCallback callback);
   client::CancellationToken GetTimeFromServer(TimeCallback callback);
   static TimeResponse ParseTimeResponse(std::stringstream& payload);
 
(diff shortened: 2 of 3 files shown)


olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -77,6 +77,12 @@ struct AUTHENTICATION_API AuthenticationSettings {
    * in-memory cache.
    */
   size_t token_cache_limit{100u};
+
+  /**
+   * @brief Use local system time in authentication request rather than requesting
+   * time from authentication server.
+   */
+  bool use_local_time{false};
```
andescu, line 85: I would not call it local time. Local time usually is not UTC and might be confusing, as we do not use actual local time but system time. I would name it `use_system_time`.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -556,6 +557,14 @@ AuthenticationClient::Impl::ParseTimeResponse(std::stringstream& payload) {
   return timestamp_it->value.GetUint();
 }
 
+client::CancellationToken AuthenticationClient::Impl::GetTime(TimeCallback callback) {
+  if (settings_.use_local_time) {
+    callback(std::time(nullptr));
```
andescu, line 562: Use std::chrono::system_time

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -530,7 +531,7 @@ client::CancellationToken AuthenticationClient::Impl::SignInClient(
   };
 
   context.ExecuteOrCancelled(
-      [&]() { return GetTimeFromServer(std::move(time_callback)); }, []() {});
+      [&]() { return GetTime(std::move(time_callback)); }, []() {});
```
andescu, line 534:
```suggestion
      [&]() { return GetTime(std::move(time_callback)); });
```
Cancellation function can be ommited.

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -77,6 +77,12 @@ struct AUTHENTICATION_API AuthenticationSettings {
    * in-memory cache.
    */
   size_t token_cache_limit{100u};
+
+  /**
+   * @brief Use local system time in authentication request rather than requesting
```
andescu, line 82:
```suggestion
   * @brief Use system time in authentication request rather than requesting
```

olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationSettings.h
```diff
@@ -77,6 +77,12 @@ struct AUTHENTICATION_API AuthenticationSettings {
...
+   * time from authentication server.
```
HalynaDumych, line 83:
```suggestion
   * @brief Uses system time in authentication requests rather than requesting
   * time from the authentication server.
```

PR #766, opened 2020-04-06
Add measurement of downloaded/uploaded data Linux.

LiubovDidkivska:
Use curl_easy_getinfo to get size of headers and data
(CURLINFO_HEADER_SIZE, CURLINFO_SIZE_DOWNLOAD).
Get upload size using CURLINFO_REQUEST_SIZE.

Relates-To: OLPEDGE-1759

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
@@ -774,6 +774,9 @@ void NetworkCurl::CompleteMessage(CURL* handle, CURLcode result) {
   int index = GetHandleIndex(handle);
   if (index >= 0 && index < static_cast<int>(handles_.size())) {
     RequestHandle& rhandle = handles_[index];
+    uint64_t uploadBytes = 0;
+    uint64_t downloadBytes = 0;
+    GetTraficData(rhandle.handle, uploadBytes, downloadBytes);
 
     if (rhandle.cancelled) {
       auto callback = rhandle.callback;
@@ -781,7 +784,9 @@ void NetworkCurl::CompleteMessage(CURL* handle, CURLcode result) {
           NetworkResponse()
               .WithRequestId(rhandle.id)
               .WithStatus(static_cast<int>(ErrorCode::CANCELLED_ERROR))
-              .WithError("Cancelled");
+              .WithError("Cancelled")
+              .WithBytesDownloaded(downloadBytes)
+              .WithBytesUploaded(uploadBytes);
       ReleaseHandleUnlocked(&rhandle);
 
       lock.unlock();
@@ -855,7 +860,9 @@ void NetworkCurl::CompleteMessage(CURL* handle, CURLcode result) {
     auto response = NetworkResponse()
                         .WithRequestId(handles_[index].id)
                         .WithStatus(status)
-                        .WithError(error);
+                        .WithError(error)
+                        .WithBytesDownloaded(downloadBytes)
+                        .WithBytesUploaded(uploadBytes);
     ReleaseHandleUnlocked(&rhandle);
     lock.unlock();
     callback(response);
@@ -959,6 +966,10 @@ void NetworkCurl::Run() {
       while (IsStarted() &&
              (msg = curl_multi_info_read(curl_, &msgs_in_queue))) {
         CURL* handle = msg->easy_handle;
+        uint64_t uploadBytes = 0;
+        uint64_t downloadBytes = 0;
+        GetTraficData(handle, uploadBytes, downloadBytes);
+
         if (msg->msg == CURLMSG_DONE) {
           CURLcode result = msg->data.result;
           curl_multi_remove_handle(curl_, handle);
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/src/http/curl/NetworkCurl.h
```diff
@@ -277,6 +277,14 @@ class NetworkCurl : public olp::http::Network,
...
    */
   inline bool IsStarted() const;
+  /**
+   * @brief CURL get upload/download data.
+   * @param[in] handle CURL easy handle.
+   * @param[out] uploadBytes uploaded bytes(headers+data).
+   * @param[out] downloadBytes downloaded bytes(headers+data).
+   */
+  static void GetTraficData(CURL* handle, uint64_t& uploadBytes,
+                            uint64_t& downloadBytes);
```
andescu, line 287: If it is static then consider converting it to an anonymous function.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -1058,5 +1071,25 @@ void NetworkCurl::Run() {
...
+    OLP_SDK_LOG_DEBUG(kLogTag, "downloaded bytes " << downloadBytes);
+  }
+
+  long lengthUpload;
+  if (curl_easy_getinfo(handle, CURLINFO_REQUEST_SIZE, &lengthUpload) ==
+      CURLE_OK) {
+    uploadBytes = lengthUpload;
+    OLP_SDK_LOG_DEBUG(kLogTag, "uploaded bytes " << uploadBytes);
+  }
+}
```
andescu, line 1092: Please adapt to coding style.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -1058,5 +1071,25 @@ void NetworkCurl::Run() {
...
+void NetworkCurl::GetTraficData(CURL* handle, uint64_t& uploadBytes,
+                                uint64_t& downloadBytes) {
+  long headersSize;
+  double lengthDownloaded;
+  if (curl_easy_getinfo(handle, CURLINFO_HEADER_SIZE, &headersSize) ==
+          CURLE_OK &&
+      curl_easy_getinfo(handle, CURLINFO_SIZE_DOWNLOAD, &lengthDownloaded) ==
+          CURLE_OK) {
+    downloadBytes = lengthDownloaded + headersSize;
+    OLP_SDK_LOG_DEBUG(kLogTag, "downloaded bytes " << downloadBytes);
```
andescu, line 1083: Log without any additional information will be very hard to read. At least the url or the request id needs to be also there so that we can associate it somehow to an existing request.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -1058,5 +1071,25 @@ void NetworkCurl::Run() {
...
 
+void NetworkCurl::GetTraficData(CURL* handle, uint64_t& uploadBytes,
+                                uint64_t& downloadBytes) {
+  long headersSize;
+  double lengthDownloaded;
+  if (curl_easy_getinfo(handle, CURLINFO_HEADER_SIZE, &headersSize) ==
+          CURLE_OK &&
+      curl_easy_getinfo(handle, CURLINFO_SIZE_DOWNLOAD, &lengthDownloaded) ==
+          CURLE_OK) {
+    downloadBytes = lengthDownloaded + headersSize;
```
andescu, line 1082: Maybe a fallback should be in place in case curl_easy_getinfo is not delivering responses, e.g. mark the payload size and the headers size in RequestHandle.count before you send the request.

LiubovDidkivska: Do you mean that we should count upload bytes before curl_multi_perform?

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
LiubovDidkivska: removed logging

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
andescu: Just in case. We should still use curl_easy_getinfo, but if those calls fail we still have a callback.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -176,6 +176,30 @@ static curl_code SslctxFunction(CURL* curl, void* sslctx, void*) {
...
+void GetTraficData(CURL* handle, uint64_t& upload_bytes,
+                   uint64_t& download_bytes) {
+  long headers_size;
+  double length_downloaded;
+  if (curl_easy_getinfo(handle, CURLINFO_HEADER_SIZE, &headers_size) ==
+          CURLE_OK &&
+      curl_easy_getinfo(handle, CURLINFO_SIZE_DOWNLOAD, &length_downloaded) ==
+          CURLE_OK) {
+    download_bytes = length_downloaded + headers_size;
+  }
```
mykhailo-kuchma, line 194: I would rather separate it, currently, if the first call failed, the second one is not called.

LiubovDidkivska: it was done intentionally. Why we should count any data, if some fail happen?

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
LiubovDidkivska: Theoretically curl_easy_getinfo should always be successful  if all parameters ok. On send headers are set internally by curl_easy_setopt. So we can count only user headers. The same on response. So maybe we should do fallback for all platforms, if upload/download is 0, but data received is not 0.

PR #767, opened 2020-04-06
Calculate total bytes sent/received on Windows

mykhailo-kuchma:
Calculations are approximate, number bytes sent is calculated as a
size of body and a size of headers. Number of bytes received is
calculated as size of headers plus content length if available, else
a sum of chunks sizes.

Relates-To: OLPEDGE-1760

### DIFF
--- a/olp-cpp-sdk-core/src/http/winhttp/NetworkWinHttp.cpp
+++ b/olp-cpp-sdk-core/src/http/winhttp/NetworkWinHttp.cpp
@@ -199,6 +199,37 @@ std::wstring ProxyString(const olp::http::NetworkProxySettings& proxy) {
   return proxy_string_stream.str();
 }
 
+DWORD QueryHeadersSize(HINTERNET handle) {
+  DWORD wide_len = 0;
+  WinHttpQueryHeaders(handle, WINHTTP_QUERY_RAW_HEADERS,
+                      WINHTTP_HEADER_NAME_BY_INDEX, WINHTTP_NO_OUTPUT_BUFFER,
+                      &wide_len, WINHTTP_NO_HEADER_INDEX);
+  if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+    return 0;
+  }
+  return wide_len / sizeof(WCHAR);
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/src/http/winhttp/NetworkWinHttp.cpp
```diff
@@ -199,6 +199,37 @@ std::wstring ProxyString(const olp::http::NetworkProxySettings& proxy) {
...
 
+DWORD QueryHeadersSize(HINTERNET handle) {
+  DWORD wide_len = 0;
+  WinHttpQueryHeaders(handle, WINHTTP_QUERY_RAW_HEADERS,
+                      WINHTTP_HEADER_NAME_BY_INDEX, WINHTTP_NO_OUTPUT_BUFFER,
+                      &wide_len, WINHTTP_NO_HEADER_INDEX);
+  if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+    return 0;
+  }
+  return wide_len / 2;
```
kzvieriev, line 210: Why do we divide by 2? Is it sizeof(WCHAR)?

olp-cpp-sdk-core/src/http/winhttp/NetworkWinHttp.cpp
```diff
@@ -599,47 +636,37 @@ void NetworkWinHttp::RequestCallback(HINTERNET, DWORD_PTR context, DWORD status,
...
             }
+            index++;
+            start = index;
+          } else {
+            index++;
           }
         }
+        headers_size += headers.size();
+      } else {
+        headers_size += QueryHeadersSize(handle->http_request);
```
kzvieriev, line 664: If I understand correctly, QueryHeadersSize returns length  of the request, not size in bytes, so probably need to multiply on `sizeof(WCHAR)`

mykhailo-kuchma: It's intentionally since winapi converts the received data to wchar

PR #771, opened 2020-04-08
Implement buckets for DefaultNetwork

mykhailo-kuchma:
Track the number of requests and metrics. Users can now track the
network in/out number of bytes. Calculations rely on network impl's
to provide the correct download/upload numbers in network response.

Relates-To: OLPEDGE-1754

### DIFF
--- a/olp-cpp-sdk-core/src/http/DefaultNetwork.cpp
+++ b/olp-cpp-sdk-core/src/http/DefaultNetwork.cpp
@@ -20,13 +20,14 @@
 #include <algorithm>
 
 #include "DefaultNetwork.h"
+#include "olp/core/http/HttpStatusCode.h"
 #include "olp/core/http/NetworkConstants.h"
 #include "olp/core/http/NetworkUtils.h"
 
 namespace olp {
 namespace http {
 DefaultNetwork::DefaultNetwork(std::shared_ptr<Network> network)
-    : network_{std::move(network)} {}
+    : current_statistics_bucket_{0}, network_{std::move(network)} {}
 
 DefaultNetwork::~DefaultNetwork() = default;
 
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/src/http/DefaultNetwork.h
```diff
@@ -55,14 +58,29 @@ class DefaultNetwork final : public Network {
...
+  /// Implements the `GetStatistics` method of the `Network` class.
+  Statistics GetStatistics(uint8_t bucket_id) override;
+
  private:
   void AppendUserAgent(Headers& request_headers) const;
   void AppendDefaultHeaders(Headers& request_headers) const;
 
-  std::shared_ptr<Network> network_;
+  void LockStatistics(uint8_t bucket_id,
+                      std::function<void(Statistics&)>&& callback);
```
andescu, line 72: Not sure if you need rvalue here. Probably pass by copy should also have the same effect if you only use it internally by reference.
```suggestion
                      std::function<void(Statistics&)> callback);
```

olp-cpp-sdk-core/src/http/DefaultNetwork.cpp
```diff
@@ -78,5 +110,12 @@ void DefaultNetwork::AppendDefaultHeaders(Headers& request_headers) const {
                          default_headers_.end());
 }
 
+void DefaultNetwork::LockStatistics(
+    uint8_t bucket_id, std::function<void(Statistics&)>&& callback) {
+  buckets_.locked([bucket_id, &callback](BucketsContainer& container) {
```
andescu, line 115:
```suggestion
  buckets_.locked([&](BucketsContainer& container) {
```
Both can be accessed by reference.

tests/integration/olp-cpp-sdk-core/DefaultNetworkTest.cpp
```diff
@@ -134,4 +154,146 @@ TEST(DefaultNetworkTest, Send) {
   }
 }
 
+TEST(DefaultNetworkTest, DefaultBucket) {
+  SCOPED_TRACE(
+      "Network response download/upload are appended to default bucket");
```
andescu, line 159: No need for such big sentences, just use "encrypted" messages, e.g. "Default bucket used"

tests/integration/olp-cpp-sdk-core/DefaultNetworkTest.cpp
```diff
@@ -134,4 +154,146 @@ TEST(DefaultNetworkTest, Send) {
...
+  EXPECT_EQ(network_response.GetBytesUploaded(), 50ull);
+
+  auto stats = default_network_adapter->GetStatistics();
+  EXPECT_TRUE(CompareStatistics(stats, Statistics(100ull, 50ull, 1u, 0u)));
+}
+
+TEST(DefaultNetworkTest, BucketSelection) {
+  SCOPED_TRACE(
+      "Network response download/upload are appended to the active bucket "
+      "when requests is fired");
```
andescu, line 198: "Active bucket used"

PR #772, opened 2020-04-09
VolatileLayerClient::PrefetchTiles fv tests.

kzvieriev:
Functional tests for VolatileLayerClient::PrefetchTiles() method. New
catalog/credentials created for the tests, since Read/Write volatile
layer with HERETile scheme required.

Resolves: OLPEDGE-1738

### DIFF
--- a/scripts/linux/fv/olp-cpp-sdk-functional-test.variables
+++ b/scripts/linux/fv/olp-cpp-sdk-functional-test.variables
@@ -22,3 +22,8 @@ export dataservice_read_volatile_test_appid="${dataservice_read_volatile_test_ap
 export dataservice_read_volatile_test_secret="${dataservice_read_volatile_test_secret}"
 export dataservice_read_volatile_layer="test-volatile-layer"
 export dataservice_read_volatile_test_catalog="hrn:here:data::olp-here-test:here-olp-sdk-cpp"
+
+export dataservice_read_volatile_test_prefetch_appid="${dataservice_read_volatile_test_prefetch_appid}"
+export dataservice_read_volatile_test_prefetch_secret="${dataservice_read_volatile_test_prefetch_secret}"
+export dataservice_read_volatile_test_prefetch_catalog="hrn:here:data::olp-here-test:test-volatile-prefetch"
+export dataservice_read_volatile_prefetch_layer="volatile_prefetch_layer"
--- a/tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
+++ b/tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
@@ -31,12 +31,17 @@
 #include <olp/core/client/OlpClientSettingsFactory.h>
 #include <olp/core/logging/Log.h>
 #include <olp/core/porting/make_unique.h>
+#include <olp/dataservice/write/VolatileLayerClient.h>
+#include <olp/dataservice/write/model/PublishPartitionDataRequest.h>
+#include <olp/dataservice/write/model/StartBatchRequest.h>
 #include <testutils/CustomParameters.hpp>
 
 #include "olp/dataservice/read/PartitionsRequest.h"
+#include "olp/dataservice/read/PrefetchTileResult.h"
 #include "olp/dataservice/read/VolatileLayerClient.h"
 
 using namespace olp::dataservice::read;
+using namespace olp::dataservice::write::model;
 using namespace testing;
 
 namespace {
@@ -46,6 +51,19 @@ const auto kAppSecretEnvName = "dataservice_read_volatile_test_secret";
 const auto kCatalogEnvName = "dataservice_read_volatile_test_catalog";
 const auto kLayerEnvName = "dataservice_read_volatile_layer";
 
+const auto kPrefetchAppId = "dataservice_read_volatile_test_prefetch_appid";
+const auto kPrefetchAppSecret =
+    "dataservice_read_volatile_test_prefetch_secret";
+const auto kPrefetchCatalog = "dataservice_read_volatile_test_prefetch_catalog";
+const auto kPrefetchLayer = "dataservice_read_volatile_prefetch_layer";
+const auto kPrefetchTile = "23618401";
+const auto kPrefetchSubTile1 = "23618410";
+const auto kPrefetchSubTile2 = "23618406";
+const auto kPrefetchEndpoint = "endpoint";
+
+// The limit for 100 retries is 10 minutes. Therefore, the wait time between
+// retries is 6 seconds.
+const auto kWaitBeforeRetry = std::chrono::seconds(6);
 const auto kTimeout = std::chrono::seconds(5);
 
 class VolatileLayerClientTest : public ::testing::Test {
@@ -69,6 +87,27 @@ class VolatileLayerClientTest : public ::testing::Test {
     olp::cache::CacheSettings cache_settings;
     settings_.cache = olp::client::OlpClientSettingsFactory::CreateDefaultCache(
         cache_settings);
+
+    // prefetch setup
+    auto prefetch_app_id = CustomParameters::getArgument(kPrefetchAppId);
+    auto prefetch_secret = CustomParameters::getArgument(kPrefetchAppSecret);
+    prefetch_catalog_ = CustomParameters::getArgument(kPrefetchCatalog);
+    prefetch_layer_ = CustomParameters::getArgument(kPrefetchLayer);
+
+    olp::authentication::Settings prefetch_auth_settings(
+        {prefetch_app_id, prefetch_secret});
+    prefetch_auth_settings.token_endpoint_url =
+        CustomParameters::getArgument(kPrefetchEndpoint);
+    prefetch_auth_settings.network_request_handler = network;
+
+    olp::client::AuthenticationSettings prefetch_auth_client_settings;
+    prefetch_auth_client_settings.provider =
+        olp::authentication::TokenProviderDefault(prefetch_auth_settings);
+
+    prefetch_settings_.authentication_settings = prefetch_auth_client_settings;
+    prefetch_settings_.network_request_handler = network;
+    prefetch_settings_.task_scheduler =
+        olp::client::OlpClientSettingsFactory::CreateDefaultTaskScheduler(1);
   }
 
   void TearDown() override {
@@ -77,6 +116,7 @@ class VolatileLayerClientTest : public ::testing::Test {
     // anywhere. Also network is still used in authentication settings and in
     // TokenProvider internally so it needs to be cleared.
     settings_ = olp::client::OlpClientSettings();
+    prefetch_settings_ = olp::client::OlpClientSettings();
     EXPECT_EQ(network.use_count(), 1);
   }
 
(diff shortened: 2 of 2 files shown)


tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,244 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
   }
 }
 
+TEST_F(VolatileLayerClientTest, Prefetch) {
+  const auto kTileId = GetTestTile();
+  auto layer = CustomParameters::getArgument(kPrefetchLayer);
+  auto catalog = CustomParameters::getArgument(kPrefetchCatalog);
+
+  PrepareForPrefetch();
```
meravingen123, line 349: Can this be done in fixture SetUp?

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,244 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
...
+      PrefetchTilesResponse response = future.get();
+      EXPECT_TRUE(response.IsSuccessful());
+      const auto& result = response.GetResult();
+
+      for (auto tile_result : result) {
+        EXPECT_TRUE(tile_result->IsSuccessful());
+        ASSERT_TRUE(tile_result->tile_key_.IsValid());
+      }
+
+      ASSERT_EQ(tile_keys.size(), result.size());
```
meravingen123, line 492: Due to min level is greater than max this must fail but there are checks for successful fetch. Is this correct?

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
kzvieriev: Prefetch tests need different setup. They use another catalog, credentials etc. + we need to write tiles data used only by these tests, so moving it to SetUp() will lead to overhead for other test.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
kzvieriev: min/max range is invalid if `min>max`, for such case we load only tiles from the request.
Actually, 0 and values > TileKey::MaxLevel(31) are invalid, so this test more like checking invalid levels behavior.
I will add test with valid levels but min>max :-)

kzvieriev: done.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -46,6 +51,19 @@ const auto kAppSecretEnvName = "dataservice_read_volatile_test_secret";
 const auto kCatalogEnvName = "dataservice_read_volatile_test_catalog";
 const auto kLayerEnvName = "dataservice_read_volatile_layer";
 
+const auto kPrefetchAppId = "dataservice_read_volatile_test_prefetch_appid";
```
LiubovDidkivska, line 54: could we have one app id and secret for all tests in VolatileLayerClientTest? So other user can set only once their id/secret.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -88,6 +106,105 @@ class VolatileLayerClientTest : public ::testing::Test {
     return CustomParameters::getArgument(kLayerEnvName);
   }
 
+  std::string GetTestTile() {
+    return CustomParameters::getArgument(kTileEnvName);
+  }
+
+  void PrepareForPrefetch() {
+    auto app_id = CustomParameters::getArgument(kPrefetchAppId);
```
LiubovDidkivska, line 114: Also could we use SetUp function for prefetch tests also.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
kzvieriev: Will need to change old tests, so they use new catalog, layers etc.
Old credentials/catalog can't be used in new tests, since need to write data with tiles first.

Anyway, there is a separated task to create accessible catalogs for all tests. After that we could use same credentials for all tests here.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
kzvieriev: Same reason here, prefetch tests require tiles in the layer.
Adding data isn't fast process, so I prefer to make this preparation only for these new tests. What do you think about it?

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,273 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
   }
 }
 
+TEST_F(VolatileLayerClientTest, Prefetch) {
+  const auto kTileId = GetTestTile();
+  auto layer = CustomParameters::getArgument(kPrefetchLayer);
+  auto catalog = CustomParameters::getArgument(kPrefetchCatalog);
+
+  PrepareForPrefetch();
```
andescu, line 349: This looks very strange that you have a function that does everything and takes the same env variables again.
Why dont you make something by using the fixture where the needed parameters are all loaded at StartUp()?

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,273 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
...
+                       .WithTileKeys(tile_keys)
+                       .WithMinLevel(10)
+                       .WithMaxLevel(12);
+
+    std::promise<PrefetchTilesResponse> promise;
+    std::future<PrefetchTilesResponse> future = promise.get_future();
+    auto token = client.PrefetchTiles(
+        request, [&promise](PrefetchTilesResponse response) {
+          promise.set_value(std::move(response));
+        });
```
andescu, line 368: If you already use the future based approach then use the PrefetchTiles() that returnes the CancellableFuture.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,273 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
...
+    ASSERT_NE(future.wait_for(kTimeout), std::future_status::timeout);
+    PrefetchTilesResponse response = future.get();
+    EXPECT_TRUE(response.IsSuccessful());
+    ASSERT_FALSE(response.GetResult().empty());
+
+    const auto& result = response.GetResult();
+
+    for (auto tile_result : result) {
+      EXPECT_TRUE(tile_result->IsSuccessful());
+      ASSERT_TRUE(tile_result->tile_key_.IsValid());
```
andescu, line 379: But you should know which tiles you expect. Why do you only test if it is valid when you should be testing that expected tiles have been downloaded. Please check the tiles that were downloaded.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,273 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
...
+    DataResponse response;
+    auto token = client.GetData(
+        olp::dataservice::read::DataRequest()
+            .WithPartitionId(kTileId)
+            .WithFetchOption(CacheOnly),
+        [&response](DataResponse resp) { response = std::move(resp); });
+    EXPECT_TRUE(response.IsSuccessful());
+    ASSERT_TRUE(response.GetResult() != nullptr);
+    ASSERT_NE(response.GetResult()->size(), 0u);
+  }
```
andescu, line 396: Do we need such a functional test. This can be, resp. should be, an integration test.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,273 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
...
+    DataResponse response;
+    auto token = client.GetData(
+        olp::dataservice::read::DataRequest()
+            .WithPartitionId(sub_partition1)
+            .WithFetchOption(CacheOnly),
+        [&response](DataResponse resp) { response = std::move(resp); });
+    EXPECT_TRUE(response.IsSuccessful());
+    ASSERT_TRUE(response.GetResult() != nullptr);
+    ASSERT_NE(response.GetResult()->size(), 0u);
+  }
```
andescu, line 410: Same here, should be an integration test.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,273 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
...
+            .WithPartitionId(sub_partition2)
+            .WithFetchOption(CacheOnly),
+        [&response](DataResponse resp) { response = std::move(resp); });
+    EXPECT_TRUE(response.IsSuccessful());
+    ASSERT_TRUE(response.GetResult() != nullptr);
+    ASSERT_NE(response.GetResult()->size(), 0u);
+  }
+}
+
+TEST_F(VolatileLayerClientTest, PrefetchWrongLevels) {
```
andescu, line 427: These are all integration tests.
The only test you need to do extra is that in case min/max are equal or 0, that only the requested tilekeys are downloaded.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
```diff
@@ -224,4 +341,273 @@ TEST_F(VolatileLayerClientTest, GetPartitionsDifferentFetchOptions) {
...
+        EXPECT_TRUE(tile_result->IsSuccessful());
+        ASSERT_TRUE(tile_result->tile_key_.IsValid());
+      }
+
+      ASSERT_EQ(tile_keys.size(), result.size());
+    }
+  }
+}
+
+TEST_F(VolatileLayerClientTest, PrefetchWithCancellableFuture) {
```
andescu, line 578: You are testing too much. You don't need to replicate the integration tests here. You only need to test the real functionality with OLP, resp. with the mocked server later on. So that you have covered the e2e scenario.
You don't need to replicate the entire integration tests here. This is not the point.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
kzvieriev: yeap. integration tests has same test. Removed.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
kzvieriev: Removed also.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
kzvieriev: Removed.
Actually, I took them from VersionedLayerClient prefetch fv tests. So they need cleanup too.

tests/functional/olp-cpp-sdk-dataservice-read/DataserviceReadVolatileLayerClientTest.cpp
kzvieriev: Removed.
Extra tests for 0/0 and same levels added.

scripts/linux/fv/olp-cpp-sdk-functional-test.variables
```diff
@@ -22,3 +22,8 @@ export dataservice_read_volatile_test_appid="${dataservice_read_volatile_test_ap
 export dataservice_read_volatile_test_secret="${dataservice_read_volatile_test_secret}"
 export dataservice_read_volatile_layer="test-volatile-layer"
 export dataservice_read_volatile_test_catalog="hrn:here:data::olp-here-test:here-olp-sdk-cpp"
+
+export dataservice_read_volatile_test_prefetch_appid="{dataservice_read_volatile_test_prefetch_appid}"
+export dataservice_read_volatile_test_prefetch_secret="{dataservice_read_volatile_test_prefetch_secret}"
+export dataservice_read_volatile_test_prefetch_catalog=hrn:here:data::olp-here-test:test-volatile-prefetch"
```
andescu, line 28:
```suggestion
export dataservice_read_volatile_test_prefetch_catalog="hrn:here:data::olp-here-test:test-volatile-prefetch"
```

PR #780, opened 2020-04-13
Enable deprecated-declaration warning as error

mykhailo-kuchma:
Removed AuthenticationCredentials explicit copy ctor.
Removed ApiResponse explicit copy ctor.
Updated many places with new AuthentionClient ctor.
Tests that are still use deprecated API have warnings disabled.
Change make_unique to make_shared since the warning was generated in
make_unique template function.

Relates-To: OLPEDGE-1800

### DIFF
--- a/olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationCredentials.h
+++ b/olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationCredentials.h
@@ -96,15 +96,6 @@ class AUTHENTICATION_API AuthenticationCredentials {
   AuthenticationCredentials(std::string key, std::string secret);
 
   /**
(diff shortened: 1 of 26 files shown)


olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationCredentials.h
```diff
@@ -95,15 +95,6 @@ class AUTHENTICATION_API AuthenticationCredentials {
    */
   AuthenticationCredentials(std::string key, std::string secret);
```
andescu, line 96: Mark as explicit?

mykhailo-kuchma: There is no need since there are 2 parameters already.

PR #789, opened 2020-04-16
Use shared_ptr to settings for prefetch tasks

mykhailo-kuchma:
As was found during profiling, prefetch spawn too many tasks, and
settings are copied to every spawned task, as a result, there is
520 bytes overhead for each task, using a shared pointer to settings
we save this memory.

Relates-To: OLPEDGE-1804

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -292,6 +292,12 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchTiles(
           return false;
         };
 
+        // Settings structure consumes a 536 bytes of heap memory when captured
+        // in lambda, shared pointer (16 bytes) saves 520 bytes of heap memory.
+        // When users prefetch few hundreds tiles it could save few mb.
+        auto shared_settings =
+            std::make_shared<client::OlpClientSettings>(settings);
+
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -305,6 +305,11 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchTiles(
           futures->emplace_back(promise->get_future());
           auto context_it = contexts.emplace(contexts.end());
 
+          // Settings structure consumes a noticeable number of bytes in the
+          // task, shared pointer saves a lot of runtime memory.
+          auto shared_settings =
+              std::make_shared<client::OlpClientSettings>(settings);
```
kzvieriev, line 311: `settings` is just a copy of client member `settings_`. Maybe, it makes sense to create shared_settings from it at the begging of the PrefetchTiles().

mykhailo-kuchma: We try to achieve stateless client, in case we capture **this**, later on they might be changed in runtime, it will break the existing functionality.

kzvieriev: I mean, you are making second copy of client's settings. I suggest not capturing `this`, but remove variable `settings` (not `settings_`)at all and use shared_settings instead.
Do we actually need 2 copies?Am i missing something?

PR #797, opened 2020-04-22
Add ccache reset condition for every week

ystefinko:
Every week ccache will be refreshed on PR and master builds
of 'Linux Clang Build' job.
So build time decrease from 9 to 3 mins.
RestoreKeys are needed for partial cache match.
Azure doc is : https://bit.ly/3bxSHcz

Resolves: OLPEDGE-1731


ystefinko:
ready for review

PR #805, opened 2020-04-24
Eviction for mutable cache.

kzvieriev:
If mutable cache reached max_disk_storage size, user will not be able
to put new values. Removing least recently used items allows to use the
cache further.

Eviction process strarts after 90% threshold and goes till 85%, so all
values can be removed in one batch and number of disk calls is
minimized.

Resolves: OLPEDGE-1592, OLPEDGE-1822

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -30,6 +30,8 @@ namespace {
 constexpr auto kLogTag = "DefaultCache";
 constexpr auto kExpirySuffix = "::expiry";
 constexpr auto kMaxDiskSize = std::uint64_t(-1);
+constexpr auto kMinDiskUsedThreshold = 0.85f;
+constexpr auto kMaxDiskUsedThreshold = 0.9f;
 
 // current epoch time contains 10 digits.
 constexpr auto kExpiryValueSize = 10;
@@ -73,14 +75,12 @@ void PurgeDiskItem(const std::string& key, olp::cache::DiskCache& disk_cache,
   removed_data_size += data_size;
 }
 
-size_t StoreExpiry(const std::string& key, olp::cache::DiskCache& disk_cache,
+size_t StoreExpiry(const std::string& key, leveldb::WriteBatch& batch,
                    time_t expiry) {
   auto expiry_key = CreateExpiryKey(key);
   auto time_str = std::to_string(
       expiry + olp::cache::InMemoryCache::DefaultTimeProvider()());
-  if (!disk_cache.Put(expiry_key, time_str)) {
-    return 0;
-  }
+  batch.Put(expiry_key, time_str);
 
   return expiry_key.size() + time_str.size();
 }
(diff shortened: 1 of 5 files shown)


kzvieriev:
Added fix for not removing expiry data on `DefaultCacheImpl::Remove()` and test for it.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -371,6 +358,53 @@ bool DefaultCacheImpl::PromoteKeyLru(const std::string& key) {
...
+    evicted += key.size() + it->value();
+    ++count;
+
+    if (memory_cache_) {
+      memory_cache_->Remove(it->key());
+    }
+
+    batch.Delete(key);
+    mutable_cache_lru_->Erase(it);
+    it = mutable_cache_lru_->rbegin();
```
andescu, line 394: mutable_cache_lru->Erase() returns the next iterator:

```cpp
const_iterator Erase(const_iterator& it) {
    auto prev = it++;
    Erase(prev->key());
    return it;
  }
```

so you should be able to do:

```cpp
it = mutable_cache_lru_->Erase(it);
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -371,6 +358,53 @@ bool DefaultCacheImpl::PromoteKeyLru(const std::string& key) {
...
+  }
+
+  int64_t elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
+                        std::chrono::steady_clock::now() - start)
+                        .count();
+  OLP_SDK_LOG_DEBUG_F(kLogTag,
```
andescu, line 400:
```suggestion
  OLP_SDK_LOG_INFO_F(kLogTag,
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -371,6 +358,53 @@ bool DefaultCacheImpl::PromoteKeyLru(const std::string& key) {
...
+                      "Lru cache evicted: items=%" PRId32 ", time=%" PRId64
```
andescu, line 401:
```suggestion
                      "Eviced from mutable cache, items=%" PRId32 ", time=%" PRId64
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
kzvieriev: well yes, but actually no :-)
The problem here - lru cache has no reverse_iterator thing.
We are removing least used - means the last element. Next element after the last one is mutable_cache_lru_->End(), not new 'last' element.

andescu: Ok

tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerClientTest.cpp
```diff
@@ -2992,6 +2999,102 @@ TEST_F(DataserviceReadVersionedLayerClientTest, CheckLookupApiCacheExpiration) {
...
+    ASSERT_TRUE(future.get().IsSuccessful());
+
+    // promote the key so its not evicted.
+    ASSERT_TRUE(is_in_cache(promote_key));
+  }
+
+  // maximum is reached.
+  ASSERT_TRUE(count == max_count);
+  EXPECT_TRUE(is_in_cache(promote_key));
+  EXPECT_TRUE(is_in_cache(promote_key));
```
andescu, line 3094: Duplicate?

kzvieriev: Oops, it should test last added element is cached. Fixed!

PR #813, opened 2020-04-28
Adapt core/utils classes to coding style.

andescu:
All core/utils classes are now adapted to our coding style, except
some Windows specific code from Dir.cpp.

Relates-To: OLPEDGE-1854

### DIFF
--- a/olp-cpp-sdk-core/src/utils/Dir.cpp
+++ b/olp-cpp-sdk-core/src/utils/Dir.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 HERE Europe B.V.
+ * Copyright (C) 2019-2020 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-core/src/utils/Dir.cpp
```diff
@@ -339,7 +349,8 @@ std::string Dir::TempDirectory() {
     // it should never reach this path unless a big change in iOS
     ret = "/private/tmp";
   } else {
-    if (ret[ret.size() - 1] == '/') ret.pop_back();
+    if (ret[ret.size() - 1] == '/')
+      ret.pop_back();
```
LiubovDidkivska, line 353: do we need braces here?
` if (ret[ret.size() - 1] == '/') {
      ret.pop_back();
}`

olp-cpp-sdk-core/src/utils/Dir.cpp
```diff
@@ -250,7 +253,8 @@ bool Dir::remove(const std::string& path) {
   const TCHAR* n_path = path.c_str();
 #endif  // _UNICODE
 
-  if (!DeleteDirectory(n_path)) ret = false;
+  if (!DeleteDirectory(n_path))
+    ret = false;
```
LiubovDidkivska, line 257:
```suggestion
  if (!DeleteDirectory(n_path)) {
      ret = false;
    }
```

PR #819, opened 2020-05-04
Fix uninitialized variables count and offset.

mykhailo-kuchma:
Uninitialized variables count and offset could potentially
cause broken response to user.

* Fix Android proxy enum.
* Fix network initialization in concurrent env.

Resolves: OLPSUP-10333

### DIFF
--- a/olp-cpp-sdk-core/src/http/android/HttpClient.java
+++ b/olp-cpp-sdk-core/src/http/android/HttpClient.java
@@ -113,15 +113,15 @@ public class HttpClient {
       this.proxyPort = proxyPort;
 
       switch (proxyType) {
-        case 0:
+        case 1:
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/src/http/android/HttpClient.java
```diff
@@ -327,7 +327,7 @@ protected Void doInBackground(Request... requests) {
             if (request.postData() != null) {
               conn.setDoOutput(true);
               conn.getOutputStream().write(request.postData());
-              Log.d(LOGTAG, "Uploaded data length:" + request.postData().length);
+              //Log.d(LOGTAG, "Uploaded data length:" + request.postData().length);
```
LiubovDidkivska, line 330: remove or uncomment

PR #839, opened 2020-05-15
Default expiration in PartitionsCacheRepository

kzvieriev:
Default cache items expiration time can be controlled through
OlpClientOptions, so PartitionsCacheRepository should use it instead of
maximum time_t value. Added unit tests.

Resolves: OLPEDGE-1916

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
@@ -40,6 +40,8 @@ PORTING_CLANG_GCC_DISABLE_WARNING("-Wdeprecated-declarations")
 
 namespace {
 constexpr auto kLogTag = "PartitionsCacheRepository";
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
```diff
@@ -55,6 +55,14 @@ std::string CreateKey(const std::string& hrn, const std::string& layer_id,
 std::string CreateKey(const std::string& hrn, const int64_t catalogVersion) {
   return hrn + "::" + std::to_string(catalogVersion) + "::layerVersions";
 }
+
+time_t ConvertTime(std::chrono::seconds time) {
+  if (time == std::chrono::seconds::max()) {
+    return std::numeric_limits<time_t>::max();
+  }
+
+  return time.count();
```
andescu, line 64:
```suggestion
  return (time == std::chrono::seconds::max()) ? std::numeric_limits<time_t>::max() 
                                                                               : time.count();
```

Consider adding some constexpr defines for readability.

PR #844, opened 2020-05-16
Fix compile errors

here-mikelley:
Add missing #include <algorithm> for std::any_of() and std::count_if()

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -34,6 +34,8 @@
 #include "repositories/PartitionsRepository.h"
 #include "repositories/PrefetchTilesRepository.h"
 
+#include <algorithm>
+
 // Needed to avoid endless warnings from GetVersion/WithVersion
 #include <olp/core/porting/warning_disable.h>
 PORTING_PUSH_WARNINGS()
--- a/olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
@@ -37,6 +37,8 @@
 #include "repositories/PartitionsRepository.h"
 #include "repositories/PrefetchTilesRepository.h"
 
+#include <algorithm>
+
 namespace olp {
 namespace dataservice {
 namespace read {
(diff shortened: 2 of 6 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -34,6 +34,8 @@
 #include "repositories/PartitionsRepository.h"
 #include "repositories/PrefetchTilesRepository.h"
 
+#include <algorithm>
```
andescu, line 37: Please move up between own header and module headers.
See coding style guide -> https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

olp-cpp-sdk-dataservice-read/src/VolatileLayerClientImpl.cpp
```diff
@@ -37,6 +37,8 @@
 #include "repositories/PartitionsRepository.h"
 #include "repositories/PrefetchTilesRepository.h"
 
+#include <algorithm>
```
andescu, line 40: Same here, see https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

andescu (commented):
Thanks for contributing to our repo!

You need to add DCO to your commit message and the commit message needs to have the following structure:

```
Summarize changes in around 50 characters or less.
   
Mandatory detailed explanatory text. Wrap it to about 72
characters or so. In some contexts, the first line is treated as the
subject of the commit and the rest of the text as the body. The
blank line separating the summary from the body is critical (unless
you omit the body entirely); various tools like `log`, `shortlog`
and `rebase` can get confused if you run the two together.
   
Explain the problem that this commit is solving. Focus on why you
are making this change as opposed to how (the code explains that).
Are there side effects or other unintuitive consequences of this
change? Here's the place to explain them.
   
Optional further paragraphs come after blank lines.
   
- Bullet points are okay, too.
   
- Typically a hyphen or asterisk is used for the bullet, preceded
  by a single space, with blank lines in between, but conventions
  vary here.
   
All commits need to reference a ticket, again separated by a blank
line from the commit message summary above:
 
Resolves: ABC-111, #123
Relates-To: ABC-333, #321
See also: ABC-432, #456, #789
 
Signed-off-by: FirstName LastName <firstname.lastname@here.com>
```

PR #848, opened 2020-05-18
Reduce excesive logging in API repositories.

andescu:
The SDK logs too much and too often especially in the API repositories.
This commit removes some of the duplicated logs, aligns all logs to a
common structure, and adds additional info like HRN and key to all
important logs. All other logs that are not important are now on level
debug. All previous trace logs which are actually printing debug level
information are now raised to debug level.

One very important fix that needs highlight is the capture by copy
which was removed from most of the cache repositories while putting
into the cache. It is not necessary to capture model data by copy as
the KeyValueCache::Put() is pure sync and will not leave the context
so that the model faces deletion. This additional copy is not needed
and might even bring performance decreases. This is now fixed and all
models are captured by reference when added to the cache.

Additionally, this commit also enables ccplint namespace check.

Resolves: OLPSUP-10501

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
@@ -59,39 +59,45 @@ ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApi(
     std::string service_version, FetchOptions options,
     client::OlpClientSettings settings) {
   repository::ApiCacheRepository repository(catalog, settings.cache);
+  const auto hrn = catalog.ToCatalogHRNString();
 
   if (options != OnlineOnly && options != CacheWithUpdate) {
     auto url = repository.Get(service, service_version);
     if (url) {
-      OLP_SDK_LOG_INFO_F(kLogTag, "LookupApi(%s, %s) -> from cache",
-                         service.c_str(), service_version.c_str());
+      OLP_SDK_LOG_DEBUG_F(kLogTag, "LookupApi(%s/%s) found in cache, hrn='%s'",
+                          service.c_str(), service_version.c_str(),
+                          hrn.c_str());
       client::OlpClient client;
       client.SetSettings(std::move(settings));
       client.SetBaseUrl(*url);
       return client;
     } else if (options == CacheOnly) {
-      return client::ApiError(
-          client::ErrorCode::NotFound,
-          "Cache only resource not found in cache (loopup api).");
+      return {{client::ErrorCode::NotFound,
+               "CacheOnly: resource not found in cache"}};
     }
   }
+
+  OLP_SDK_LOG_INFO_F(kLogTag,
+                     "LookupApi(%s/%s) cache miss, requesting, hrn='%s'",
+                     service.c_str(), service_version.c_str(), hrn.c_str());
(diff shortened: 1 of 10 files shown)


olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
```diff
@@ -58,40 +58,46 @@ ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApi(
...
       return client;
     } else if (options == CacheOnly) {
-      return client::ApiError(
-          client::ErrorCode::NotFound,
-          "Cache only resource not found in cache (loopup api).");
+      return {{ErrorCode::NotFound, "CacheOnly: resource not found in cache"}};
     }
   }
 
+  OLP_SDK_LOG_INFO_F(kLogTag,
```
mykhailo-kuchma, line 80: Trace/Debug?

olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
```diff
@@ -106,24 +112,24 @@ ApiClientLookup::ApiClientResponse ApiClientLookup::LookupApi(
...
-                   [&](const olp::dataservice::read::model::Api& obj) -> bool {
+                   [&](const dataservice::read::model::Api& obj) -> bool {
                      return (obj.GetApi().compare(service) == 0 &&
                              obj.GetVersion().compare(service_version) == 0);
                    });
   if (it == api_result.end()) {
-    OLP_SDK_LOG_INFO_F(kLogTag, "LookupApi(%s/%s): %s - service not available",
-                       service.c_str(), service_version.c_str(),
-                       catalog.partition.c_str());
+    OLP_SDK_LOG_INFO_F(kLogTag, "LookupApi(%s/%s) service not found, hrn='%s'",
```
mykhailo-kuchma, line 120: Error?

olp-cpp-sdk-dataservice-read/src/repositories/CatalogRepository.cpp
```diff
@@ -131,9 +130,15 @@ CatalogVersionResponse CatalogRepository::GetLatestVersion(
   if (version_response.IsSuccessful() && fetch_option != OnlineOnly) {
     repository.PutVersion(version_response.GetResult());
   }
+
   if (!version_response.IsSuccessful()) {
     const auto& error = version_response.GetError();
     if (error.GetHttpStatusCode() == http::HttpStatusCode::FORBIDDEN) {
+      OLP_SDK_LOG_INFO_F(kLogTag,
```
mykhailo-kuchma, line 137: Warning/Error?

olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
andescu: I think info is good so that we see how often we request API lookup. Ideally, it is only once and then stored to cache.

olp-cpp-sdk-dataservice-read/src/ApiClientLookup.cpp
andescu: Ill make it warning.

PR #850, opened 2020-05-19
Add CMake option OLP_SDK_DISABLE_DEBUG_LOGGING

mykhailo-kuchma:
Users can use it to disable debug/trace log messages. Default is OFF
which mean means that messages are enabled.

Resolves: OLPEDGE-247

### DIFF
--- a/olp-cpp-sdk-core/CMakeLists.txt
+++ b/olp-cpp-sdk-core/CMakeLists.txt
@@ -328,6 +328,12 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/include)
 
 add_library(${PROJECT_NAME} ${OLP_SDK_CORE_SOURCES} ${OLP_SDK_CORE_HEADERS})
 
+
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/CMakeLists.txt
```diff
@@ -328,6 +328,12 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/include)
 
 add_library(${PROJECT_NAME} ${OLP_SDK_CORE_SOURCES} ${OLP_SDK_CORE_HEADERS})
 
+
+if (OLP_SDK_DISABLE_DEBUG_LOGGING)
+    target_compile_definitions(${PROJECT_NAME} 
+        PUBLIC LOGGING_DISABLE_DEBUG_LEVEL)
```
andescu, line 334: Why do we need it public?

mykhailo-kuchma: Other modules must pick the correct macros

PR #859, opened 2020-05-22
Add compression to CacheSettings.

andescu:
This commit adds the possible compression settings to the CacheSettings
structure to be passed on to leveldb. Possible values so far are only
no compression or Snappy compression. Nevertheless there are user who
will profit from not applying any compression if the data added to the
cache is already compressed, e.g. protobuf.

Additionally this commit adds filter policy to bloom filter with 10
bits precision which will enhance reading speeds.

Relates-To: OLPEDGE-1984, OLPEDGE-1983

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
+++ b/olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
@@ -46,6 +46,15 @@ enum class EvictionPolicy : unsigned char {
 };
 
 /**
+ * @brief Options for database compression.
+ */
+enum class CompressionType {
+  kNoCompression,     /*!< No compression applied on the data before storing. */
+  kDefaultCompression /*!< Default compression will be applied to data before
+                        storing. */
+};
+
+/**
  * @brief Settings for memory and disk caching.
  */
 struct CORE_API CacheSettings {
@@ -112,6 +121,20 @@ struct CORE_API CacheSettings {
    * is EvictionPolicy::kLeastRecentlyUsed.
    */
   EvictionPolicy eviction_policy = EvictionPolicy::kLeastRecentlyUsed;
+
+  /**
+   * @brief This flag sets the compression policy to be applied on the database.
+   *
+   * In some cases though, when all the data to be inserted is already
+   * compressed by any means, e.g. protobuf or other serialization protocols, it
+   * might not be worth to enable any compression at all as it will eat up some
+   * CPU to compress and decompress the metadata without major gain. This
+   * parameter is dynamic and can be changed between runs. If changed only new
+   * values which are added will be using the new compression policy all
+   * existing entries will remain unchanged. The default value is
+   * CompressionType::kDefaultCompression.
+   */
+  CompressionType compression = CompressionType::kDefaultCompression;
 
   /**
    * @brief The path to the protected (read-only) cache.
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
```diff
@@ -113,6 +122,18 @@ struct CORE_API CacheSettings {
...
+   * @brief This flag sets the compression policy to be applied on the database.
+   *
+   * Even if the input data is incompressible, the kSnappyCompression
+   * implementation will efficiently detect that and will switch to uncompressed
+   * mode. In some cases though, when the data to be inserted is already
+   * compressed by any means, e.g. protobuf or other serialization protocols, it
+   * might not be worth it to enable any compression at all as it will eat up
+   * some CPU. The default value is CompressionType::kSnappyCompression.
+   */
+  CompressionType compression = CompressionType::kSnappyCompression;
```
r0busta, line 135: That makes the cache abstraction leaking. And might fire back with not the best-looking API.

olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
```diff
@@ -113,6 +122,18 @@ struct CORE_API CacheSettings {
...
 
+  /**
+   * @brief This flag sets the compression policy to be applied on the database.
+   *
+   * Even if the input data is incompressible, the kSnappyCompression
+   * implementation will efficiently detect that and will switch to uncompressed
+   * mode. In some cases though, when the data to be inserted is already
+   * compressed by any means, e.g. protobuf or other serialization protocols, it
+   * might not be worth it to enable any compression at all as it will eat up
+   * some CPU. The default value is CompressionType::kSnappyCompression.
```
mykhailo-kuchma, line 133: Should we add information that the parameter is dynamic and can be changed between runs?

andescu: Will add.

olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
andescu: Any suggestions what would be better?

r0busta: I advise against this option. 
First off, this "Even if the input data is incompressible, the kSnappyCompression implementation will efficiently detect that and will switch to uncompressed mode. [...]  it might not be worth it to enable any compression at all as it will eat up some CPU." contradicts itself. We either trust in Snappy's ability to detect incompressible data, or we don't and invest time in proper API design.
Second, there are some benchmarks[1] indicating that without compression, LevelDB works even slower.
Third, the original discussion that by disabling snappy we gain some perf was unfortunately void wasn't evidence-based.

--
[1] http://www.lmdb.tech/bench/microbench/benchmark.html

andescu: This benchmark http://www.lmdb.tech/bench/microbench/benchmark.html uses compressible data to measure if it makes sense or not, but in most of our cases, the data you download from the platform is already compressed, e.g. protobuf. And based on my benchmarking we definitely gain a lot of performance increase by disabling compression as the compression threads are still working in the background consuming CPU.

andescu: As discussed, changed to something more generic.

PR #863, opened 2020-05-25
Remove expired elements during cache eviction

mykhailo-kuchma:
When eviction happens, first we remove all expired elements, and only
then we remove other elements if still needed.
Change the internal LRU representation, so that we have fast access to
key expiration. Saves extra cache lookup during eviction.
Fix the use case when protected cache contains expired data.

Relates-To: OLPEDGE-1925

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -151,6 +151,13 @@ bool DefaultCacheImpl::Clear() {
   return SetupStorage() == DefaultCache::StorageOpenResult::Success;
 }
 
+void DefaultCacheImpl::Compact() {
+  std::lock_guard<std::mutex> lock(cache_lock_);
+  if (mutable_cache_) {
+    mutable_cache_->Compact();
+  }
+}
+
 bool DefaultCacheImpl::Put(const std::string& key, const boost::any& value,
                            const Encoder& encoder, time_t expiry) {
   std::lock_guard<std::mutex> lock(cache_lock_);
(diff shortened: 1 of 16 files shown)


olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
```diff
@@ -56,9 +56,15 @@ class DefaultCacheImpl {
   bool RemoveKeysWithPrefix(const std::string& key);
 
  protected:
-  /// Alias for the LRU cache using the leveldb keys as key and the value size
+  /// The LRU value property.
+  struct ValueProperties {
+    size_t size{0ull};
+    time_t expiry{olp::cache::KeyValueCache::kDefaultExpiry};
```
andescu, line 62:
```suggestion
    time_t expiry{cache::KeyValueCache::kDefaultExpiry};
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -379,24 +400,62 @@ uint64_t DefaultCacheImpl::MaybeEvictData(leveldb::WriteBatch& batch) {
   int64_t evicted = 0u;
   auto count = 0u;
   const auto min_size = kMinDiskUsedThreshold * settings_.max_disk_storage;
+
+  const auto current_time = olp::cache::InMemoryCache::DefaultTimeProvider()();
+
+  // Remove the expired elements first
+  for (auto it = mutable_cache_lru_->begin();
```
andescu, line 407: Maybe start from rbegin(), same as below and search for expired. After a certain point i guess the elements that are at the end are more likely to be expired.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -378,24 +399,62 @@ uint64_t DefaultCacheImpl::MaybeEvictData(leveldb::WriteBatch& batch) {
...
+      batch.Delete(expiry_key);
+      evicted += expiry_key.size() + kExpiryValueSize;
+
+      ++count;
+
+      if (memory_cache_) {
+        memory_cache_->Remove(key);
+      }
+
+      it = mutable_cache_lru_->Erase(it);
```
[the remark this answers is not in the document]
mykhailo-kuchma: It does not work when you do the reverse pass

mykhailo-kuchma:
Created a new PR https://github.com/heremaps/here-data-sdk-cpp/pull/870

PR #864, opened 2020-05-25
Remove unnecessary allocations in HRN lookup

haraldF:
Remove dozens of needless mallocs and also remove the unneeded static
global object, removing the auto-generated memory barrier.

Since all the data we have is static, use a POD. Also, with only 4
items, a linear search is superior.

Signed-off-by: Harald Fernengel <harald.fernengel@here.com>


haraldF:
Abandoning. It's a micro-optimization that's not worth pursuing.

PR #879, opened 2020-06-02
Replace AuthenticationError class with ApiError

mykhailo-kuchma:
Class AuthenticationError duplicates the ApiError class.
Class is removed, but the alias to ApiError is still present and
deprecated.

Relates-To: OLPEDGE-1027


olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationError.h
```diff
@@ -21,8 +21,9 @@
 
 #include <string>
 
+#include <olp/core/client/ApiError.h>
+
```
andescu, line 25: Remove empty line.

PR #883, opened 2020-06-03
GetAggregatedPartition for PartitionRepository.

kzvieriev:
This is a part of GetAggregatedData API update. PartitionRepository
handle loading quad tree and storing it into the cache, so it can be
used by clients further.

Resolves: OLPEDGE-1932

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.cpp
@@ -134,6 +134,34 @@ QueryApi::QuadTreeIndexResponse QueryApi::QuadTreeIndex(
   return olp::parser::parse<model::Index>(response.response);
 }
 
+olp::client::HttpResponse QueryApi::QuadTreeIndex(
+    const client::OlpClient& client, const std::string& layer_id,
+    const std::string& quad_key, boost::optional<int64_t> version,
+    int32_t depth, boost::optional<std::vector<std::string>> additional_fields,
+    boost::optional<std::string> billing_tag,
+    client::CancellationContext context) {
+  std::multimap<std::string, std::string> header_params;
+  header_params.emplace("Accept", "application/json");
+
+  std::multimap<std::string, std::string> query_params;
+  if (additional_fields) {
+    query_params.emplace("additionalFields",
+                         ConcatStringArray(*additional_fields, ","));
+  }
+  if (billing_tag) {
+    query_params.emplace("billingTag", *billing_tag);
+  }
+
+  std::string metadata_uri =
+      "/layers/" + layer_id +
+      (version ? "/versions/" + std::to_string(version.get()) : "") +
+      "/quadkeys/" + quad_key + "/depths/" + std::to_string(depth);
+
+  return client.CallApi(metadata_uri, "GET", std::move(query_params),
+                        std::move(header_params), {}, nullptr, std::string{},
+                        std::move(context));
+}
+
 QueryApi::QuadTreeIndexResponse QueryApi::QuadTreeIndexVolatile(
     const client::OlpClient& client, const std::string& layer_id,
     const std::string& quad_key, int32_t depth,
--- a/olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.h
+++ b/olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.h
@@ -25,6 +25,7 @@
 #include <olp/core/client/ApiError.h>
 #include <olp/core/client/ApiResponse.h>
 #include <olp/core/client/CancellationContext.h>
+#include <olp/core/client/HttpResponse.h>
 #include <boost/optional.hpp>
 #include "generated/model/Index.h"
 #include "olp/dataservice/read/model/Partitions.h"
@@ -109,6 +110,14 @@ class QueryApi {
   static QuadTreeIndexResponse QuadTreeIndex(
       const client::OlpClient& client, const std::string& layer_id,
       int64_t version, const std::string& quad_key, int32_t depth,
+      boost::optional<std::vector<std::string>> additional_fields,
+      boost::optional<std::string> billing_tag,
+      client::CancellationContext context);
+
+  static olp::client::HttpResponse QuadTreeIndex(
+      const client::OlpClient& client, const std::string& layer_id,
+      const std::string& quad_key, boost::optional<int64_t> version,
+      int32_t depth,
       boost::optional<std::vector<std::string>> additional_fields,
       boost::optional<std::string> billing_tag,
       client::CancellationContext context);
--- a/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
@@ -40,8 +40,13 @@ PORTING_PUSH_WARNINGS()
 PORTING_CLANG_GCC_DISABLE_WARNING("-Wdeprecated-declarations")
 
 namespace {
+namespace read = olp::dataservice::read;
+namespace model = olp::dataservice::read::model;
+namespace repository = olp::dataservice::read::repository;
+
 constexpr auto kLogTag = "PartitionsRepository";
 constexpr uint32_t kMaxQuadTreeIndexDepth = 4u;
+constexpr auto kAggregateQuadTreeDepth = 4;
 
 using LayerVersionReponse =
     olp::client::ApiResponse<int64_t, olp::client::ApiError>;
@@ -63,6 +68,61 @@ TtlForLayer(const std::vector<olp::dataservice::read::model::Layer>& layers,
   return olp::client::ApiError(olp::client::ErrorCode::NotFound,
                                "Layer specified doesn't exist");
 }
+
+repository::PartitionResponse FindPartition(
+    const read::QuadTreeIndex& quad_tree, const read::TileRequest& request,
+    bool aggregated) {
+  const auto& tile_key = request.GetTileKey();
+
+  // Look for requested tile in the quad tree or its closest ancestor
+  const auto found_index_data = quad_tree.Find(tile_key, aggregated);
+
+  if (!found_index_data) {
+    OLP_SDK_LOG_WARNING_F(
+        kLogTag,
+        "FindPartition: tile not found, tile='%s', depth='%" PRId32
+        "', aggregated='%s'",
+        tile_key.ToHereTile().c_str(), kAggregateQuadTreeDepth,
+        aggregated ? "true" : "false");
+    return {{olp::client::ErrorCode::NotFound,
+             "Tile or its closest ancestors not found"}};
+  }
+
+  const auto index_data = found_index_data.get();
+  const auto found_tile_key = index_data.tile_key;
+  const auto data_handle = index_data.data_handle;
+
+  auto aggregated_partition = model::Partition();
+  aggregated_partition.SetDataHandle(data_handle);
+  aggregated_partition.SetPartition(found_tile_key.ToHereTile());
+
+  return std::move(aggregated_partition);
+}
+
+read::QuadTreeIndex FindQuadTree(
+    repository::PartitionsCacheRepository& repository, const std::string& layer,
+    boost::optional<int64_t> version, const olp::geo::TileKey& tile_key) {
+  auto max_depth =
+      std::min<std::uint32_t>(tile_key.Level(), kMaxQuadTreeIndexDepth);
+  for (int i = max_depth; i >= 0; --i) {
+    const auto& root_tile_key = tile_key.ChangedLevelBy(-i);
+    auto cached_tree =
+        repository.Get(layer, root_tile_key, kAggregateQuadTreeDepth, version);
+    if (!cached_tree.IsNull()) {
+      OLP_SDK_LOG_DEBUG_F(kLogTag,
+                          "FindQuadTreeInCache found in cache, tile='%s', "
+                          "root='%s', depth='%" PRId32 "'",
+                          tile_key.ToHereTile().c_str(),
+                          root_tile_key.ToHereTile().c_str(),
+                          kAggregateQuadTreeDepth);
+
+      return cached_tree;
+    }
+  }
+
+  return read::QuadTreeIndex();
+}
+
 }  // namespace
 
 namespace olp {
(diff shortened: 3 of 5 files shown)


olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.h
```diff
@@ -113,6 +114,13 @@ class QueryApi {
       boost::optional<std::string> billing_tag,
       client::CancellationContext context);
 
+  static olp::client::HttpResponse CreateQuadTreeIndex(
+      const client::OlpClient& client, const std::string& layer_id,
+      int64_t version, const std::string& quad_key, int32_t depth,
+      boost::optional<std::vector<std::string>> additional_fields,
+      boost::optional<std::string> billing_tag,
+      client::CancellationContext context);
+
```
andescu, line 123: `CreateQuadTreeIndex` is a very unsuitable name. Are you creating the quadtree index yourself or are you requesting it from the platform?
Name it QuadTreeIndex() with boost::optional<int64_t> as version and then we can remove both others.
Return type HttpResponse is good.

olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.cpp
```diff
@@ -134,6 +134,33 @@ QueryApi::QuadTreeIndexResponse QueryApi::QuadTreeIndex(
...
 }
 
+olp::client::HttpResponse QueryApi::CreateQuadTreeIndex(
+    const client::OlpClient& client, const std::string& layer_id,
+    int64_t version, const std::string& quad_key, int32_t depth,
+    boost::optional<std::vector<std::string>> additional_fields,
+    boost::optional<std::string> billing_tag,
+    client::CancellationContext context) {
+  std::multimap<std::string, std::string> header_params;
+  header_params.insert(std::make_pair("Accept", "application/json"));
```
andescu, line 144:
```suggestion
  header_params.emplace("Accept", "application/json");
```
Same for below inserts also.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
```diff
@@ -60,6 +61,12 @@ class PartitionsRepository {
   static model::Partition PartitionFromSubQuad(
       const model::SubQuad& sub_quad, const std::string& partition);
 
+  static PartitionsResponse GetAggregatedPartitionForVersionedTile(
+      client::HRN catalog, std::string layer,
```
andescu, line 65:
```suggestion
      const client::HRN& catalog, const std::string& layer,
```
Any async operations? Then make them pass by reference.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
```diff
@@ -60,6 +61,12 @@ class PartitionsRepository {
   static model::Partition PartitionFromSubQuad(
       const model::SubQuad& sub_quad, const std::string& partition);
 
+  static PartitionsResponse GetAggregatedPartitionForVersionedTile(
```
andescu, line 64: Why do we need only aggregated partition? We can use the same for Prefetch and normal GetData(TileRequest).
Please make a more generic solution that we can reuse for other functions that do not need the nearest parent but the tile itself.

Also `GetAggregatedPartitionForVersionedTile` is not the best name. Change to `GetAggregatedTile` and make the version an optional so that we can also use it for volatile layers.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +320,77 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
   return ret;
 }
 
+PartitionsResponse PartitionsRepository::GetAggregatedPartitionForVersionedTile(
+    client::HRN catalog, std::string layer, client::CancellationContext context,
+    const TileRequest& request, int64_t version,
+    const client::OlpClientSettings& settings) {
+  const auto fetch_option = request.GetFetchOption();
+  const auto& tile_key = request.GetTileKey();
+  const auto& root_tile_key = tile_key.ChangedLevelBy(-kAggregateQuadTreeDepth);
```
andescu, line 329: This is not correct. If you are doing it like this we will have endless duplicates in the cache. E.g. user requests tile at level 14, you download quadtree at level 10, which includes levels {14, 13, 12, 11, 10} and parents up to level 1.
Now the user requests a level 13 tile. So you go on level 9 and request the quadtree index, which includes levels {13, 12, 11, 10, 9}. But you already have these partitions in the level 10 quadtree index. Why do you download it again?

kzvieriev: But strategy was to use quad trees with depth 4, so they are different for tiles from different levels.
Quad tree will be the same for 256 level 14 tiles.
I suppose this is actual use case for GetAggregatedData() - request neighbor tiles from same level. Am I wrong?

andescu: Yes, and it still is to request with depth 4, but not if we already have it in cache within one of the quadtrees that were downloaded before.

olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.h
kzvieriev: Done, but others QuadTreeIndex() are still there. Replacing them requires many corrections in *Repository classes and tests. Probably, need separate ticket for that.

olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.h
```diff
@@ -113,6 +114,14 @@ class QueryApi {
       boost::optional<std::string> billing_tag,
       client::CancellationContext context);
 
+  static olp::client::HttpResponse QuadTreeIndex(
+      const client::OlpClient& client, const std::string& layer_id,
+      const std::string& quad_key, int32_t depth,
+      boost::optional<std::vector<std::string>> additional_fields,
+      boost::optional<std::string> billing_tag,
+      client::CancellationContext context,
+      boost::optional<int64_t> version = boost::none);
```
andescu, line 123: CancellationContext should be last. No default value needed for version.

olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.cpp
```diff
@@ -134,6 +134,34 @@ QueryApi::QuadTreeIndexResponse QueryApi::QuadTreeIndex(
...
 }
 
+olp::client::HttpResponse QueryApi::QuadTreeIndex(
+    const client::OlpClient& client, const std::string& layer_id,
+    const std::string& quad_key, int32_t depth,
+    boost::optional<std::vector<std::string>> additional_fields,
+    boost::optional<std::string> billing_tag,
+    client::CancellationContext context, boost::optional<int64_t> version) {
+  std::multimap<std::string, std::string> header_params;
+  header_params.emplace(std::make_pair("Accept", "application/json"));
```
andescu, line 144: You don't need std::make_pair. `emplace` will construct a pair in-place.

olp-cpp-sdk-dataservice-read/src/generated/api/QueryApi.h
andescu: Yes, we should do this in a separate ticket.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
```diff
@@ -60,6 +61,12 @@ class PartitionsRepository {
   static model::Partition PartitionFromSubQuad(
       const model::SubQuad& sub_quad, const std::string& partition);
 
+  static PartitionsResponse GetAggregatedTile(
+      const client::HRN& catalog, const std::string& layer,
+      client::CancellationContext cancellation_context,
+      const TileRequest& request, boost::optional<int64_t> version,
+      const client::OlpClientSettings& settings, bool aggregated);
```
andescu, line 68: I don't think that the GetAggregatedTile should give you back something else than an aggregated tile. Having an additional parameter `bool aggregated` doesn't make much sense.
A more generic solution you found, almost, with the FindPartition() which can be used by both GetAggregatedTile() and GetPartitionForVersionedTile() or others who use or should use the QuadTreeIndex.
What is missing is a common solution for finding the correct quadtree index in the cache, which currently is in GetAggregatedTile() only. But let's postpone that to another time.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -63,6 +67,40 @@ TtlForLayer(const std::vector<olp::dataservice::read::model::Layer>& layers,
...
+                                       const olp::geo::TileKey& root_tile_key,
+                                       const read::TileRequest& request,
+                                       bool aggregated) {
+  const auto& tile_key = request.GetTileKey();
+
+  // Look for requested tile in the quad tree or its closest ancestor
+  const auto found_index_data = quad_tree.Find(tile_key, aggregated);
+
+  if (!found_index_data) {
+    OLP_SDK_LOG_ERROR_F(kLogTag,
```
andescu, line 81: Rather warning.
Usually, logs should look like this: `FindPartition: tile not found, tile=%s, root=%s, depth=%d, aggregated='%s'`
Doing all these grammatically correct sentences will bring no gain when it crashes :)

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
```diff
@@ -60,6 +61,12 @@ class PartitionsRepository {
   static model::Partition PartitionFromSubQuad(
       const model::SubQuad& sub_quad, const std::string& partition);
 
+  static PartitionsResponse GetAggregatedTile(
```
andescu, line 64: Why do we need a vector as output? One partition is not enough?

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+  const auto& tile_key = request.GetTileKey();
+
+  // Download quad tree index depth 4 or load from cache
+  if (fetch_option != OnlineOnly && fetch_option != CacheWithUpdate) {
+    repository::PartitionsCacheRepository repository(
+        catalog, settings.cache, settings.default_cache_expiration);
+
+    // handle tiles from level < kAggregateQuadTreeDepth
+    auto max_depth =
+        std::min<int>(int(tile_key.Level()), kAggregateQuadTreeDepth);
```
andescu, line 344: Use uint32_t or int32_t and no C-Style casting.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+    for (auto i = max_depth; i >= 0; --i) {
+      const auto& root_tile_key = tile_key.ChangedLevelBy(-i);
+      cached_tree = repository.Get(layer, root_tile_key,
+                                   kAggregateQuadTreeDepth, version);
+      if (!cached_tree.IsNull()) {
+        OLP_SDK_LOG_DEBUG_F(kLogTag, "cache QuadTree (%s, %" PRId32 ") found!",
+                            root_tile_key.ToHereTile().c_str(),
+                            kAggregateQuadTreeDepth);
+
+        return FindPartition(cached_tree, root_tile_key, request, aggregated);
```
andescu, line 356: I'd prefer you move this to a separate function.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+    auto max_depth =
+        std::min<int>(int(tile_key.Level()), kAggregateQuadTreeDepth);
+    QuadTreeIndex cached_tree;
+    // search in the cache for any quad tree covering the tile
+    for (auto i = max_depth; i >= 0; --i) {
+      const auto& root_tile_key = tile_key.ChangedLevelBy(-i);
+      cached_tree = repository.Get(layer, root_tile_key,
+                                   kAggregateQuadTreeDepth, version);
+      if (!cached_tree.IsNull()) {
+        OLP_SDK_LOG_DEBUG_F(kLogTag, "cache QuadTree (%s, %" PRId32 ") found!",
```
andescu, line 352: I would prefer you follow same approach as below methods:

`GetPartitionForVersionedTile found in cache, hrn='%s', key='%s'`

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+                            root_tile_key.ToHereTile().c_str(),
+                            kAggregateQuadTreeDepth);
+
+        return FindPartition(cached_tree, root_tile_key, request, aggregated);
+      }
+    }
+
+    if (cached_tree.IsNull() && fetch_option == CacheOnly) {
+      OLP_SDK_LOG_INFO_F(kLogTag, "cache QuadTree for Tile(%s) not found!",
```
andescu, line 361: Same here.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+                         tile_key.ToHereTile().c_str());
+      return client::ApiError(
+          client::ErrorCode::NotFound,
+          "Cache only resource not found in cache (QuadTree).");
```
andescu, line 365:
```suggestion
      return {{client::ErrorCode::NotFound,
             "CacheOnly: resource not found in cache"}};
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -63,6 +67,40 @@ TtlForLayer(const std::vector<olp::dataservice::read::model::Layer>& layers,
...
+
+  if (!found_index_data) {
+    OLP_SDK_LOG_ERROR_F(kLogTag,
+                        "Tile (%s) and its ancestors not present in "
+                        "QuadTreeIndex (%s, %" PRId32 ")",
+                        tile_key.ToHereTile().c_str(),
+                        root_tile_key.ToHereTile().c_str(),
+                        kAggregateQuadTreeDepth);
+    return {{olp::client::ErrorCode::NotFound,
+             "Tile and its ancestors not found."}};
```
andescu, line 88:
```suggestion
             "Tile or its closest ancestors not found"}};
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+          "Cache only resource not found in cache (QuadTree).");
+    }
+  }
+
+  // quad tree data not found in the cache
+  auto query_api = ApiClientLookup::LookupApi(
+      catalog, context, "query", "v1", request.GetFetchOption(), settings);
+
+  if (!query_api.IsSuccessful()) {
+    OLP_SDK_LOG_ERROR(kLogTag,
```
andescu, line 374: Warning.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+                      "QueryPartitionForVersionedTile: LookupApi failed.");
```
andescu, line 375: Print hrn and what API you are looking for.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+  }
+
+  const auto& root_tile_key = tile_key.ChangedLevelBy(-kAggregateQuadTreeDepth);
+  const auto root_tile_here = root_tile_key.ToHereTile();
+  auto quadtree_response = QueryApi::QuadTreeIndex(
+      query_api.GetResult(), layer, root_tile_here, version,
+      kAggregateQuadTreeDepth, boost::none, request.GetBillingTag(), context);
+
+  if (quadtree_response.status != olp::http::HttpStatusCode::OK) {
+    OLP_SDK_LOG_ERROR_F(kLogTag,
```
andescu, line 386: Warning, print hrn and layer.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+                        root_tile_here.c_str(), version.get_value_or(-1),
+                        kAggregateQuadTreeDepth);
+    return client::ApiError(quadtree_response.status,
+                            quadtree_response.response.str());
+  }
+
+  QuadTreeIndex tree(root_tile_key, kAggregateQuadTreeDepth,
+                     quadtree_response.response);
+  if (tree.IsNull()) {
+    OLP_SDK_LOG_ERROR_F(kLogTag,
```
andescu, line 397: Warning, print HRN and layer.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+                        "QuadTreeIndex failed (%s, %" PRId64 ", %" PRId32 ")",
+                        root_tile_here.c_str(), version.get_value_or(-1),
+                        kAggregateQuadTreeDepth);
+    return {
+        {client::ErrorCode::Unknown, "Failed to parse QuadTreeIndex json."}};
```
andescu, line 402:
```suggestion
        {client::ErrorCode::Unknown, "Failed to parse QuadTreeIndex json"}};
```
Formatting seems wrong!

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +326,91 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
...
+  }
+
+  if (fetch_option != OnlineOnly) {
+    repository::PartitionsCacheRepository repository(
```
andescu, line 406: You already have one PartitionsCacheRepository setup above, can you not use the same?

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
kzvieriev: I have chosen PartitionsResponse because it can contain desired partition and can provide error things.
But yes, vector is overkill here.

Maybe, it makes sense to return QuadTreeIndex here? So in VersionedLayerClient we can at least differ online/cache errors from tile not found in quad tree case.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
kzvieriev: Moved this search algorithm into separate function, so now we have a way to find in the cache QuadTreeIndex covering some tile :-)

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -63,6 +68,63 @@ TtlForLayer(const std::vector<olp::dataservice::read::model::Layer>& layers,
...
+read::PartitionsResponse FindPartition(const read::QuadTreeIndex& quad_tree,
+                                       const read::TileRequest& request,
+                                       bool aggregated) {
+  const auto& tile_key = request.GetTileKey();
+
+  // Look for requested tile in the quad tree or its closest ancestor
+  const auto found_index_data = quad_tree.Find(tile_key, aggregated);
+
+  if (!found_index_data) {
+    OLP_SDK_LOG_ERROR_F(
```
andescu, line 81: Error is a serious failure. I guess finding a tile in the quad tree does not qualify as a serious failure. A warning should be enough.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
andescu: Creating a PartitionResponse, means `Response<model::Partition>`, is probably the least problematic part.
If this is the only issue then please add an alias and lets only return one single model::Partition.
Returning QuadTreeIndex means that all the methods using this GetAggregatedTile() actually need to implement the logic themselves. Then there is no point anymore to call it GetAggregatedTile() but rather GetQuadTreeIndex().

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -288,6 +350,82 @@ model::Partition PartitionsRepository::PartitionFromSubQuad(
   return ret;
 }
 
+PartitionsResponse PartitionsRepository::GetAggregatedTile(
+    const client::HRN& catalog, const std::string& layer,
+    client::CancellationContext context, const TileRequest& request,
+    boost::optional<int64_t> version, const client::OlpClientSettings& settings,
+    bool aggregated) {
```
andescu, line 357: As i said we dont need the `bool aggregated` parameter. Just hard code to true in FindPartition().

PR #885, opened 2020-06-03
Auth fallback mechanism when system time is wrong

LiubovDidkivska:
For case if system time is wrong read time from
headers and send authorization request again with readed timestamp.
Add integration test.

Relates-To: OLPEDGE-1921

### DIFF
--- a/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
+++ b/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
@@ -137,10 +137,12 @@ std::time_t ParseTime(const std::string& value) {
 }
 
 std::time_t GetTimestampFromHeaders(const olp::http::Headers& headers) {
-  auto it = std::find_if(begin(headers), end(headers),
-                         [](const std::pair<std::string, std::string>& obg) {
-                           return (obg.first.compare(kDate) == 0);
-                         });
+  auto it =
+      std::find_if(begin(headers), end(headers),
+                   [](const std::pair<std::string, std::string>& obg) {
+                     return olp::http::NetworkUtils::CaseInsensitiveCompare(
+                         obg.first, kDate);
+                   });
   if (it != end(headers)) {
     return ParseTime(it->second);
   }
--- a/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
+++ b/olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
@@ -51,6 +51,7 @@
 #include "olp/core/http/NetworkRequest.h"
 #include "olp/core/http/NetworkResponse.h"
 #include "olp/core/http/NetworkUtils.h"
+#include "olp/core/logging/Log.h"
 #include "olp/core/porting/make_unique.h"
 #include "olp/core/thread/Atomic.h"
 #include "olp/core/thread/TaskScheduler.h"
@@ -122,16 +123,28 @@ constexpr auto kVersion = "1.0";
 constexpr auto kHmac = "HMAC-SHA256";
 constexpr auto kDate = "date";
 constexpr auto kErrorWrongTimestamp = 401204;
+constexpr auto kLogTag = "AuthenticationClient";
+
+#ifdef _WIN32
+#define timegm _mkgmtime
+#endif
 
 std::time_t ParseTime(const std::string& value) {
   std::tm tm = {};
   std::istringstream ss(value);
   ss >> std::get_time(&tm, "%a, %d %b %Y %H:%M:%S %z");
-#ifdef _WIN32
-  return _mkgmtime(&tm);
-#else
   return timegm(&tm);
-#endif
+}
+
+std::time_t GetTimestampFromHeaders(const olp::http::Headers& headers) {
+  auto it = std::find_if(begin(headers), end(headers),
+                         [](const std::pair<std::string, std::string>& obg) {
+                           return (obg.first.compare(kDate) == 0);
+                         });
+  if (it != end(headers)) {
+    return ParseTime(it->second);
+  }
+  return 0;
 }
 
 void ExecuteOrSchedule(
@@ -454,6 +467,14 @@ class AuthenticationClient::Impl final {
       const std::string& reacceptance_token);
   client::OlpClient::RequestBodyType GenerateAuthorizeBody(
       AuthorizeRequest properties);
+  olp::client::HttpResponse CallAuth(
+      const client::OlpClient& client, client::CancellationContext context,
+      const AuthenticationCredentials& credentials,
+      const SignInProperties& properties, std::time_t timestamp);
+  SignInResult ParseAuthResponse(int status, std::stringstream& auth_response);
+  AuthenticationClient::SignInClientResponse FindSingInResponceInCache(
+      int status, const std::string& error_message,
+      AuthenticationCredentials credentials);
 
   std::string GenerateUid();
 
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -477,66 +491,99 @@ client::CancellationToken AuthenticationClient::Impl::SignInClient(
...
+    std::time_t timestamp_from_header = 0;
+    ProcessReturnHeaders get_timestamp_from_headers = nullptr;
 
-      if (!cached_response) {
-        return {{status, error_message}};
+    if (!settings_.use_system_time) {
+      auto server_time = GetTimeFromServer(context, client);
+      if (server_time.IsSuccessful()) {
+        timestamp = server_time.GetResult();
       }
```
kzvieriev, line 561: Just curious, are we ok with failed  GetTimeFromServer()?
I think, we should have some log message and return the error.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -477,66 +493,101 @@ client::CancellationToken AuthenticationClient::Impl::SignInClient(
...
+      get_timestamp_from_headers = [&](const olp::http::Headers& headers) {
+        auto it =
+            std::find_if(begin(headers), end(headers),
+                         [](const std::pair<std::string, std::string>& obg) {
+                           return (obg.first.compare(kDate) == 0);
+                         });
+        if (it != end(headers)) {
+          timestamp_from_header = ParseTime(it->second);
+        }
+      };
```
mykhailo-kuchma, line 579: Please make it as an anonymous function, don't mix logic and algorithms.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -477,66 +493,101 @@ client::CancellationToken AuthenticationClient::Impl::SignInClient(
...
-          [&](utils::LruCache<std::string, SignInResult>& c) {
-            return c.InsertOrAssign(credentials.GetKey(), response);
-          });
+    auto response =
+        call_sign_in_request(timestamp, std::move(get_timestamp_from_headers));
+    if (settings_.use_system_time &&
+        response.GetResult().GetStatus() ==
+            http::HttpStatusCode::UNAUTHORIZED &&
+        response.GetResult().GetErrorResponse().code == kErrorWrongTimestamp &&
+        timestamp_from_header != 0) {
```
mykhailo-kuchma, line 588: No check if we can use Result

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -477,66 +493,101 @@ client::CancellationToken AuthenticationClient::Impl::SignInClient(
     }
 
     client::OlpClient client = CreateOlpClient(settings_, boost::none);
+    using ProcessReturnHeaders = std::function<void(const olp::http::Headers&)>;
 
-    std::time_t timestamp =
-        std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+    auto call_sign_in_request = [&](std::time_t& in_timestamp,
+                                    ProcessReturnHeaders get_timestamp =
+                                        nullptr) -> SignInClientResponse {
```
mykhailo-kuchma, line 500: Looks like it's easier to return client::HttpResponse and do the work outside (parsing and time from headers).
Then you don't need to provide any kind of callbacks.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -477,66 +493,101 @@ client::CancellationToken AuthenticationClient::Impl::SignInClient(
...
-    auto auth_response = client.CallApi(
-        kOauthEndpoint, "POST", {}, std::move(headers), {},
-        GenerateClientBody(properties), kApplicationJson, context);
+        auto cached_response = client_token_cache_->locked(
+            [&](utils::LruCache<std::string, SignInResult>& c)
+                -> boost::optional<SignInResult> {
+              auto it = c.Find(credentials.GetKey());
+              return it != c.end() ? boost::make_optional(it->value())
+                                   : boost::none;
+            });
```
mykhailo-kuchma, line 529: It should be done at the end, probably outside of this lambda.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
mykhailo-kuchma: In the end it can be an anonymous function.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -119,6 +121,20 @@ constexpr auto kOperator = "operator";
...
+constexpr auto kDate = "date";
+constexpr auto kErrorWrongTimestamp = 401204;
+constexpr auto kLogTag = "AuthenticationClient";
+
+std::time_t ParseTime(const std::string& value) {
+  std::tm tm = {};
+  std::istringstream ss(value);
+  ss >> std::get_time(&tm, "%a, %d %b %Y %H:%M:%S %z");
+#ifdef _WIN32
+  return _mkgmtime(&tm);
```
andescu, line 133: Optionally would be to make above something like

```cpp
#ifdef _WIN32
#define timegm _mkgmtime
#endif
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -466,78 +501,127 @@ AuthenticationClient::Impl::Impl(AuthenticationSettings settings)
...
+  }
 
-    auto auth_response = client.CallApi(
-        kOauthEndpoint, "POST", {}, std::move(headers), {},
-        GenerateClientBody(properties), kApplicationJson, context);
+  return *cached_response;
+}
 
-    const auto status = auth_response.status;
+client::CancellationToken AuthenticationClient::Impl::SignInClient(
```
andescu, line 557: Maybe I am missing this but I think you handle cancellation a little wrong. At least you check cancellation status in CallAuth but you do not stop afterwards but keep on going and even trying to parse the response from CallAuth() which in this case is `{static_cast<int>(olp::http::ErrorCode::CANCELLED_ERROR), "Cancelled"}`.
I would recommend that you stop directly and return status to the user that it was cancelled. This means checking after each major step that the result is not cancelled.

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -440,6 +467,14 @@ class AuthenticationClient::Impl final {
       const std::string& reacceptance_token);
   client::OlpClient::RequestBodyType GenerateAuthorizeBody(
       AuthorizeRequest properties);
+  olp::client::HttpResponse CallAuth(
+      const client::OlpClient& client, client::CancellationContext context,
+      const AuthenticationCredentials& credentials,
+      const SignInProperties& properties, std::time_t timestamp);
+  SignInResult ParseAuthResponse(int status, std::stringstream& auth_response);
+  AuthenticationClient::SignInClientResponse FindSingInResponceInCache(
```
andescu, line 475:
```suggestion
  AuthenticationClient::SignInClientResponse FindSingInResponseInCache(
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -440,6 +467,14 @@ class AuthenticationClient::Impl final {
...
+      int status, const std::string& error_message,
+      AuthenticationCredentials credentials);
```
andescu, line 477:
```suggestion
      const AuthenticationCredentials& credentials);
```

tests/integration/olp-cpp-sdk-authentication/AuthenticationClientTest.cpp
```diff
@@ -317,6 +317,93 @@ TEST_F(AuthenticationClientTest, SignInClientUseLocalTime) {
...
+  auth::AuthenticationCredentials credentials(key_, secret_);
+  std::promise<auth::AuthenticationClient::SignInClientResponse> request;
+  auto request_future = request.get_future();
+
+  EXPECT_CALL(*network_, Send(common::IsGetRequest(kTimestampUrl), _, _, _, _))
+      .Times(0);
+
+  EXPECT_CALL(*network_,
+              Send(common::IsPostRequest(
+                       "https://authentication.server.url/oauth2/token"),
```
andescu, line 338: Please create a const for this.

tests/integration/olp-cpp-sdk-authentication/AuthenticationClientTest.cpp
```diff
@@ -317,6 +317,93 @@ TEST_F(AuthenticationClientTest, SignInClientUseLocalTime) {
...
+                     .WithRequestId(request_id)
+                     .WithStatus(olp::http::HttpStatusCode::OK));
+        if (data_callback) {
+          auto raw = const_cast<char*>(kResponseWithScope.c_str());
+          data_callback(reinterpret_cast<uint8_t*>(raw), 0,
+                        kResponseWithScope.size());
+        }
+
+        return olp::http::SendOutcome(request_id);
+      });
```
andescu, line 382: Please use `common::ReturnHttpResponse` for this as in:

```cpp
EXPECT_CALL(*network_, Send(_, _, _, _, _))
      .WillOnce(common::ReturnHttpResponse(
          GetResponse(olp::http::HttpStatusCode::NO_CONTENT)
              .WithError(kErrorNoContent),
          kResponseNoContent));
```

olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
LiubovDidkivska: I have added additional check for context cancellation. Cancelation after CallAuth should be handled in FindSingInResponseInCache.
```
          if (status < 0) {
            return FindSingInResponseInCache(
                status, auth_response.response.str(), credentials);
          } 
```

andescu: Ok, now I see. I didn't see that you return http:ErrorCode which is negative. Then ignore this comment.

PR #896, opened 2020-06-05
Add info on `default_cache_expiration`

HalynaDumych:
Include `default_cache_expiration` in the following instructions:

* Read Example
* Cache Example

Relates-To: OLPEDGE-2004

### DIFF
--- a/docs/dataservice-cache-example.md
+++ b/docs/dataservice-cache-example.md
@@ -111,9 +111,21 @@ You need to create the `OlpClientSettings` object to get catalog and partition m
 
 3. [Authenticate](#authenticate-to-here-olp-using-client-credentials) to the HERE platform.
 
-4. Set up the `OlpClientSettings` object with the path to the needed cache settings properties (the limit of runtime memory, disk space, maximum file size, memory data cache size, options, and paths).
-  For instructions on how to get the path to the cache settings, see the [Build and Run on Linux](#build) section.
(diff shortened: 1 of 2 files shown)


docs/dataservice-cache-example.md
```diff
@@ -111,9 +111,19 @@ You need to create the `OlpClientSettings` object to get catalog and partition m
 
 3. [Authenticate](#authenticate-to-here-olp-using-client-credentials) to the HERE platform.
 
-4. Set up the `OlpClientSettings` object with the path to the needed cache settings properties (the limit of runtime memory, disk space, maximum file size, memory data cache size, options, and paths).
-  For instructions on how to get the path to the cache settings, see the [Build and Run on Linux](#build) section.
-  > **Note:** Perform the first call with the valid mutable cache path, and the second call – with the protected cache path.
+4. If your data has an expiration limit and is stored in the cache, to disable expiration, set the `default_cache_expiration` to `std::chrono::seconds::max()`.
+
```
kzvieriev, line 115: I am not sure about this.
By default expiry of some data(e.g. versionad layer) was  std::chrono::seconds::max() which means no expiry at all. This option gives possibility to change the behavior and make it expiry after awhile.
Data from volatile layer has expiry time and will not be effected by the option.

But from that description it seems vice versa for me.

HalynaDumych: Nice catch :)
Sorry, misunderstood the related comment in OlpClientSettings.h
Changed.

PR #898, opened 2020-06-06
Fix a crash inside InitializeLru

mykhailo-kuchma:
During testing was found that function InitializeLru could crash
during expiration parsing, the data length was not taken into account
when it was parsed, and std::stol throw an exception.

Resolves: OLPEDGE-2029

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -341,7 +341,11 @@ void DefaultCacheImpl::InitializeLru() {
       }
 
       if (expiration_key) {
-        props.expiry = std::stol(value.data());
+        // value.data() could point to a value without null character at the
+        // end, this could cause exception in std::stol. This is fixed by
+        // constructing a string, (We rely on small string optimization here).
+        std::string timestamp(value.data(), value.size());
+        props.expiry = std::stol(timestamp.c_str());
       } else {
         props.size = value.size();
       }


olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -341,7 +341,11 @@ void DefaultCacheImpl::InitializeLru() {
       }
 
       if (expiration_key) {
-        props.expiry = std::stol(value.data());
+        // value.data() could point to a value without null character at the
+        // end, this could cause exception in std::stol. This is fixed by
+        // constructing a string, (We rely on small string optimization here).
+        std::string timestamp(value.data(), value.size());
+        props.expiry = std::stol(timestamp.c_str());
```
andescu, line 348: Looking at https://en.cppreference.com/w/cpp/string/basic_string/stol, I see that there is a possibility to pass a size:

```cpp
long stol( const std::string& str, std::size_t* pos = 0, int base = 10 )
```

mykhailo-kuchma: It doesn't look like what we expect.
https://en.cppreference.com/w/cpp/string/basic_string/stol
pos - address of an integer to store the number of characters processed

andescu: Yes, looks like it. Sorry, I missed to read what it is used for.

PR #900, opened 2020-06-08
Add metadata api to get versions list

LiubovDidkivska:
Implement MetadataApi function.

Relates-To: OLPEDGE-1606

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/generated/api/MetadataApi.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/generated/api/MetadataApi.cpp
@@ -30,6 +30,7 @@
 #include "generated/parser/LayerVersionsParser.h"
 #include "generated/parser/PartitionsParser.h"
 #include "generated/parser/VersionResponseParser.h"
+#include "generated/parser/VersionInfosParser.h"
 #include <olp/core/generated/parser/JsonParser.h>
 // clang-format on
 
@@ -53,10 +54,9 @@ std::string concatStringArray(const std::vector<std::string>& strings,
 namespace olp {
 namespace dataservice {
 namespace read {
-using namespace olp::client;
 
 MetadataApi::LayerVersionsResponse MetadataApi::GetLayerVersions(
-    const OlpClient& client, int64_t version,
+    const client::OlpClient& client, std::int64_t version,
     boost::optional<std::string> billing_tag,
     const client::CancellationContext& context) {
   std::multimap<std::string, std::string> header_params;
@@ -74,7 +74,7 @@ MetadataApi::LayerVersionsResponse MetadataApi::GetLayerVersions(
                                      header_params, {}, nullptr, "", context);
 
   if (api_response.status != http::HttpStatusCode::OK) {
-    return ApiError(api_response.status, api_response.response.str());
+    return client::ApiError(api_response.status, api_response.response.str());
   }
 
   return LayerVersionsResponse(
@@ -82,8 +82,8 @@ MetadataApi::LayerVersionsResponse MetadataApi::GetLayerVersions(
 }
 
 MetadataApi::PartitionsResponse MetadataApi::GetPartitions(
-    const OlpClient& client, const std::string& layer_id,
-    boost::optional<int64_t> version,
+    const client::OlpClient& client, const std::string& layer_id,
+    boost::optional<std::int64_t> version,
     const std::vector<std::string>& additional_fields,
     boost::optional<std::string> range,
     boost::optional<std::string> billing_tag,
@@ -112,7 +112,7 @@ MetadataApi::PartitionsResponse MetadataApi::GetPartitions(
                                      header_params, {}, nullptr, "", context);
 
   if (api_response.status != http::HttpStatusCode::OK) {
-    return ApiError(api_response.status, api_response.response.str());
+    return {{api_response.status, api_response.response.str()}};
   }
 
   return PartitionsResponse(
@@ -120,7 +120,7 @@ MetadataApi::PartitionsResponse MetadataApi::GetPartitions(
 }
 
 MetadataApi::CatalogVersionResponse MetadataApi::GetLatestCatalogVersion(
-    const OlpClient& client, int64_t startVersion,
+    const client::OlpClient& client, std::int64_t startVersion,
     boost::optional<std::string> billing_tag,
     const client::CancellationContext& context) {
   std::multimap<std::string, std::string> header_params;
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
```diff
@@ -341,6 +342,12 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    */
   bool RemoveFromCache(const geo::TileKey& tile);
 
+  client::CancellationToken GetListVersions(VersionsRequest request,
+                                            VersionsResponseCallback callback);
+
+  client::CancellableFuture<VersionsResponse> GetListVersions(
+      VersionsRequest request);
+
```
andescu, line 350: This is a catalog specific API and does not belong to a layer. Also, it should be called `ListVersions` as **list** is already a verb and does not need help from **get** to describe the action. The meaning is _list the available catalog versions_ and not _get the list of catalog versions_.

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/CatalogClient.h
```diff
@@ -137,6 +138,12 @@ class DATASERVICE_READ_API CatalogClient final {
   client::CancellableFuture<CatalogVersionResponse> GetLatestVersion(
       CatalogVersionRequest request);
 
+  client::CancellationToken ListVersions(VersionsRequest request,
+                                         VersionsResponseCallback callback);
+
+  client::CancellableFuture<VersionsResponse> ListVersions(
+      VersionsRequest request);
```
andescu, line 145: Please add proper Doxygen.

olp-cpp-sdk-dataservice-read/src/CatalogClientImpl.cpp
```diff
@@ -124,6 +125,39 @@ CancellableFuture<CatalogVersionResponse> CatalogClientImpl::GetLatestVersion(
...
+        [=](client::CancellationContext context) -> VersionsResponse {
+      return repository::CatalogRepository::GetVersionsList(
+          std::move(catalog), context, std::move(request), std::move(settings));
+    };
+
+    return AddTask(settings.task_scheduler, pending_requests,
+                   std::move(versions_list_task), std::move(callback));
+  };
+
+  return ScheduleFetch(std::move(schedule_get_versions_list),
```
andescu, line 147: Is it even possible to have CacheWithUpdate? As soon as you have the same parameters you always get the same values. I don't think that CacheWithUpdate makes any sense for versioned catalogs.

olp-cpp-sdk-dataservice-read/src/CatalogClientImpl.cpp
```diff
@@ -124,6 +125,39 @@ CancellableFuture<CatalogVersionResponse> CatalogClientImpl::GetLatestVersion(
...
+  auto schedule_get_versions_list = [&](VersionsRequest request,
+                                        VersionsResponseCallback callback) {
+    auto catalog = catalog_;
+    auto settings = settings_;
+    auto pending_requests = pending_requests_;
+
+    auto versions_list_task =
+        [=](client::CancellationContext context) -> VersionsResponse {
+      return repository::CatalogRepository::GetVersionsList(
+          std::move(catalog), context, std::move(request), std::move(settings));
```
andescu, line 140: I wonder why do we need then copy parameters on CatalogRepository::GetVersionsList() if it is sync? Can we move towards pass by reference to save copies?

LiubovDidkivska: Could we do this in separate pull request and make this changes also in other clients?

olp-cpp-sdk-dataservice-read/src/repositories/CatalogCacheRepository.cpp
```diff
@@ -46,6 +48,11 @@ std::string CreateKey(const std::string& hrn) { return hrn + "::catalog"; }
 std::string VersionKey(const std::string& hrn) {
   return hrn + "::latestVersion";
 }
+std::string VersionInfosKey(const std::string& hrn, std::int16_t start,
+                            std::int16_t end) {
```
andescu, line 52: Why do you use int16_t here but below versions are int64_t?

olp-cpp-sdk-dataservice-read/src/repositories/CatalogCacheRepository.cpp
```diff
@@ -111,6 +118,31 @@ boost::optional<model::VersionResponse> CatalogCacheRepository::GetVersion() {
...
+
+  auto cached_versions = cache_->Get(key, [](const std::string& value) {
+    return parser::parse<model::VersionInfos>(value);
+  });
+
+  if (cached_versions.empty()) {
+    return boost::none;
+  }
+  return boost::any_cast<model::VersionInfos>(cached_versions);
+}
```
andescu, line 144: I would prefer if we implement all new cache repositories using the non decoder/encoder versions of Put/Get and convert themselves here in CatalogCacheRepository from JSON to char* and back.

```cpp
bool Put(const std::string& key, const ValueTypePtr value, time_t expiry = kDefaultExpiry);
ValueTypePtr Get(const std::string& key)
```

olp-cpp-sdk-dataservice-read/src/repositories/CatalogCacheRepository.cpp
```diff
@@ -111,6 +120,34 @@ boost::optional<model::VersionResponse> CatalogCacheRepository::GetVersion() {
...
 
+void CatalogCacheRepository::PutVersionInfos(
+    std::int64_t start, std::int64_t end, const model::VersionInfos& versions) {
+  std::string hrn(hrn_.ToCatalogHRNString());
+  auto key = VersionInfosKey(hrn, start, end);
+  OLP_SDK_LOG_DEBUG_F(kLogTag, "PutVersionInfos -> '%s'", key.c_str());
+  auto list_versions = olp::serializer::serialize(versions);
+
+  auto versions_data = std::make_shared<std::vector<unsigned char>>(
+      std::begin(list_versions), std::end(list_versions));
```
andescu, line 131: Something like this should be faster:

```cpp
std::make_shared<KeyValueCache::ValueType>(list_versions.data(), list_versions.data() + list_versions.size());
```

olp-cpp-sdk-dataservice-read/src/repositories/CatalogCacheRepository.cpp
```diff
@@ -111,6 +120,34 @@ boost::optional<model::VersionResponse> CatalogCacheRepository::GetVersion() {
...
+  auto key = VersionInfosKey(hrn, start, end);
+  OLP_SDK_LOG_DEBUG_F(kLogTag, "GetVersionInfos -> '%s'", key.c_str());
+
+  auto value = cache_->Get(key);
+  if (!value) {
+    return boost::none;
+  }
+
+  auto list_versions = std::string(value->begin(), value->end());
+  return parser::parse<model::VersionInfos>(list_versions);
```
andescu, line 148: Should be possible to add a new `parse` method, which takes in a `KeyValueCache::ValueTypePtr` or a `KeyValueCache::ValueType`.
Doesn't need to be in `core` for now it can just be in dataservice-read where the parsers are.

olp-cpp-sdk-dataservice-read/tests/CatalogCacheRepositoryTest.cpp
```diff
@@ -65,4 +68,50 @@ TEST(PartitionsCacheRepositoryTest, DefaultExpiry) {
   }
 }
 
+TEST(CatalogCacheRepositoryTest, VersionsList) {
```
andescu, line 71: Please also add a test case for `start_version = -1` as specified by the documentation:

**startVersion**
The beginning of the range of versions you want to get (exclusive). By convention -1 indicates the initial version before the first publication. After the first publication, the catalog version is 0.

LiubovDidkivska: Test will be added to CatalogRepositoryTest

PR #902, opened 2020-06-08
Moved AuthenticationClient::Impl to separate file.

andescu:
To lighten the AuthenticationClient.cpp file and align pimpl to other
classes, this commit moves AuthenticationClient::Impl to a separate
file and renames it to AuthenticationClientImpl.

Relates-To: OLPEDGE-1161


olp-cpp-sdk-authentication/src/AuthenticationClient.cpp
```diff
@@ -19,1124 +19,23 @@
...
 #include "olp/authentication/AuthenticationClient.h"
 
-#include <rapidjson/document.h>
-#include <rapidjson/istreamwrapper.h>
-#include <rapidjson/stringbuffer.h>
-#include <rapidjson/writer.h>
-
 #include <chrono>
 #include <iomanip>
 #include <sstream>
```
kzvieriev, line 24:
```suggestion
#include <utility>
```
Looks like we don't need these includes. Only std::move and SDK classes used here.
Ignore this if I am wrong ;-)

andescu: Good catch. Resolved.

PR #911, opened 2020-06-11
Merge concurrent API requests.

mykhailo-kuchma:
Exact same API requests are blocked, and wait for the first one to
complete.

Relates-To: OLPEDGE-2040

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
@@ -30,6 +30,7 @@
 #include "CatalogRepository.h"
 #include "DataCacheRepository.h"
 #include "ExecuteOrSchedule.inl"
+#include "NamedMutex.h"
 #include "PartitionsCacheRepository.h"
 #include "PartitionsRepository.h"
 #include "generated/api/BlobApi.h"
@@ -147,6 +148,14 @@ DataResponse DataRepository::GetBlobData(
 
   if (!data_handle) {
     return {{client::ErrorCode::PreconditionFailed, "Data handle is missing"}};
+  }
+
+  NamedMutex mutex(catalog.ToString() + layer + *data_handle);
+  std::unique_lock<NamedMutex> lock(mutex, std::defer_lock);
+
+  // If we are not planning to go online or access the cache, do not lock.
+  if (fetch_option != CacheOnly && fetch_option != OnlineOnly) {
+    lock.lock();
   }
 
   repository::DataCacheRepository repository(catalog, settings.cache,
--- a/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
@@ -367,6 +367,18 @@ PartitionResponse PartitionsRepository::GetAggregatedTile(
     const client::OlpClientSettings& settings) {
   const auto fetch_option = request.GetFetchOption();
   const auto& tile_key = request.GetTileKey();
+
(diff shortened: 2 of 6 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
```diff
@@ -149,6 +136,14 @@ DataResponse DataRepository::GetBlobData(
     return {{client::ErrorCode::PreconditionFailed, "Data handle is missing"}};
   }
 
+  NamedMutex mutex(catalog.ToString() + *data_handle);
+  std::unique_lock<NamedMutex> lock(mutex, std::defer_lock);
+
```
kzvieriev, line 141: Maybe add layer to the mutex name.

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
```diff
@@ -93,20 +94,6 @@ DataResponse DataRepository::GetVersionedData(
...
-      version_request.WithFetchOption(request.GetFetchOption())
-          .WithBillingTag(request.GetBillingTag());
-      auto latest_version_response =
-          repository::CatalogRepository::GetLatestVersion(
-              catalog, context, std::move(version_request), settings);
-      if (!latest_version_response.IsSuccessful()) {
-        return latest_version_response.GetError();
-      }
-      request.WithVersion(latest_version_response.GetResult().GetVersion());
-    }
```
kzvieriev, line 108: DataRequest::GetVersion is deprecated, but I think we can't just remove this block.
GetPartitionById() used below expecting for a version in the request, so this change breaking DataRepository::GetVersionedData() logic.
Is this related to merging requests?

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
andescu: Good point. According to DS documentation, the data handle is only to be unique per layer.

mykhailo-kuchma: Added layer

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
mykhailo-kuchma: Actually we don't enter this branch. But I reverted this code for now.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -367,6 +367,18 @@ PartitionResponse PartitionsRepository::GetAggregatedTile(
     const client::OlpClientSettings& settings) {
   const auto fetch_option = request.GetFetchOption();
   const auto& tile_key = request.GetTileKey();
+
+  const auto& root_tile_key = tile_key.ChangedLevelBy(-kAggregateQuadTreeDepth);
+  const auto root_tile_here = root_tile_key.ToHereTile();
+
+  NamedMutex mutex(catalog.ToString() + layer + root_tile_here + "Index");
+  std::unique_lock<NamedMutex> lock(mutex, std::defer_lock);
```
kzvieriev, line 375: This is tricky one :-)
GetAggregatedTile is looking in cache for any tree containing it, not only `tile_key.ChangedLevelBy(-4)`.
So we need to block loading of quad trees from root_tile_key.Level() to tile_key.Level(). Not sure this can be done using just string.

mykhailo-kuchma: It's needed. Since this is the tile that will be used for API call. Our goal is to merge the online requests.

PR #917, opened 2020-06-15
Change hard-coded values to HttpStatusCodes 

LiubovDidkivska:
Use defines from HttpStatusCodes.h instead of hard-coded values

Relates-To: OLPEDGE-686

### DIFF
--- a/olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
+++ b/olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "olp/core/context/Context.h"
+#include "olp/core/http/HttpStatusCode.h"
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-core/src/http/android/NetworkAndroid.cpp
```diff
@@ -604,7 +607,8 @@ void NetworkAndroid::HeadersCallback(JNIEnv* env, RequestId request_id,
 void NetworkAndroid::DateAndOffsetCallback(JNIEnv* env, RequestId request_id,
                                            jlong date, jlong offset) {
   std::unique_lock<std::mutex> lock(requests_mutex_);
-  if (!started_) return;
+  if (!started_)
+    return;
```
andescu, line 611: Always add braces.

PR #925, opened 2020-06-16
Change hard-coded values to HttpStatusCodes

LiubovDidkivska:
Use defines from HttpStatusCodes.h instead of hard-coded values

Relates-To: OLPEDGE-686


olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -821,7 +821,8 @@ void NetworkCurl::CompleteMessage(CURL* handle, CURLcode result) {
       long http_status = 0;
       curl_easy_getinfo(rhandle.handle, CURLINFO_RESPONSE_CODE, &http_status);
       status = static_cast<int>(http_status);
-      if ((rhandle.offset == 0) && (status == 206)) {
+      if ((rhandle.offset == 0) &&
+          (status == olp::http::HttpStatusCode::PARTIAL_CONTENT)) {
         status = 200;
```
mykhailo-kuchma, line 826: http::HttpStatusCode::OK

PR #934, opened 2020-06-25
Implement Prefetch caching with QuadTreeIndex

LiubovDidkivska:
Change caching to QuadTreeIndex.
Always query depth 4. Update tests.

Relates-To: OLPEDGE-2084

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
@@ -53,7 +53,8 @@ constexpr std::uint32_t kMaxQuadTreeIndexDepth = 4u;
 
 void PrefetchTilesRepository::SplitSubtree(
     RootTilesForRequest& root_tiles_depth,
-    RootTilesForRequest::iterator subtree_to_split, geo::TileKey tile_key) {
+    RootTilesForRequest::iterator subtree_to_split,
+    const geo::TileKey& tile_key, unsigned int min) {
   unsigned int depth = subtree_to_split->second;
   auto tileKey = subtree_to_split->first;
   if (depth <= kMaxQuadTreeIndexDepth) {
@@ -72,21 +73,20 @@ void PrefetchTilesRepository::SplitSubtree(
     for (std::uint64_t key = beginTileKey; key < endTileKey; ++key) {
       auto child = geo::TileKey::FromQuadKey64(key);
       // skip child, if it is not a parent, or a child of prefetched tile
-      if (!tile_key.IsParentOf(child) && !child.IsParentOf(tile_key) &&
-          child != tile_key) {
+      if ((!tile_key.IsParentOf(child) && !child.IsParentOf(tile_key) &&
+           child != tile_key) ||
+          (child.Level() + kMaxQuadTreeIndexDepth < min)) {
         continue;
       }
-
-      auto it = root_tiles_depth.insert({child, kMaxQuadTreeIndexDepth});
-      // element already exist, update depth with max value(should never
-      // hapen)
-      if (it.second == false) {
-        it.first->second = kMaxQuadTreeIndexDepth;
-      }
+      root_tiles_depth.insert({child, kMaxQuadTreeIndexDepth});
     }
     depth -= (kMaxQuadTreeIndexDepth + 1);
   }
-  subtree_to_split->second = depth;
+  if (tileKey.Level() + depth < min) {
+    root_tiles_depth.erase(subtree_to_split);
+  } else {
+    subtree_to_split->second = depth;
+  }
 }
 
 RootTilesForRequest PrefetchTilesRepository::GetSlicedTiles(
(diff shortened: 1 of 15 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
```diff
@@ -177,56 +225,37 @@ SubQuadsResponse PrefetchTilesRepository::GetSubQuads(
...
-      catalog, settings.cache, settings.default_cache_expiration);
-  cache.Put(PartitionsRequest().WithVersion(version), partitions, layer_id,
-            boost::none, false);
-
-  return result;
+  repository.Put(layer_id, tile, depth, tree, version);
+  return get_sub_quads(tree);
 }
 
 SubQuadsResponse PrefetchTilesRepository::GetVolatileSubQuads(
```
kzvieriev, line 261: For volatile client `GetVolatileSubQuads` used instead of `GetSubQuads`, so need update it too

olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
```diff
@@ -169,6 +191,32 @@ SubQuadsResponse PrefetchTilesRepository::GetSubQuads(
...
+        index_data.begin(), index_data.end(),
+        std::inserter(result, result.end()),
+        [](const QuadTreeIndex::IndexData& data) -> SubQuadsResult::value_type {
+          return {data.tile_key, data.data_handle};
+        });
+    return result;
+  };
+
+  // check if quad tree with requested tile and depth already in cache
+  auto cached_tree = repository.Get(layer_id, tile, depth, version);
```
kzvieriev, line 210: Don't we need to search for any quad tree with metadata of the tile?
Its fine for tiles from one Prefetch request, since their level differs on `kMaxQuadTreeIndexDepth+1`, but for tiles from different Prefetch requests metadata from quad tree will duplicate. We are doing search in GetData/GetAggregatedData  to avoid such extra quad trees in the cache.

olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
LiubovDidkivska: This change will be postponed, because we don't have TileRequest for volatile layer client

olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
LiubovDidkivska: This is tricky, if we will search quad tree on different levels, we should check if it covers full tree, if not, request missing tiles. And even than we can have duplicates, because for example the missing coverage only one level in between of cached ones.

olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
```diff
@@ -67,58 +71,75 @@ void PrefetchTilesRepository::SplitSubtree(
...
-      if (max_level == 0) {  // special case, adjust levels to input tile level
-        max_level = tile_key.Level();
-        min_level = (max_level > kMaxQuadTreeIndexDepth)
-                        ? (max_level - kMaxQuadTreeIndexDepth)
-                        : (1u);
+    // min level should always start from level of root tile for quad tree
+    // index, if requested tile is on upper level, adjust min_level up
+    if (tile_key.Level() < min_level || min_level == 0) {
+      min_level = tile_key.Level();
+      max_level = (max_level < min_level) ? min_level : max_level;
```
andescu, line 106:
```suggestion
      max_level = std::max(max_level, min_level);
```

olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
```diff
@@ -67,58 +71,75 @@ void PrefetchTilesRepository::SplitSubtree(
...
-        // if depth is greater than kMaxQuadTreeIndexDepth, need to split
-        if (it.first->second > kMaxQuadTreeIndexDepth) {
-          // split subtree on chunks with depth kMaxQuadTreeIndexDepth
-          SplitSubtree(root_tiles_depth, it.first);
-        }
-      }
-      tile_key = parent;
+    OLP_SDK_LOG_DEBUG_F(
+        kLogTag, "GetSlicedTiles for tile %s use min='%d', max='%d' levels",
+        tile_key.ToHereTile().c_str(), min_level, max_level);
```
andescu, line 128: Add empty newline after log.

olp-cpp-sdk-dataservice-read/src/repositories/PrefetchTilesRepository.cpp
```diff
@@ -177,56 +224,37 @@ SubQuadsResponse PrefetchTilesRepository::GetSubQuads(
...
-        PartitionsRepository::PartitionFromSubQuad(*subquad,
-                                                   subtile.ToHereTile()));
+  if (tree.IsNull()) {
+    OLP_SDK_LOG_WARNING_F(kLogTag,
+                          "QuadTreeIndex failed, hrn='%s', "
+                          "layer='%s', root='%s', version='%" PRId64
+                          "', depth='%" PRId32 "'",
+                          catalog.ToString().c_str(), layer_id.c_str(),
+                          tile_key.c_str(), version, depth);
+    return {{client::ErrorCode::Unknown, "Failed to parse QuadTreeIndex json"}};
```
andescu, line 253:
```suggestion
    return {{client::ErrorCode::Unknown, "Failed to parse quad tree response"}};
```

PR #935, opened 2020-06-25
Remove using from tests

LiubovDidkivska:
Remove using from tests

Relates-To: OLPEDGE-2118


olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
```diff
@@ -29,9 +29,11 @@
...
 
 namespace {
-using namespace olp::dataservice::read;
-using namespace ::testing;
-using namespace olp::tests::common;
+namespace read = olp::dataservice::read;
+namespace model = olp::dataservice::read::model;
+using ::testing::_;
+using ::testing::Mock;
+namespace common = olp::tests::common;
```
andescu, line 36: We should not use namespace in test. I would propose to remove the olp::tests::common and not polute the olp namespace further.

PR #937, opened 2020-06-25
Mark SignInGoogle as deprecated.

kzvieriev:
Sign with Google token is not supported anymore and so marked as
deprecated.

Resolves: OLPEDGE-2087


olp-cpp-sdk-authentication/include/olp/authentication/AuthenticationClient.h
```diff
@@ -395,7 +395,11 @@ class AUTHENTICATION_API AuthenticationClient {
    *
    * @return The `CancellationToken` instance that can be used to cancel
    * the request.
+   *
+   * @deprecated Will be removed by 12.2020.
    */
+  OLP_SDK_DEPRECATED(
+      "Signs in with Google token is deprecated and will be removed by 12.2020")
```
andescu, line 402:
```suggestion
      "Sign in with Google token is deprecated and will be removed by 12.2020")
```

PR #941, opened 2020-07-01
Add check if a certain key/value cached.

LiubovDidkivska:
Add functionality to check if key in the cache.
Add tests.
    
Resolves: OLPEDGE-2110
    

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -317,6 +317,18 @@ bool DiskCache::Get(const std::string& key,
...
 }
 
+bool DiskCache::Exist(const std::string& key) {
+  if (!database_) {
+    OLP_SDK_LOG_ERROR(kLogTag, "Get: Database is not initialized");
+    return false;
+  }
+  leveldb::ReadOptions options;
+  options.verify_checksums = check_crc_;
+  auto iterator = NewIterator(options);
```
kzvieriev, line 327: Set `fill_cache` option to false, so we don't load data.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -303,6 +303,21 @@ bool DefaultCacheImpl::RemoveKeysWithPrefix(const std::string& key) {
...
+    auto iterator = mutable_cache_lru_->FindNoPromote(key);
+    return (iterator != mutable_cache_lru_->end());
+  }
+  if (mutable_cache_ && mutable_cache_->Exist(key)) {
+    return true;
+  }
+  if (protected_cache_ && protected_cache_->Exist(key)) {
+    return true;
+  }
+  return false;
```
kzvieriev, line 318: I believe, need to check memory cache too.
Also no need to check mutable cache if key not found in the lru.

olp-cpp-sdk-core/include/olp/core/cache/KeyValueCache.h
```diff
@@ -121,6 +121,14 @@ class CORE_API KeyValueCache {
...
    */
   virtual bool RemoveKeysWithPrefix(const std::string& prefix) = 0;
+  /**
+   * @brief Check if key is in the cache.
+   *
+   * @param key The key for the value.
+   *
+   * @return True if the key/value cached ; false otherwise.
+   */
+  virtual bool Exist(const std::string& key) = 0;
```
andescu, line 131: Lets call it Contains().
NOTE: This is a breaking change! This means all user code will not compile as there is a new abstract API that is not overridden. You need to provide a default impl to not break API.

```suggestion
  virtual bool Contains(const std::string& key) {
      CORE_UNUSED(key);
   };
```

olp-cpp-sdk-core/include/olp/core/cache/KeyValueCache.h
```diff
@@ -121,6 +121,14 @@ class CORE_API KeyValueCache {
    * @return True if the values are removed; false otherwise.
    */
   virtual bool RemoveKeysWithPrefix(const std::string& prefix) = 0;
+  /**
```
andescu, line 124: Seems to be a recurring issue.
Please separate methods with an empty newline.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -303,6 +303,30 @@ bool DefaultCacheImpl::RemoveKeysWithPrefix(const std::string& key) {
   return true;
 }
 
+bool DefaultCacheImpl::Exist(const std::string& key) {
```
andescu, line 306: Please separate individual blocks by a newline.

olp-cpp-sdk-core/src/cache/DiskCache.h
```diff
@@ -141,6 +141,8 @@ class DiskCache {
   bool RemoveKeysWithPrefix(const std::string& prefix,
                             uint64_t& removed_data_size);
 
+  bool Exist(const std::string& key);
```
andescu, line 144:
```suggestion
  bool Contains(const std::string& key);
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -303,6 +303,30 @@ bool DefaultCacheImpl::RemoveKeysWithPrefix(const std::string& key) {
...
+  std::lock_guard<std::mutex> lock(cache_lock_);
+  if (!is_open_) {
+    return false;
+  }
+  if (memory_cache_ && !(memory_cache_->Get(key).empty())) {
+    return true;
+  }
```
kzvieriev, line 313: Can we add Exist() method for memory cache too, so we don't copy data from it?

olp-cpp-sdk-core/src/cache/InMemoryCache.cpp
```diff
@@ -105,6 +105,22 @@ void InMemoryCache::RemoveKeysWithPrefix(const std::string& key_prefix) {
   }
 }
 
+bool InMemoryCache::Contains(const std::string& key) {
+  std::lock_guard<std::mutex> lock{mutex_};
+  auto it = item_tuples_.Find(key);
+  if (it != item_tuples_.end()) {
+    auto expiry_time = std::get<1>(it.value());
+    if (expiry_time < time_provider_()) {
+      PurgeExpired(expiry_time);
```
andescu, line 114: Why do we erase during checking if it exists? We should only delete if we add something new.

PR #943, opened 2020-07-03
Implement cache protection functionality

LiubovDidkivska:
Add read/write protected list to mutable cache.
Filter keys besed on protected list.
Do not add protected keys to lru, that they
could not be evicted. Reset expiration for protected keys,
So they could not expire. Add functionality to release
keys from protected list, so they will be added to lru,
and could expire or be evicted. Add unit tests for protected
keys and default cache implementation.

Relates-To: OLPEDGE-2124
    

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
```diff
@@ -113,19 +119,37 @@ class DefaultCacheImpl {
...
 
   boost::optional<std::pair<std::string, time_t>> GetFromDiscCache(
       const std::string& key);
 
+  // custom comparator needed to reduce duplicates for keys, which are already
+  // protected by prefix
+  struct CustomCompare {
+    bool operator()(const std::string& lhs, const std::string& rhs) const {
+      if (rhs.length() < lhs.length()) {
+        if (lhs.substr(0, rhs.length()) == rhs)
```
andescu, line 135: substr() creates a copy of a string which is very inefficient. Use std::string::find().

olp-cpp-sdk-core/src/cache/ProtectedData.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include "olp/core/cache/DefaultCache.h"
+
+#include <set>
+#include <string>
+
+namespace olp {
+namespace cache {
+
+class ProtectedData {
```
andescu, line 30: What you are actually protecting is a key list. So I would call it `ProtectedKeyList`

olp-cpp-sdk-core/src/cache/ProtectedData.h
```diff
@@ -0,0 +1,74 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#include "olp/core/cache/DefaultCache.h"
```
andescu, line 22: Add the KeyValueCache abstract interface and not the implementation.

olp-cpp-sdk-core/src/cache/ProtectedData.h
```diff
@@ -0,0 +1,74 @@
...
+namespace olp {
+namespace cache {
+
+class ProtectedData {
+ public:
+  using ProtectedKeyChanged = std::function<void(const std::string&)>;
+
+  ProtectedData();
+
+  ~ProtectedData();
```
andescu, line 36: Mark as default here. Also as you provide a custom destructor you will not get any move or copy operators. Probably you will need some move operators.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -222,8 +249,12 @@ boost::any DefaultCacheImpl::Get(const std::string& key,
   if (disc_cache) {
     auto decoded_item = decoder(disc_cache->first);
     if (memory_cache_) {
-      memory_cache_->Put(key, decoded_item, disc_cache->second,
-                         disc_cache->first.size());
+      auto expiry = disc_cache->second;
+      // reset expiry if key is protected only for memory cache
+      if (IsProtected(key)) {
+        expiry = olp::cache::KeyValueCache::kDefaultExpiry;
+      }
```
andescu, line 256:
```suggestion
      // reset expiry if key is protected only for memory cache
      auto expiry = IsProtected(key) ? KeyValueCache::kDefaultExpiry : disc_cache->second;
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -168,7 +184,13 @@ bool DefaultCacheImpl::Put(const std::string& key, const boost::any& value,
   auto encoded_item = encoder();
   if (memory_cache_) {
     const auto size = encoded_item.size();
-    const bool result = memory_cache_->Put(key, value, expiry, size);
+    // reset expiry if key is protected only for memory cache, in mutable cache
```
andescu, line 187: If a tile is protected it should not have any expiry also. Please also reset the expiry for the mutable cache.
And if a protected cache entry cannot expire you should not need to have all this "reset expiry" blocks as the expiry value in the mutable cache is already kDefaultExpiry.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -168,7 +184,13 @@ bool DefaultCacheImpl::Put(const std::string& key, const boost::any& value,
...
+    // store expiry as usual
+    auto expiry_copy = expiry;
+    if (IsProtected(key)) {
+      expiry_copy = olp::cache::KeyValueCache::kDefaultExpiry;
+    }
```
andescu, line 192:
```suggestion
    auto expiry_copy = IsProtected(key) ? KeyValueCache::kDefaultExpiry : expiry;
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -168,7 +184,13 @@ bool DefaultCacheImpl::Put(const std::string& key, const boost::any& value,
   auto encoded_item = encoder();
   if (memory_cache_) {
     const auto size = encoded_item.size();
-    const bool result = memory_cache_->Put(key, value, expiry, size);
+    // reset expiry if key is protected only for memory cache, in mutable cache
+    // store expiry as usual
+    auto expiry_copy = expiry;
+    if (IsProtected(key)) {
+      expiry_copy = olp::cache::KeyValueCache::kDefaultExpiry;
```
andescu, line 191: You are already in the `olp::cache` namespace, no need to mention it again.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -113,11 +114,26 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::Open() {
...
+    auto value = protected_keys_.WriteListOfProtectedKeys();
+    if (value->size() > 0) {
+      leveldb::Slice slice(reinterpret_cast<const char*>(value->data()),
+                           value->size());
+      batch->Put(kProtectedKeys, slice);
+      auto result = mutable_cache_->ApplyBatch(std::move(batch));
+      OLP_SDK_LOG_DEBUG_F(kLogTag, "Store list of protected keys result=%s",
+                          result.IsSuccessful() ? "true" : "false");
+    }
+  }
```
andescu, line 136: I would also consider writing the protected list to cache on Protect if it changed.

olp-cpp-sdk-core/src/cache/ProtectedData.cpp
```diff
@@ -0,0 +1,157 @@
...
+      if (IsPrefix(*hint, key)) {
+        continue;
+      }
+      // if key is prefix for hint key, remove hint and add prefix
+      while (hint != protected_data_.end() && IsPrefix(key, *hint)) {
+        hint = protected_data_.erase(hint);
+      }
+    }
+
+    // remove keys from lru and memory cache
```
andescu, line 90: This is irrelevant here and should not be covered by this class, means remove comment or change to something more generic like: "Notify listener that keys have changed".
Also, why do you need to remove something from the memory cache?

olp-cpp-sdk-core/src/cache/ProtectedData.cpp
```diff
@@ -0,0 +1,157 @@
...
+    if (hint != protected_data_.end()) {
+      // if hint is prefix for this key, ignore this key, it is allready
+      // protected
+      if (IsPrefix(*hint, key)) {
+        continue;
+      }
+      // if key is prefix for hint key, remove hint and add prefix
+      while (hint != protected_data_.end() && IsPrefix(key, *hint)) {
+        hint = protected_data_.erase(hint);
+      }
```
andescu, line 87: Do you think it is bad that we have keys which are covering the same area? I dont think it is bad and probably we should not spend so much effort. If the user protects one tile and the entire catalog we have two entries here which eventually will result in the correct data being protected. The only downfall is that we use a little bit more leveldb space when this list is persisted.

olp-cpp-sdk-core/src/cache/ProtectedData.h
```diff
@@ -0,0 +1,74 @@
...
+
+  bool Protect(const DefaultCache::KeyListType& keys,
+               const ProtectedKeyChanged& change_key_to_protected);
+
+  bool Release(const DefaultCache::KeyListType& keys,
+               const ProtectedKeyChanged& released_key_from_protected);
+
+  bool ReadListOfProtectedKeys(KeyValueCache::ValueTypePtr value);
```
andescu, line 44: Please use something like `Serialize/Deserialize` and not such exotic names.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
LiubovDidkivska: marking list dirty on Protect and Release. Writing on put, or on close

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
```diff
@@ -27,13 +27,16 @@
...
+#include "ProtectedKeyList.h"
 
 namespace olp {
 namespace cache {
 
 class DefaultCacheImpl {
  public:
-  DefaultCacheImpl(CacheSettings settings);
+  explicit DefaultCacheImpl(CacheSettings settings);
+
```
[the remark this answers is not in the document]
LiubovDidkivska: all functions are separated by empty line

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
```diff
@@ -96,17 +101,21 @@ class DefaultCacheImpl {
...
 
   /// Removes key from the mutable lru cache;
-  void RemoveKeyLru(const std::string& key);
+  bool RemoveKeyLru(const std::string& key);
 
   /// Removes all keys with specified prefix from LRU mutable cache.
   void RemoveKeysWithPrefixLru(const std::string& key);
 
-  /// Returns true if key is found in the LRU cache, false - otherwise.
+  /// Returns true if key is found in the LRU cache, or protected false -
```
andescu, line 109:
```suggestion
  /// Returns true if key is found in the LRU or protected cache, false -
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
```diff
@@ -119,13 +128,18 @@ class DefaultCacheImpl {
   boost::optional<std::pair<std::string, time_t>> GetFromDiscCache(
       const std::string& key);
 
+  time_t GetExpiryForMemoryCache(const std::string& key, const time_t& expiry);
```
andescu, line 131: Can be an anonymous function.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -681,5 +747,82 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
...
+  protected_keys_.Protect(keys, [&](const std::string& key) {
+    if (!RemoveKeyLru(key)) {
+      RemoveKeysWithPrefixLru(key);
+    }
+  });
+  if (memory_cache_) {
+    memory_cache_->Clear();
+  }
+
+  OLP_SDK_LOG_INFO_F(kLogTag, "Protect, time=%" PRId64 " ms",
```
andescu, line 762: Please also print the size of the input keys.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -681,5 +747,82 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
...
+          ValueProperties props;
+          props.size = value->size();
+          props.expiry = expiry;
+          mutable_cache_lru_->InsertOrAssign(key, props);
+        }
+      } else {
+        // key is not in mutable_cache_lru_, key is prefix, need reload memory
+        // cache and lru to store all keys with prefix
+        need_reset = true;
+        OLP_SDK_LOG_INFO(kLogTag,
```
andescu, line 795: Merge into below log.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -681,5 +747,82 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
...
+
+  if (need_reset) {
+    if (memory_cache_) {
+      memory_cache_->Clear();
+    }
+    mutable_cache_lru_.reset();
+    InitializeLru();
+  }
+
+  OLP_SDK_LOG_INFO_F(kLogTag, "Release, time=%" PRId64 " ms",
```
andescu, line 809: Please also print the size of the input list and the total size of the protected list.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -681,5 +747,82 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
...
+
+      auto result = GetFromDiskCache(key, value, expiry);
+      if (result && value) {
+        if (memory_cache_) {
+          if (expiry > 0) {
+            memory_cache_->Put(key, value, expiry, value->size());
+          } else {
+            memory_cache_->Remove(key);
+          }
+        }
```
andescu, line 784: This block is a little strange to me. Why do you put the released key into `memory_cache_`?

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -681,5 +747,82 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
...
+        if (expiry > 0) {
```
andescu, line 785: Why does the expiry matter to put it in the LRU cache? If a key was released from protection we should add it to LRU either way so that it can be removed when DB size is growing. Is there a reason to only put it if it still has expiry left?

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -598,6 +654,13 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::SetupStorage() {
       mutable_cache_.reset();
       settings_.disk_path_mutable = boost::none;
       result = DefaultCache::OpenDiskPathFailure;
+    } else {
+      // read protected keys
+      KeyValueCache::ValueTypePtr value = nullptr;
+      auto result = mutable_cache_->Get(kProtectedKeys, value);
+      if (result && value) {
+        protected_keys_.Deserialize(value);
```
andescu, line 662: Please check the return value and add a log in case it failed.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -681,5 +747,82 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
...
+}
+
+time_t DefaultCacheImpl::GetExpiryForMemoryCache(const std::string& key,
+                                                 const time_t& expiry) {
+  // reset expiry if key is protected for memory cache
+  if (protected_keys_.IsProtected(key)) {
+    return KeyValueCache::kDefaultExpiry;
+  }
+  return expiry;
+}
```
andescu, line 821: This method does basically nothing else than the ternary operator. Please make it an anonymous function. If you have a method which does not even use any member, then why is it a method at all?

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -681,5 +747,82 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
...
+
+bool DefaultCacheImpl::IsInternalKey(const std::string& key) const {
```
andescu, line 823: Same here. Why is this a method?

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -93,6 +95,11 @@ leveldb::CompressionType GetCompression(
              : leveldb::kSnappyCompression;
 }
 
+int64_t GetElapsedTimeFromStart(std::chrono::steady_clock::time_point start) {
```
andescu, line 98:
```suggestion
int64_t GetElapsedTime(std::chrono::steady_clock::time_point start) {
```
It is quite obvious that it is from start, no need to extra mention it.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -113,11 +120,20 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::Open() {
...
 void DefaultCacheImpl::Close() {
   std::lock_guard<std::mutex> lock(cache_lock_);
   if (!is_open_) {
     return;
   }
+  if (mutable_cache_ && protected_keys_.IsListDirty()) {
+    auto batch = std::make_unique<leveldb::WriteBatch>();
+    MaybeUpdatedProtectedKeys(*batch);
+    auto result = mutable_cache_->ApplyBatch(std::move(batch));
+    OLP_SDK_LOG_DEBUG_F(kLogTag, "Store list of protected keys result=%s",
```
andescu, line 134:
```suggestion
    OLP_SDK_LOG_INFO_F(kLogTag, "Close(): store list of protected keys, result=%s",
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -164,11 +180,17 @@ bool DefaultCacheImpl::Put(const std::string& key, const boost::any& value,
   if (!is_open_) {
     return false;
   }
+  if (IsInternalKey(key)) {
+    OLP_SDK_LOG_WARNING_F(kLogTag, "Failed to write key with prefix %s",
+                          kInternalKeysPrefix);
+    return false;
+  }
```
andescu, line 187: Why? Maybe we need it on the higher level also. Why should we block this?

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -186,10 +208,16 @@ bool DefaultCacheImpl::Put(const std::string& key,
   if (!is_open_) {
     return false;
   }
+  if (IsInternalKey(key)) {
```
andescu, line 211: Same here.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -681,5 +749,59 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
...
+  if (memory_cache_) {
+    memory_cache_->Clear();
+  }
+  if (mutable_cache_) {
+    auto it = mutable_cache_->NewIterator(leveldb::ReadOptions());
+
+    for (it->SeekToFirst(); it->Valid(); it->Next()) {
+      AddKeyToLru(it->key().ToString(), it->value());
+    }
+  }
```
andescu, line 787: Why are you adding each key to LRU again? You should only add released keys, resp. keys that have the same prefix as the released key.

PR #949, opened 2020-07-08
Add Prefetch test for mock server

LiubovDidkivska:
Split dataservice read and api responses.
Move generating QuadTreeIndex responce to
ReadDefaultResponses. Implement Prefetch test
for mock server

Relates-To: OLPEDGE-1961


tests/functional/olp-cpp-sdk-dataservice-read/VersionedLayerClientPrefetchTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+          mockserver::ApiDefaultResponses::GenerateResourceApisResponse(
+              kTestHrn));
+      mock_server_client_->MockGetVersionResponse(
+          mockserver::ReadDefaultResponses::GenerateVersionResponse(kVersion));
+      mock_server_client_->MockGetResponse(
+          kLayer, root_tile, kVersion,
+          mockserver::ReadDefaultResponses::GenerateQuadTreeResponse(
+              root_tile, kQuadTreeDepth, {12}));
+      const olp::geo::TileKey firstChild = root_tile.ChangedLevelBy(1);
+      const std::uint64_t beginTileKey = firstChild.ToQuadKey64();
```
kzvieriev, line 108:
```suggestion
      const olp::geo::TileKey first_child = root_tile.ChangedLevelBy(1);
      const std::uint64_t begin_tile_key = first_child.ToQuadKey64();
```

tests/functional/olp-cpp-sdk-dataservice-read/VersionedLayerClientPrefetchTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+      mock_server_client_->MockLookupResourceApiResponse(
+          mockserver::ApiDefaultResponses::GenerateResourceApisResponse(
+              kTestHrn));
+      mock_server_client_->MockGetVersionResponse(
+          mockserver::ReadDefaultResponses::GenerateVersionResponse(kVersion));
+      mock_server_client_->MockGetResponse(
+          kLayer, root_tile, kVersion,
+          mockserver::ReadDefaultResponses::GenerateQuadTreeResponse(
+              root_tile, kQuadTreeDepth, {12}));
+      const olp::geo::TileKey firstChild = root_tile.ChangedLevelBy(1);
```
kzvieriev, line 107: Test seems wrong for me. `tile` level is 12, generated quad tree has only tiles of level 12, so result of prefetch should be `tile` only. Why we get 4 tiles of level 13 instead?

tests/common/ApiDefaultResponses.h
```diff
@@ -0,0 +1,86 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#include <random>
```
kzvieriev, line 22: <random> now used here.

tests/common/ApiDefaultResponses.h
```diff
@@ -0,0 +1,86 @@
...
+#include <rapidjson/writer.h>
+
+#include "generated/model/Api.h"
+
+namespace mockserver {
+
+class ApiDefaultResponses {
+ public:
+  static olp::dataservice::read::model::Apis GenerateResourceApisResponse(
+      std::string catalog) {
```
kzvieriev, line 39: Looks like `ReadApiDefaultResponses` :-)
The problem here `WriteDefaultResponses` has same generate* functions but for `write::model::Apis`. We are going to add LookupApiClient soon and `core::model::Apis` will be used for both libs.
Maybe better to postpone the split till we can use core model and `LookupApiEndpointProvider` stuff?

tests/functional/olp-cpp-sdk-dataservice-read/VersionedLayerClientPrefetchTest.cpp
LiubovDidkivska: It is valid that we receiving from server not all tiles, but only available. You can see it on old prefetch tests.

tests/common/ApiDefaultResponses.h
LiubovDidkivska: Split is needed to be able to move generation of QuadTreeIndex to common lib.

tests/functional/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ set(OLP_SDK_FUNCTIONAL_TESTS_SOURCES
     ./olp-cpp-sdk-dataservice-read/ApiTest.cpp
     ./olp-cpp-sdk-dataservice-read/CatalogClientTest.cpp
     ./olp-cpp-sdk-dataservice-read/VersionedLayerClientTest.cpp
+        ./olp-cpp-sdk-dataservice-read/VersionedLayerClientPrefetchTest.cpp
```
meravingen123, line 24: alignmen t

PR #954, opened 2020-07-13
Add check if partition/tile cached

LiubovDidkivska:
Add implementation to VersionLayerClientImpl
if partition/tile cached. Added check to
DataCacheRepository if data handle cached.
Add tests for PartitionCacheRepository to check
if partition is cached.

Relates-To: OLPEDGE-2111

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -19,6 +19,8 @@
 
 #include "VersionedLayerClientImpl.h"
 
+#include <string>
+
 #include <olp/core/cache/DefaultCache.h>
 #include <olp/core/client/OlpClientSettingsFactory.h>
 #include <olp/core/client/PendingRequests.h>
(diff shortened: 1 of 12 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
```diff
@@ -279,6 +282,29 @@ bool PartitionsCacheRepository::ClearPartitionMetadata(
...
+        });
+
+    if (cached_partition.empty()) {
+      return false;
+    }
+    auto partition = boost::any_cast<model::Partition>(cached_partition);
+    auto data_key = CreateKey(hrn, layer_id, partition.GetDataHandle());
+    OLP_SDK_LOG_DEBUG_F(kLogTag, "IsPartitionCached data -> '%s'", key.c_str());
+    return cache_->Contains(data_key);
+  }
```
kzvieriev, line 304: Looks misleading to me. This method should check is metadata cached only.
Also, I think, calling `cache_->Get()` after `cache_->Contains()` is an overhead.

andescu: We need to check if the data handle is also cached. Knowing that the metadata is cached does not bring anything.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -519,6 +519,18 @@ bool VersionedLayerClientImpl::RemoveFromCache(const geo::TileKey& tile) {
...
+  repository::PartitionsCacheRepository cache_repository(catalog_,
+                                                         settings_.cache);
+  return cache_repository.IsPartitionCached(catalog_version_.load(),
+                                            partition_id, layer_id_);
+}
+
+bool VersionedLayerClientImpl::IsCached(const geo::TileKey& tile) const {
+  auto partition_id = tile.ToHereTile();
+  return IsCached(partition_id);
+}
```
andescu, line 532: This is wrong. You need to use quad tree index for a tile, not a single partition entry from cache!

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -519,6 +519,18 @@ bool VersionedLayerClientImpl::RemoveFromCache(const geo::TileKey& tile) {
   return RemoveFromCache(partition_id);
```
andescu, line 519: After our switch to quad tree index this is also wrong and will probably never delete anything. Please fix!

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
kzvieriev: Yeah, but `PartitionsCacheRepository` isn't a best place to do it.
I would prefer to check data handle in `VersionedLayerClientImpl` instead.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
LiubovDidkivska: Will be fixed in separate PR

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
LiubovDidkivska: moved

olp-cpp-sdk-dataservice-read/tests/DataCacheRepositoryTest.cpp
```diff
@@ -66,6 +67,31 @@ TEST(PartitionsCacheRepositoryTest, DefaultExpiry) {
...
+    SCOPED_TRACE("Is not cached");
+
+    std::shared_ptr<cache::KeyValueCache> cache =
+        olp::client::OlpClientSettingsFactory::CreateDefaultCache({});
+    repository::DataCacheRepository repository(hrn, cache);
+
+    const auto result = repository.IsCached(layer, kDataHandle);
+
+    EXPECT_FALSE(result);
+  }
```
kzvieriev, line 94: these test cases are not related to DefaultExiry, lets separate them

PR #962, opened 2020-07-17
Fix implementation RemoveFromCache with TileKey

LiubovDidkivska:
Search for tile in quadtree. If data handle was found,
remove data. Fix tests to work with quadtree.

Relates-To: OLPEDGE-2124

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -522,11 +522,11 @@ bool VersionedLayerClientImpl::RemoveFromCache(const geo::TileKey& tile) {
           catalog_, settings_, layer_id_, catalog_version_.load(), tile,
           cached_tree)) {
     auto data = cached_tree.Find(tile, false);
-    if (!data) {
-      return false;
+    if (data) {
+      repository::DataCacheRepository cache_repository(catalog_,
+                                                       settings_.cache);
+      return cache_repository.Clear(layer_id_, data->data_handle);
     }
-    repository::DataCacheRepository cache_repository(catalog_, settings_.cache);
-    return cache_repository.Clear(layer_id_, data->data_handle);
   }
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -517,8 +517,18 @@ bool VersionedLayerClientImpl::RemoveFromCache(
...
-  auto partition_id = tile.ToHereTile();
-  return RemoveFromCache(partition_id);
+  read::QuadTreeIndex cached_tree;
+  if (repository::PartitionsRepository::FindQuadTree(
+          catalog_, settings_, layer_id_, catalog_version_.load(), tile,
+          cached_tree)) {
+    auto data = cached_tree.Find(tile, false);
+    if (!data) {
+      return false;
+    }
```
kzvieriev, line 527: For that method false means some error occurred, e.g. cache is not initialized.
metadata not found in quad tree is 'ok' case, so return true here.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -517,8 +517,18 @@ bool VersionedLayerClientImpl::RemoveFromCache(
...
-  return RemoveFromCache(partition_id);
+  read::QuadTreeIndex cached_tree;
+  if (repository::PartitionsRepository::FindQuadTree(
+          catalog_, settings_, layer_id_, catalog_version_.load(), tile,
+          cached_tree)) {
+    auto data = cached_tree.Find(tile, false);
+    if (data) {
+      repository::DataCacheRepository cache_repository(catalog_,
+                                                       settings_.cache);
+      return cache_repository.Clear(layer_id_, data->data_handle);
```
andescu, line 528: Please keep consistent. Above we have data_repository, you use cache_repository. Please use same as above.

PR #967, opened 2020-07-20
Add api to check if partition/tile cached

LiubovDidkivska:
Add functionality to check if certain tile/partition is cached.

For partition we are checking if data handle is found in cache,
if  data handle exist, check if data for corresponding partition 
in cache. 

For tile key search corresponding QuadTreeIndex,
considering that depth for all cached QuadTreeIndex is 4.
If quad tree is found, which should contain our tile, read 
data handle for that tile. If data handle exist, check if data 
for corresponding tile is cached. 

Add integration tests for to check if partition and tile is cached.
Get partition and tile. Check if it is in cache using IsCached
function. Remove partition/tile with RemoveFromCache function and
check again that it is no longer cached.

Relates-To: OLPEDGE-2111

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
@@ -375,6 +375,24 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    */
   bool RemoveFromCache(const geo::TileKey& tile);
 
+  /**
+   * @brief Checks whether the partition is cached.
+   *
+   * @param partition_id The partition ID.
+   *
+   * @return True if the partition data is cached; false otherwise.
+   */
+  bool IsCached(const std::string& partition_id) const;
+
+  /**
+   * @brief Checks whether the tile is cached.
+   *
+   * @param tile The tile key.
+   *
+   * @return True if the tile data is cached; false otherwise.
+   */
+  bool IsCached(const geo::TileKey& tile) const;
+
  private:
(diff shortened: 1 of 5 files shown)


olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
```diff
@@ -375,6 +375,24 @@ class DATASERVICE_READ_API VersionedLayerClient final {
...
+   */
+  bool IsCached(const std::string& partition_id) const;
+
+  /**
+   * @brief Checks if tile is cached.
+   *
+   * @param tile The tile key that we check.
+   *
+   * @return True if tile data is cached; false otherwise.
+   */
```
HalynaDumych, line 393:
```suggestion
  /**
   * @brief Checks whether the tile is cached.
   *
   * @param tile The tile key.
   *
   * @return True if the tile data is cached; false otherwise.
   */
```

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
```diff
@@ -375,6 +375,24 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    */
   bool RemoveFromCache(const geo::TileKey& tile);
 
+  /**
+   * @brief Checks whether the partition is cached.
+   *
+   * @param partition_id The partition ID.
+   *
+   * @return True if partition data is cached; false otherwise.
```
HalynaDumych, line 383:
```suggestion
   * @return True if the partition data is cached; false otherwise.
```

PR #969, opened 2020-07-21
Fix the issue with CMake paths

mykhailo-kuchma:
Fix the issue with CMake, when configuring the SDK in the root folder
that contain whitespaces in the path.

Resolves: OLPEDGE-527


mykhailo-kuchma:
@andescu Added

PR #971, opened 2020-07-21
Fix failed integration test

meravingen123:
* Fix olp::authentication::Permission class which was missing
  default initializer for decision field.
* Change kAuthorizeResponseValid to contain 'decision' field
  in Permission array instead of 'effect' field that was causing
  integration test AuthenticationClientTest.Authorize to fail.

Resolves: OLPEDGE-2190

### DIFF
--- a/tests/integration/olp-cpp-sdk-authentication/AuthenticationMockedResponses.h
+++ b/tests/integration/olp-cpp-sdk-authentication/AuthenticationMockedResponses.h
@@ -112,7 +112,7 @@ const std::string kInvalidAccessTokenResponse = R"JSON(
     {"errorId":"ERROR-cf976ca6-bf6e-44f7-a9e6-e271766c61fe","httpStatus":401,"errorCode":400601,"message":"Invalid accessToken.","error":"invalid_request","error_description":"errorCode: '400601'. Invalid accessToken."}
     )JSON";
 const std::string kAuthorizeResponseValid = R"JSON(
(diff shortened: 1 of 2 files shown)


tests/integration/olp-cpp-sdk-authentication/AuthenticationMockedResponses.h
```diff
@@ -112,7 +112,7 @@ const std::string kInvalidAccessTokenResponse = R"JSON(
     {"errorId":"ERROR-cf976ca6-bf6e-44f7-a9e6-e271766c61fe","httpStatus":401,"errorCode":400601,"message":"Invalid accessToken.","error":"invalid_request","error_description":"errorCode: '400601'. Invalid accessToken."}
     )JSON";
 const std::string kAuthorizeResponseValid = R"JSON(
-    {"identity":{"userId":"some_id","countryCode":"USA","emailVerified":false,"realm":"HERE"},"decision":"allow","diagnostics":[{"decision":"allow","permissions":[{"effect":"allow","action":"read","resource":"some_resource"}]}]}
+    {"identity":{"userId":"some_id","countryCode":"USA","emailVerified":false,"realm":"HERE"},"decision":"allow","diagnostics":[{"decision":"allow","permissions":[{"decision":"allow","action":"read","resource":"some_resource"}]}]}
```
LiubovDidkivska, line 115: I think we should add "effect" keyword for parsing, as it is mentioned in documentation.

PR #974, opened 2020-07-22
Add Protect tile keys functionality

LiubovDidkivska:
Protect tile keys from eviction and expiration.
If some tile key was marked as protected, it is
expected to stay in cache, that its data can be
accessible.

For protecting tile keys we should protect some
keys in cache. This means find quad tree which
contains data hanle for our key, protect this
quad tree and protect data with found data handle.
At least one tile key needs to be protected to
protect entire quad tree.

Added unit test with changed default expiration,
which requests 2 tiles within one quad tree and
caches it. After calls protect on tile keys,
3 keys should be protected, this means they do
not expire after timeout.

Relates-To: OLPEDGE-2126

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -20,6 +20,7 @@
 #include "VersionedLayerClientImpl.h"
 
 #include <string>
+#include <vector>
 
 #include <olp/core/cache/DefaultCache.h>
 #include <olp/core/client/OlpClientSettingsFactory.h>
@@ -662,6 +663,57 @@ VersionedLayerClientImpl::GetAggregatedData(TileRequest request) {
   return client::CancellableFuture<AggregatedDataResponse>(
       std::move(cancel_token), std::move(promise));
 }
+
+bool VersionedLayerClientImpl::Protect(const TileKeys& tiles) {
+  // could protect keys, which is already in cache
+  // because we don't know quad tree which will contain tiles
+  auto version = catalog_version_.load();
+  repository::DataCacheRepository data_cache_repository(catalog_,
+                                                        settings_.cache);
+  repository::PartitionsCacheRepository partitions_cache_repository(
+      catalog_, settings_.cache);
+
+  cache::KeyValueCache::KeyListType keys_to_protect;
+  keys_to_protect.reserve(tiles.size() * 2);
+
+  std::vector<read::QuadTreeIndex> quad_trees;
+  quad_trees.reserve(tiles.size());
+  for (const auto& tile : tiles) {
+    auto it =
+        std::find_if(quad_trees.begin(), quad_trees.end(),
+                     [&tile](const read::QuadTreeIndex& quad) {
+                       auto root = quad.GetRootTile();
+                       return (tile.Level() >= root.Level() &&
+                               tile.Level() - root.Level() <= kQuadTreeDepth &&
+                               root.IsParentOf(tile));
+                     });
+    if (it == quad_trees.end()) {
+      quad_trees.emplace_back();
+      read::QuadTreeIndex& cached_tree = quad_trees.back();
+      it = (quad_trees.rbegin() + 1).base();
+      if (!repository::PartitionsRepository::FindQuadTree(
+              catalog_, settings_, layer_id_, version, tile, cached_tree)) {
+        return false;
+      }
+    }
+
+    auto data = it->Find(tile, false);
+    if (!data) {
+      return false;
+    }
+    // add data handle
+    keys_to_protect.emplace_back(
+        data_cache_repository.CreateKey(layer_id_, data->data_handle));
+  }
+
+  // need to protect all quad trees
+  for (const auto& quad_tree : quad_trees) {
+    keys_to_protect.emplace_back(partitions_cache_repository.CreateQuadKey(
+        layer_id_, quad_tree.GetRootTile(), kQuadTreeDepth, version));
+  }
+  return settings_.cache->Protect(keys_to_protect);
+}
+
 PORTING_POP_WARNINGS()
 }  // namespace read
 }  // namespace dataservice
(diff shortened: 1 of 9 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -662,6 +663,68 @@ VersionedLayerClientImpl::GetAggregatedData(TileRequest request) {
...
+    std::vector<read::QuadTreeIndex> quad_trees;
+    quad_trees.reserve(tiles.size());
+    for (const auto& tile : tiles) {
+      auto it = std::find_if(
+          quad_trees.begin(), quad_trees.end(),
+          [&tile](const read::QuadTreeIndex& quad) {
+            auto root = quad.GetRootTile();
+            return (tile.Level() >= root.Level() &&
+                    tile.Level() - root.Level() <= kQuadTreeDepth &&
+                    root.IsParentOf(tile));
```
andescu, line 691: How about extending the QuadTreeIndex with a `bool Contains(TileKey)` method instead of "assuming" the tile is included. 
If it is enough to just look if the quadtree root is your parent, then I would instead just look if the quadtree is cached and work with string keys instead of QuadTreeIndex in your `quad_trees` vector. And then each time you have a new tile just check if one of its parent's key is already present in `quad_trees`.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -662,6 +663,68 @@ VersionedLayerClientImpl::GetAggregatedData(TileRequest request) {
...
+          });
+      if (it == quad_trees.end()) {
+        quad_trees.emplace_back();
+        read::QuadTreeIndex& cached_tree = quad_trees.back();
```
andescu, line 695: Better move instead of creating an empty element and never removing it when there is no such quad tree.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -662,6 +663,68 @@ VersionedLayerClientImpl::GetAggregatedData(TileRequest request) {
...
+        it = (quad_trees.rbegin() + 1).base();
+        if (!repository::PartitionsRepository::FindQuadTree(
+                catalog_, settings_, layer_id_, version, tile, cached_tree)) {
+          return false;
```
andescu, line 699: Why are you returning false here? Does this mean that if we do not find one of the requested tiles in cache we will abort directly?

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -662,6 +663,68 @@ VersionedLayerClientImpl::GetAggregatedData(TileRequest request) {
...
+    }
+    if (context.IsCancelled()) {
+      return {{static_cast<int>(olp::http::ErrorCode::CANCELLED_ERROR),
+               "Cancelled"}};
+    }
+    return settings_.cache->Protect(keys_to_protect);
+  };
+
+  return AddTask(settings_.task_scheduler, pending_requests_,
+                 std::move(protect_task), std::move(callback));
```
andescu, line 725: How much does this operation take? Do we really need async?

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
LiubovDidkivska: This what is actually checked if root lile in quad tree is parent. Should I move it to QuadTreeIndex, that it could be used in other cases?

andescu: I dont think it is enough to just check if it is parent as there are layers without certain data, e.g. only level 12 tiles, or only level 14 tiles, etc. If you check there that a level 14 tile is a child of a level 10 root but you will never have these tiles downloaded, then what is the point? You should check that the tile to be protected is actually inside the quad tree index and for this i would propose to add a new method in QuadTreeIndex.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
LiubovDidkivska: changed to sync

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -662,6 +664,64 @@ VersionedLayerClientImpl::GetAggregatedData(TileRequest request) {
...
+        quad_trees.emplace_back(std::move(cached_tree));
+      }
+    }
+  }
+
+  // need to protect all quad trees
+  for (const auto& quad_tree : quad_trees) {
+    keys_to_protect.emplace_back(partitions_cache_repository.CreateQuadKey(
+        layer_id_, quad_tree.GetRootTile(), kQuadTreeDepth, version));
+  }
```
andescu, line 718: Why don't you do this directly when you emplace to quad_trees? Why iterate again trough the vector?

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -662,6 +664,64 @@ VersionedLayerClientImpl::GetAggregatedData(TileRequest request) {
...
+  if (keys_to_protect.empty()) {
+    return false;
+  }
+  return settings_.cache->Protect(keys_to_protect);
```
andescu, line 722: I would check that the cache is not nullptr before and abort.
We kind of take it always for granted that the user provides a cache.

LiubovDidkivska: In VersionedLayerClientImpl we have check, but check added. Maybe need to do it in other places.
 ` if (!settings_.cache) {
    settings_.cache = client::OlpClientSettingsFactory::CreateDefaultCache({});
  }`

andescu: Only here for now is ok. We have a ticket for other cases.

PR #976, opened 2020-07-23
Fix boost windows build

meravingen123:
Fix boost build on windows by placing CMAKE_COMMAND in quotas
to prevent wrong interpretation of spaces in path.

Resolves: OLPSUP-11126


ystefinko:
Before merging please explain why this issue was not caught on CI build which we have.

PR #977, opened 2020-07-23
Adapt dataservice::read to ApiLookupClient.

kzvieriev:
We have new ApiLookupClient class in the core for Api requests. It
should replace internal ApiClientLookup for both read and write libs.
New lookup shared between repositories, so only one OlpClient used for
api requests, this will help to merge requests for same client.

Resolves: OLPEDGE-1697

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/CatalogRepository.h
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/CatalogRepository.h
@@ -20,6 +20,7 @@
 #pragma once
 
 #include <olp/core/cache/KeyValueCache.h>
+#include <olp/core/client/ApiLookupClient.h>
 #include <olp/core/client/CancellationContext.h>
 #include <olp/core/client/CancellationToken.h>
 #include <olp/core/client/HRN.h>
@@ -38,7 +39,8 @@ namespace repository {
 
 class CatalogRepository final {
  public:
-  CatalogRepository(client::HRN catalog, client::OlpClientSettings settings);
+  CatalogRepository(client::HRN catalog, client::OlpClientSettings settings,
+                    client::ApiLookupClient client);
 
   CatalogResponse GetCatalog(const CatalogRequest& request,
                              client::CancellationContext context);
@@ -52,6 +54,7 @@ class CatalogRepository final {
  private:
   client::HRN catalog_;
   client::OlpClientSettings settings_;
+  client::ApiLookupClient lookup_client_;
 };
 
 }  // namespace repository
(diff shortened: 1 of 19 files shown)


olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
```diff
@@ -862,6 +862,8 @@ TEST(VolatileLayerClientImplTest, PrefetchTilesCancellableFutureCancel) {
   auto cancellable = client.PrefetchTiles(
       read::PrefetchTilesRequest().WithTileKeys(tile_keys));
 
+  std::this_thread::sleep_for(std::chrono::milliseconds(300));
+
```
kzvieriev, line 866: oops, will remove that

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/FetchOptions.h
```diff
@@ -19,6 +19,8 @@
 
 #pragma once
 
+#include <olp/core/client/FetchOptions.h>
```
andescu, line 22: Why is this needed?

olp-cpp-sdk-dataservice-read/src/repositories/CatalogRepository.h
```diff
@@ -38,20 +39,21 @@ namespace repository {
 
 class CatalogRepository final {
  public:
-  static CatalogResponse GetCatalog(client::HRN catalog,
+  static CatalogResponse GetCatalog(client::ApiLookupClient client,
```
andescu, line 42: I don't like this approach. Better move away from static methods and add CatalogClient constructor which takes in HRN and ApiLookupClient.
Same for all other repository classes.
On the long run we will have more and more parameters.

olp-cpp-sdk-core/src/client/ApiLookupClient.cpp
```diff
@@ -28,6 +28,15 @@ ApiLookupClient::ApiLookupClient(const HRN& catalog,
...
 
+ApiLookupClient::ApiLookupClient(const ApiLookupClient& other) = default;
+
+ApiLookupClient& ApiLookupClient::operator=(const ApiLookupClient& other) =
+    default;
+
+ApiLookupClient::ApiLookupClient(ApiLookupClient&& other) = default;
+
+ApiLookupClient& ApiLookupClient::operator=(ApiLookupClient&& other) = default;
+
```
andescu, line 39: Remove `other` else we might face unused variable issues.

PR #980, opened 2020-07-23
Change Prefetch default levels

LiubovDidkivska:
Prefetch works with input parameters min and max levels. Defaults 
levels should indicate that user wants to get only tiles in request. 
As level 0 is valid, min and max values for request is initialized as 
invalid, that during reques we can distinguish if levels was set by 
user. Fix tests to run with default initialized levels instead of 0/0.

Relates-To: OLPEDGE-2198

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PrefetchTilesRequest.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PrefetchTilesRequest.h
@@ -203,8 +203,8 @@ class DATASERVICE_READ_API PrefetchTilesRequest final {
  private:
   std::string layer_id_;
   std::vector<geo::TileKey> tile_keys_;
-  unsigned int min_level_{0};
-  unsigned int max_level_{0};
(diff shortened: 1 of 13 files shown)


olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PrefetchTilesRequest.h
```diff
@@ -203,8 +203,8 @@ class DATASERVICE_READ_API PrefetchTilesRequest final {
  private:
   std::string layer_id_;
   std::vector<geo::TileKey> tile_keys_;
-  unsigned int min_level_{0};
-  unsigned int max_level_{0};
+  unsigned int min_level_{geo::TileKey().Level()};
```
andescu, line 206: Is it not better to add a constexpr or a alias to TileKey for this. e.g. kInvalidLevel = LevelCount.

LiubovDidkivska: changed to existing geo::TileKey::LevelCount

PR #983, opened 2020-07-27
Remove static from PartitionsRepository methods.

kzvieriev:
PartitionsRepository contains only static methods and they all are taking
HRN and OlpClientSettings as args. Moving these args to constructor
will simplify further extending, e.g. with ApiClientLookup.

Resolves: OLPEDGE-2200

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
@@ -106,26 +106,25 @@ namespace dataservice {
 namespace read {
 namespace repository {
 
+PartitionsRepository::PartitionsRepository(client::HRN catalog,
+                                           client::OlpClientSettings settings)
+    : catalog_(std::move(catalog)), settings_(std::move(settings)) {}
+
 PartitionsResponse PartitionsRepository::GetVersionedPartitions(
-    client::HRN catalog, std::string layer, int64_t version,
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -154,75 +153,78 @@ PartitionsResponse PartitionsRepository::GetVolatilePartitions(
...
-  auto fetch_option = request.GetFetchOption();
-  const auto key = request.CreateKey(layer);
+    std::string layer, const PartitionsRequest& request,
+    client::CancellationContext context, boost::optional<std::int64_t> version,
+    boost::optional<time_t> expiry) {
+  // PartitionsCacheRepository still using deprecated
+  // PartitionsRequest::GetVersion, so remove `request_copy` when the repository
+  // supports version.
+  auto request_copy = request;
+  request_copy.WithVersion(version);
```
kzvieriev, line 168: As mentioned in the comment, `PartitionsCacheRepository` has no version arg yet.
Adding it is a part of another task(Removing `WithVersion()` from `PartitionRequest`), which will be done in the nearest feature.
We are splitting changes, because some files(e.g. PartitionsRepositoryTest.cpp) are changed a lot in both tasks.
The PR makes `PartitionsRepository` support version arg, so we don't have to do similar changes twice.

kzvieriev: removed

PR #986, opened 2020-07-27
Release tile keys from protected list.

LiubovDidkivska:
Release keys, which was marked before as protected. After releasing tile
keys associated data in cache could expire or be evicted.

For releasing tile keys we should release quad tree and tile data.
This means find quad tree which contains data hanle for our key,
release this quad tree only in case if no other protected tiles depend
on this quad tree. For each tile release data under data handle.

Added unit test with releasing 2 tiles keys under common quad tree,
which has 3 protected tile keys, so quad tree associated key should not
be released. Other test tries to release last tile key for quad tree,
so 2 keys should be released. Test which tries to relese not protected
key.

Relates-To: OLPEDGE-2127

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -805,6 +805,11 @@ bool DefaultCacheImpl::Release(const DefaultCache::KeyListType& keys) {
   return result;
 }
 
+bool DefaultCacheImpl::IsProtected(const std::string& key) {
+    std::lock_guard<std::mutex> lock(cache_lock_);
+  return protected_keys_.IsProtected(key);
+}
+
 time_t DefaultCacheImpl::GetExpiryForMemoryCache(const std::string& key,
                                                  const time_t& expiry) {
   // reset expiry if key is protected for memory cache
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -144,6 +144,7 @@ void DefaultCacheImpl::Close() {
   mutable_cache_.reset();
   mutable_cache_lru_.reset();
   protected_cache_.reset();
+  protected_keys_.Clear();
   mutable_cache_data_size_ = 0;
   is_open_ = false;
 }
@@ -627,6 +628,7 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::SetupStorage() {
   mutable_cache_.reset();
   mutable_cache_lru_.reset();
   protected_cache_.reset();
+  protected_keys_.Clear();
   mutable_cache_data_size_ = 0;
 
   if (settings_.max_memory_cache_size > 0) {
@@ -646,6 +648,7 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::SetupStorage() {
     auto status = mutable_cache_->Open(settings_.disk_path_mutable.get(),
                                        settings_.disk_path_mutable.get(),
                                        storage_settings, OpenOptions::Default);
+    OLP_SDK_LOG_WARNING(kLogTag, settings_.disk_path_mutable.get().c_str());
     if (status == OpenResult::Fail) {
       OLP_SDK_LOG_ERROR_F(kLogTag, "Failed to open the mutable cache %s",
                           settings_.disk_path_mutable.get().c_str());
@@ -751,12 +754,12 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
 bool DefaultCacheImpl::Protect(const DefaultCache::KeyListType& keys) {
   std::lock_guard<std::mutex> lock(cache_lock_);
   auto start = std::chrono::steady_clock::now();
-  protected_keys_.Protect(keys, [&](const std::string& key) {
+  auto result = protected_keys_.Protect(keys, [&](const std::string& key) {
     if (!RemoveKeyLru(key)) {
       RemoveKeysWithPrefixLru(key);
     }
   });
-  if (memory_cache_) {
+  if (memory_cache_ && result) {
     memory_cache_->Clear();
   }
 
@@ -766,13 +769,17 @@ bool DefaultCacheImpl::Protect(const DefaultCache::KeyListType& keys) {
                      GetElapsedTime(start),
                      static_cast<std::uint64_t>(keys.size()),
                      protected_keys_.Count());
-  return true;
+  return result;
 }
 
 bool DefaultCacheImpl::Release(const DefaultCache::KeyListType& keys) {
   std::lock_guard<std::mutex> lock(cache_lock_);
   auto start = std::chrono::steady_clock::now();
   auto result = protected_keys_.Release(keys);
+
+  if (!result) {
+    return result;
+  }
 
   for (const auto& key : keys) {
     if (memory_cache_) {
@@ -805,8 +812,8 @@ bool DefaultCacheImpl::Release(const DefaultCache::KeyListType& keys) {
   return result;
 }
 
-bool DefaultCacheImpl::IsProtected(const std::string& key) {
-    std::lock_guard<std::mutex> lock(cache_lock_);
+bool DefaultCacheImpl::IsProtected(const std::string& key) const {
+  std::lock_guard<std::mutex> lock(cache_lock_);
   return protected_keys_.IsProtected(key);
 }
 
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -144,7 +144,7 @@ void DefaultCacheImpl::Close() {
   mutable_cache_.reset();
   mutable_cache_lru_.reset();
   protected_cache_.reset();
-  protected_keys_.Clear();
+  protected_keys_ = ProtectedKeyList();
   mutable_cache_data_size_ = 0;
   is_open_ = false;
 }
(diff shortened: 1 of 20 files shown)


olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -751,12 +754,12 @@ std::string DefaultCacheImpl::GetExpiryKey(const std::string& key) const {
 bool DefaultCacheImpl::Protect(const DefaultCache::KeyListType& keys) {
   std::lock_guard<std::mutex> lock(cache_lock_);
   auto start = std::chrono::steady_clock::now();
-  protected_keys_.Protect(keys, [&](const std::string& key) {
+  auto result = protected_keys_.Protect(keys, [&](const std::string& key) {
     if (!RemoveKeyLru(key)) {
```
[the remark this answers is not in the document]
LiubovDidkivska: this is lambda, which is called in ProtectedKeyList::Protect

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -654,6 +654,106 @@ bool VersionedLayerClientImpl::Protect(const TileKeys& tiles) {
...
+
+  cache::KeyValueCache::KeyListType keys_to_release;
+
+  std::map<geo::TileKey, std::map<geo::TileKey, std::string>>
+      quad_trees_with_protected_tiles;
+
+  auto process_tile = [&](const geo::TileKey& tile_key_to_release) {
+    auto max_depth =
+        std::min<std::uint32_t>(tile_key_to_release.Level(), kQuadTreeDepth);
+    for (int i = max_depth; i >= 0; --i) {
```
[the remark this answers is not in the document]
LiubovDidkivska: max_depth changed to signed, due to check >=0

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.cpp
```diff
@@ -0,0 +1,152 @@
...
+    : layer_id_(layer_id),
+      version_(version),
+      cache_(settings.cache),
+      data_cache_repository_(catalog, settings.cache),
+      partitions_cache_repository_(catalog, settings.cache),
+      repository_(catalog, settings),
+      quad_trees_with_protected_tiles_(),
+      keys_to_release_() {}
+
+cache::KeyValueCache::KeyListType&&
```
andescu, line 57: Why do you return a rvalue when you can simply return the cache::KeyValueCache::KeyListType which is the same except you are not forced to use std::move.
Also the name of the method will get much shorter: `GetKeysToRelease`

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.cpp
```diff
@@ -0,0 +1,152 @@
...
+ReleaseDependencyResolver::GenerateAndMoveKeysToRelease(const TileKeys& tiles) {
+  for (const auto& tile : tiles) {
+    if (!ProcessTileKeyInMap(tile)) {
+      ProcessTileKeyInCache(tile);
+    }
+  }
+  return std::move(keys_to_release_);
+}
+
+bool ReleaseDependencyResolver::ProcessTileKeyInMap(
```
andescu, line 67:
```suggestion
bool ReleaseDependencyResolver::ProcessTileKey(
```

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.cpp
```diff
@@ -0,0 +1,152 @@
...
+        quad_trees_with_protected_tiles_.erase(it);
+      }
+    }
+    return true;
+  }
+  return false;
+}
+
+ReleaseDependencyResolver::MapOfQuads::iterator
+ReleaseDependencyResolver::FindQuadInMap(const geo::TileKey& tile_key) {
```
andescu, line 95:
```suggestion
ReleaseDependencyResolver::FindQuad(const geo::TileKey& tile_key) {
```

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.cpp
```diff
@@ -0,0 +1,152 @@
...
+namespace {
+constexpr auto kQuadTreeDepth = 4;
+}  // namespace
+
+namespace olp {
+namespace dataservice {
+namespace read {
+
+cache::KeyValueCache::KeyListType
+ReleaseDependencyResolver::GenerateKeysToRelease(
```
andescu, line 36:
```suggestion
ReleaseDependencyResolver::GetKeysToRelease(
```

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.h
```diff
@@ -0,0 +1,81 @@
...
+#include "repositories/DataCacheRepository.h"
+#include "repositories/DataRepository.h"
+#include "repositories/PartitionsCacheRepository.h"
+#include "repositories/PartitionsRepository.h"
+
+namespace olp {
+namespace dataservice {
+namespace read {
+
+class ReleaseDependencyResolver {
```
andescu, line 38: Consider also using this for Protect().
Means the class name should be ProtectionDependencyResolver.

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.h
```diff
@@ -0,0 +1,81 @@
...
+ public:
+  // group quad trees and protected tiles. Check if for each quad tree we want
+  // to release all protected keys associated, if true add this quad tree to
+  // released keys
+  static cache::KeyValueCache::KeyListType GenerateKeysToRelease(
```
andescu, line 43: This class setup is very strange. A static method that creates a temporary instance of itself to generate some data and copies internally not very cheap data. If it is only temporary class then do not copy but use references and remove the static from this method.

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.h
```diff
@@ -0,0 +1,81 @@
...
+      const client::HRN& catalog, const std::string& layer_id,
+      const int64_t& version, const client::OlpClientSettings& settings,
+      const TileKeys& tiles);
+
+ private:
+  using MapOfTileDataKeys = std::map<geo::TileKey, std::string>;
```
andescu, line 49: Aliases are usually suffixed by "Type", e.g. `DataKeysType`
As there is no difference between partition data or tile data key, both use data handles.

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.h
```diff
@@ -0,0 +1,81 @@
...
+  using MapOfQuads = std::map<geo::TileKey, MapOfTileDataKeys>;
```
andescu, line 50:
```suggestion
  using QuadsType = std::map<geo::TileKey, MapOfTileDataKeys>;
```

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.h
LiubovDidkivska: Static method removed

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.h
```diff
@@ -0,0 +1,77 @@
...
+  QuadsType::iterator FindQuad(const geo::TileKey& tile_key);
+
+  TilesDataKeysType ProcessQuad(const read::QuadTreeIndex& cached_tree,
+                                const geo::TileKey& tile);
+
+  void ProcessTileKeyInCache(const geo::TileKey& tile);
+
+ private:
+  std::string layer_id_;
+  int64_t version_;
```
andescu, line 66:
```suggestion
  const std::string& layer_id_;
  const int64_t version_;
```

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.h
```diff
@@ -0,0 +1,77 @@
...
+#include "repositories/PartitionsCacheRepository.h"
+#include "repositories/PartitionsRepository.h"
+
+namespace olp {
+namespace dataservice {
+namespace read {
+
+// group quad trees and protected tiles. Check if for each quad tree we want
+// to release all protected keys associated, if true add this quad tree to
+// released keys
```
andescu, line 40: This is not a proper comment. Use `///` for simple Doxygen comments.

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.cpp
```diff
@@ -420,14 +420,15 @@ PartitionResponse PartitionsRepository::GetTile(
 }
 PORTING_POP_WARNINGS()
 
-bool PartitionsRepository::FindQuadTree(
-    const std::string& layer, boost::optional<int64_t> version,
-    const olp::geo::TileKey& tile_key, read::QuadTreeIndex& tree) {
+bool PartitionsRepository::FindQuadTree(const std::string& layer,
+                                        boost::optional<int64_t> version,
+                                        const olp::geo::TileKey& tile_key,
+                                        read::QuadTreeIndex& tree) {
```
andescu, line 426: Why is this again formatted differently? I don't see anything's been changed.

olp-cpp-sdk-dataservice-read/src/ReleaseDependencyResolver.h
```diff
@@ -0,0 +1,77 @@
...
+
+/// group quad trees and protected tiles. Check if for each quad tree we want
+/// to release all protected keys associated, if true add this quad tree to
+/// released keys
+class ReleaseDependencyResolver {
+ public:
+  ReleaseDependencyResolver(const client::HRN& catalog,
+                            const std::string& layer_id, const int64_t& version,
+                            const client::OlpClientSettings& settings,
+                            const client::ApiLookupClient& lookup_client);
```
andescu, line 46: It seems to me like over-engineered to take in ApiLookupClient to just create a PartitionsRepository to just have a helper method for accessing the cache. Please ask yourself if the FindQuadTree really belongs in PartitionsRepositoy and not in PartitionsCacheRepository. I would place it in PartitionsCacheRepository.

PR #987, opened 2020-07-28
Remove deprecated API

meravingen123:
Removed deprecated method WithVersion from
dataservice read ParttitionsRequest

Resolves: OLPEDGE-2188

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PartitionsRequest.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PartitionsRequest.h
@@ -21,8 +21,8 @@
 
 #include <sstream>
 #include <string>
-#include <vector>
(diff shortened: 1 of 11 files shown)


olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/PartitionsRequest.h
```diff
@@ -218,11 +186,11 @@ class DATASERVICE_READ_API PartitionsRequest final {
    *
    * @return A string representation of the request.
    */
-  std::string CreateKey(const std::string& layer_id) const {
+  std::string CreateKey(const std::string& layer_id, boost::optional<int64_t> version) const {
```
andescu, line 189: This is an API break. Please provide a default value for the second parameter. Then the users are not forced to change their code.

PR #990, opened 2020-07-29
Fix Prefetch volatile test

LiubovDidkivska:
When requesting tiles we should get tiles
only on input levels, if default levels not set.
Filter quads to be on requested levels, and
have relation parent/child with root tiles from
request.

Relates-To: OLPEDGE-2204

### DIFF
--- a/olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
+++ b/olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
@@ -490,8 +490,6 @@ TEST(VolatileLayerClientImplTest, PrefetchTiles) {
                             olp::http::HttpStatusCode::OK);
     SetupNetworkExpectation(*network_mock, kUrlPrefetchBlobData1, kData1,
                             olp::http::HttpStatusCode::OK);
-    SetupNetworkExpectation(*network_mock, kUrlPrefetchBlobData2, kData2,
-                            olp::http::HttpStatusCode::OK);
 
     EXPECT_CALL(*network_mock, Send(IsGetRequest(kUrlLookup), _, _, _, _))
         .WillRepeatedly(
@@ -554,8 +552,7 @@ TEST(VolatileLayerClientImplTest, PrefetchTiles) {
 
     ASSERT_NE(future.wait_for(kTimeout), std::future_status::timeout);
     read::PrefetchTilesResponse response = future.get();
-    ASSERT_TRUE(response.IsSuccessful()) << response.GetError().GetMessage();
-    ASSERT_TRUE(response.GetResult().empty());
+    ASSERT_FALSE(response.IsSuccessful()) << response.GetError().GetMessage();
   }
   {
     SCOPED_TRACE("Levels not specified.");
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
```diff
@@ -554,8 +548,7 @@ TEST(VolatileLayerClientImplTest, PrefetchTiles) {
 
     ASSERT_NE(future.wait_for(kTimeout), std::future_status::timeout);
     read::PrefetchTilesResponse response = future.get();
-    ASSERT_TRUE(response.IsSuccessful()) << response.GetError().GetMessage();
-    ASSERT_TRUE(response.GetResult().empty());
+    ASSERT_FALSE(response.IsSuccessful()) << response.GetError().GetMessage();
```
kzvieriev, line 551:
```suggestion
    ASSERT_FALSE(response.IsSuccessful());
```
Check will fail if response is successful. In this case error message is always empty.

Btw, why response should be failed? Min/max levels are default, so only tiles from request loaded.
If behavior for default levels changed, we must update doxygen too :-)

olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
```diff
@@ -585,8 +578,7 @@ TEST(VolatileLayerClientImplTest, PrefetchTiles) {
...
```
kzvieriev, line 581:
```suggestion
    ASSERT_FALSE(response.IsSuccessful());
```

olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
```diff
@@ -554,8 +548,7 @@ TEST(VolatileLayerClientImplTest, PrefetchTiles) {
...
   }
   {
     SCOPED_TRACE("Levels not specified.");
```
kzvieriev, line 554: Are there any difference between "Prefetch tiles with default levels" and "Levels not specified." tests? Looks identical for me.

olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
```diff
@@ -723,8 +713,7 @@ TEST(VolatileLayerClientImplTest, PrefetchTilesCancellableFuture) {
 
     ASSERT_NE(future.wait_for(kTimeout), std::future_status::timeout);
     read::PrefetchTilesResponse response = future.get();
-    ASSERT_TRUE(response.IsSuccessful()) << response.GetError().GetMessage();
-    ASSERT_TRUE(response.GetResult().empty());
+    ASSERT_FALSE(response.IsSuccessful()) << response.GetError().GetMessage();
```
kzvieriev, line 716:
```suggestion
    ASSERT_FALSE(response.IsSuccessful());
```

olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
```diff
@@ -748,8 +737,7 @@ TEST(VolatileLayerClientImplTest, PrefetchTilesCancellableFuture) {
...
```
kzvieriev, line 740:
```suggestion
    ASSERT_FALSE(response.IsSuccessful());
```

olp-cpp-sdk-dataservice-read/tests/VolatileLayerClientImplTest.cpp
LiubovDidkivska: There is no tiles to load, so request fails if it is empty. Now behavior the same if we received empty quad, or request had wrong root tile, so there is no children or parent on requested levels.

kzvieriev: I think, we must fix it than. Purpose of the test is to load only tiles from request.
It makes no sense, if quad tree doesn't contain tiles.

PR #993, opened 2020-07-29
Adds PendingUrlRequests and PendingUrlRequest.

andescu:
Both classes will be used by OlpClient to merge same URL requests and
thus avoid requesting the same data multiple times. For this a change
is needed in HttpResponse to handle the std::stringstream copy on
copy constructor and copy assignment operator. Without this it is not
possible to copy an existing HttpResponse and return it to multiple
waiting callbacks. At the same time we deprecate the public members
of HttpResponse and prepare it to be changed by 01.2020 into a shared
pimpl class so that we do not need to deep copy the responses.
The PendingUrlRequest class manages one Network request, for which it
remebers the request id and is able to cancel it, and is able to hold
one or multiple response callback. It also handles correctly one, more
or all callbacks cancellation scenarios, incl. cancelling the network
request once all pending callbacks are cancelled.

Both PendingUrlRequests and PendingUrlRequest are mutex protected and
thread safe, additionaly PendingUrlRequests class takes care to not
get a conflict in case a URL request was cancelled fully and there is
a new request comming for the same URL. In this case a new request will
be triggered and the old one will be taken out from the pending request
and placed in a separate container for cacelled requests only.

Relates-To: OLPEDGE-1805

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
@@ -77,18 +77,21 @@ class NetworkStatistics {
 };
 
 /**
- * @brief An HTTP response.
+ * @brief This class represents the HTTP response created from the
+ * NetworkResponse and the request body.
  */
 class CORE_API HttpResponse {
  public:
   HttpResponse() = default;
+  virtual ~HttpResponse() = default;
+
   /**
    * @brief Creates the `HttpResponse` instance.
    *
    * @param status The HTTP status.
    * @param response The response body.
    */
-  HttpResponse(int status, std::string response = std::string())  // NOLINT
+  HttpResponse(int status, std::string response = {})  // NOLINT
       : status(status), response(std::move(response)) {}
 
   /**
@@ -96,6 +99,7 @@ class CORE_API HttpResponse {
    *
    * @param status The HTTP status.
    * @param response The response body.
+   *
    */
   HttpResponse(int status, std::stringstream&& response)
       : status(status), response(std::move(response)) {}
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-core/include/olp/core/client/HttpResponse.h
```diff
@@ -77,25 +77,29 @@ class NetworkStatistics {
...
   /**
    * @brief Creates the `HttpResponse` instance.
    *
    * @param status The HTTP status.
    * @param response The response body.
    */
-  HttpResponse(int status, std::string response = std::string())  // NOLINT
-      : status(status), response(std::move(response)) {}
+  HttpResponse(int status, std::string response = {})  // NOLINT
+      : status(status), response(response) {}
```
meravingen123, line 95: why remove std::move?

andescu: User error :)
Ill add it back.

olp-cpp-sdk-core/src/client/PendingUrlRequests.h
```diff
@@ -0,0 +1,147 @@
...
+
+  /// Get the Network request Id associated with this request.
+  http::RequestId GetRequestId() const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return http_request_id_;
+  }
+
+  /// Cancels the ongoing Network request. Do not block for long!
+  /// NOTE: This should only be called if you know what you are doing.
+  void Cancel() { context_.CancelOperation(); }
```
meravingen123, line 72: I would rather rename this to something different so there will be less confusion for the user between this and Cancel(callback_id)

olp-cpp-sdk-core/src/client/PendingUrlRequests.cpp
```diff
@@ -0,0 +1,254 @@
...
+  std::lock_guard<std::mutex> lock(mutex_);
+  const auto callback_id = callbacks_.size();
+  callbacks_.emplace(callback_id, std::move(callback));
+  return callback_id;
+}
+
+bool PendingUrlRequest::ExecuteOrCancelled(const ExecuteFuncType& func,
+                                           const CancelFuncType& cancel_func) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  return context_.ExecuteOrCancelled(
```
meravingen123, line 47: Why this mutex is needed? Context already has a mutex.

olp-cpp-sdk-core/src/client/PendingUrlRequests.cpp
```diff
@@ -0,0 +1,254 @@
...
+    auto cancelled_it = cancelled_requests_.find(url);
+
+    // Either the request is in the pending or in the cancelled requests list.
+    // Check that it is the same request. It might happen that we have a
+    // cancelled request and then a followup valid request with the same URL.
+    if (pending_it != pending_requests_.end() &&
+        request_id == pending_it->second->GetRequestId()) {
+      request_ptr = std::move(pending_it->second);
+      pending_requests_.erase(pending_it);
+    } else if (cancelled_it != cancelled_requests_.end() &&
```
meravingen123, line 238: May there be a situations where one url request has been cancelled and the second one with the same url is added?

olp-cpp-sdk-core/src/client/PendingUrlRequests.cpp
mykhailo-kuchma, line 47: context_ can't be reused multiple times.

olp-cpp-sdk-core/src/client/PendingUrlRequests.cpp
andescu: Yes. It might be that we one request cancels its callback but the Network takes some time to return the answer and there is another request from a different call for the same URL.

olp-cpp-sdk-core/src/client/PendingUrlRequests.h
andescu: Will do.

PR #994, opened 2020-07-29
Fix Prefetch empty responce

LiubovDidkivska:
When requesting tiles we should get success
even with empty vector of tiles result .
If default levels set return tile not found
for each tile from request.
Add tests to check empty tiles result.

Relates-To: OLPEDGE-2204


kzvieriev:
I think, it would be good to update prefetch doxygen with these new details also.

PR #1004, opened 2020-08-04
Adding CatalogEndpointProvider.

kzvieriev:
Api url can be static. In such cases we can avoid lookup request and
save some time. For that purpose we are adding CatalogEndpointProvider
to ApiLookupSettings.

ApiLookupClient will try to get url using the provider from settings
instead of making api lookup request. If it can't be done, behavior
of the client is unchanged.

Also adding unit and integration test for ApiLookupClient.

Resolves: OLPEDGE-2187

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/client/ApiLookupClient.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/ApiLookupClient.h
@@ -79,6 +79,12 @@ class CORE_API ApiLookupClient final {
    * @param callback The function callback used to receive the
    * `LookupApiResponse` instance.
    *
+   * @note If the catalog endpoint provider is set and provides a static URL
+   * for this catalog, the method does not trigger any asynchronous download and
+   * provides the synchronous result instead. This means that
+   * the user needs to take special care in case the callback is called within
+   * the same context and avoid locking any mutex twice.
+   *
    * @return The method used to call or to cancel the request.
    */
   CancellationToken LookupApi(const std::string& service,
--- a/olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
@@ -29,6 +29,7 @@
 #include <olp/core/client/BackdownStrategy.h>
 #include <olp/core/client/CancellationToken.h>
 #include <olp/core/client/DefaultLookupEndpointProvider.h>
+#include <olp/core/client/HRN.h>
 #include <olp/core/client/HttpResponse.h>
 #include <olp/core/http/Network.h>
 
@@ -249,6 +250,21 @@ struct CORE_API ApiLookupSettings {
   using LookupEndpointProvider = std::function<std::string(const std::string&)>;
 
   /**
+   * @brief The type alias of the catalog endpoint provider function.
+   *
+   * Catalogs that have a static URL or can be accessed through
+   * a proxy service can input the URL provider here. This URL provider is taken
+   * by the `ApiLookupClient` and returned directly to the caller without any
+   * requests to the API Lookup Service.
+   *
+   * @note This call should be synchronous without any tasks scheduled in
+   * `TaskScheduler` as it might result in a dead-lock.
+   *
+   * @return An empty string if the catalog is invalid or unknown.
+   */
+  using CatalogEndpointProvider = std::function<std::string(const HRN&)>;
+
+  /**
    * @brief The provider of endpoint for API lookup requests.
    *
    * The lookup API endpoint provider will be called prior to every API lookup
@@ -259,6 +275,18 @@ struct CORE_API ApiLookupSettings {
    */
   LookupEndpointProvider lookup_endpoint_provider =
       DefaultLookupEndpointProvider();
+
+  /**
+   * @brief The endpoint provider for API requests.
+   *
+   * If some of the catalogs have fixed URLs and do not need the API Lookup
+   * Service, you can provide the static URL via `CatalogEndpointProvider`.
+   * Every request will receive this URL from `ApiLookupClient` without
+   * any HTTP requests to the API Lookup Service. `CatalogEndpointProvider` is
+   * called before `lookup_endpoint_provider`, and if the output is not empty,
+   * `lookup_endpoint_provider` is not called additionally.
+   */
+  CatalogEndpointProvider catalog_endpoint_provider = nullptr;
 };
 
 /**
--- a/olp-cpp-sdk-core/src/client/ApiLookupClientImpl.cpp
+++ b/olp-cpp-sdk-core/src/client/ApiLookupClientImpl.cpp
@@ -42,6 +42,24 @@ std::string FindApi(const Apis& apis, const std::string& service,
   return it->GetBaseUrl();
 }
 
+olp::client::OlpClient GetStaticUrl(
+    const olp::client::HRN& catalog,
+    const olp::client::OlpClientSettings& settings) {
+  const auto& provider = settings.api_lookup_settings.catalog_endpoint_provider;
+  if (provider) {
+    auto url = provider(catalog);
+    if (!url.empty()) {
+      url += "/catalogs/" + catalog.ToCatalogHRNString();
+      OlpClient result_client;
+      result_client.SetBaseUrl(url);
+      result_client.SetSettings(settings);
+      return result_client;
+    }
+  }
+
+  return {};
+}
+
 }  // namespace
 
 ApiLookupClientImpl::ApiLookupClientImpl(const HRN& catalog,
@@ -56,6 +74,11 @@ ApiLookupClientImpl::ApiLookupClientImpl(const HRN& catalog,
 ApiLookupClient::LookupApiResponse ApiLookupClientImpl::LookupApi(
     const std::string& service, const std::string& service_version,
     FetchOptions options, CancellationContext context) {
+  auto result_client = GetStaticUrl(catalog_, settings_);
+  if (!result_client.GetBaseUrl().empty()) {
+    return result_client;
+  }
+
   repository::ApiCacheRepository repository(catalog_, settings_.cache);
   const auto hrn = catalog_.ToCatalogHRNString();
 
@@ -65,7 +88,6 @@ ApiLookupClient::LookupApiResponse ApiLookupClientImpl::LookupApi(
       OLP_SDK_LOG_DEBUG_F(kLogTag, "LookupApi(%s/%s) found in cache, hrn='%s'",
                           service.c_str(), service_version.c_str(),
                           hrn.c_str());
-      OlpClient result_client;
       result_client.SetBaseUrl(*url);
       result_client.SetSettings(settings_);
       return result_client;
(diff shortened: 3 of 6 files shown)


olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
```diff
@@ -248,6 +249,20 @@ struct CORE_API ApiLookupSettings {
...
 
+  /**
+   * @brief The type alias of the catalog endpoint provider function.
+   *
+   * Catalogs which already have an static URL or can be accessed through a
+   * proxy service can input the URL provider here and it will be taken by the
+   * ApiLookupClient and returned directly to the caller without any request to
+   * the Lookup API service.
+   *
+   * @note Return empty string in case of an invalid or unknown partition.
```
andescu, line 260:
```suggestion
   * @note Return empty string in case of an invalid or unknown catalog.
```

olp-cpp-sdk-core/src/client/ApiLookupClientImpl.cpp
```diff
@@ -125,6 +138,20 @@ ApiLookupClient::LookupApiResponse ApiLookupClientImpl::LookupApi(
...
+  const auto& provider =
+      settings_.api_lookup_settings.catalog_endpoint_provider;
+  if (provider) {
+    auto url = provider(catalog_);
+    if (!url.empty()) {
+      url += "/catalogs/" + catalog_.ToCatalogHRNString();
+      OlpClient result_client;
+      result_client.SetBaseUrl(url);
+      result_client.SetSettings(settings_);
+      callback(result_client);
```
andescu, line 150: Please make sure you write in the ApiLookupClientImpl::LookupApi() doxygen that the callback can be called sync in case the static URL is set. You can imagine the surprise on the users face when his mutex locked twice crashes.

olp-cpp-sdk-core/src/client/ApiLookupClientImpl.cpp
```diff
@@ -56,6 +56,19 @@ ApiLookupClientImpl::ApiLookupClientImpl(const HRN& catalog,
...
     FetchOptions options, CancellationContext context) {
+  const auto& provider =
+      settings_.api_lookup_settings.catalog_endpoint_provider;
+  if (provider) {
+    auto url = provider(catalog_);
+    if (!url.empty()) {
+      url += "/catalogs/" + catalog_.ToCatalogHRNString();
+      OlpClient result_client;
+      result_client.SetBaseUrl(url);
+      result_client.SetSettings(settings_);
```
andescu, line 67: This code block is repetitive. Please move to some common method.

olp-cpp-sdk-core/tests/client/ApiLookupClientImplTest.cpp
```diff
@@ -393,6 +393,69 @@ TEST_F(ApiLookupClientImplTest, LookupApiCustomProvider) {
   testing::Mock::VerifyAndClearExpectations(cache_.get());
 }
 
+TEST_F(ApiLookupClientImplTest, LookupApiCustomCatalogProvider) {
+  using testing::_;
+  using testing::Return;
```
andescu, line 398: How about moving this to the global space and avoid repetitive code?

olp-cpp-sdk-core/tests/client/ApiLookupClientImplTest.cpp
```diff
@@ -393,6 +393,69 @@ TEST_F(ApiLookupClientImplTest, LookupApiCustomProvider) {
...
+  const std::string service_name = "random_service";
+  const std::string service_url = "http://random_service.com";
+  const std::string service_version = "v8";
+  const std::string provider_url = "https://some-lookup-url.com/lookup/v1";
+  const std::string static_base_url = provider_url + "/catalogs/" + catalog;
+  const std::string lookup_url =
+      "https://api-lookup.data.api.platform.here.com/lookup/v1/resources/" +
+      catalog + "/apis";
+  {
+    SCOPED_TRACE("Static url catalog");
```
andescu, line 412: Please add a blank line after the SCOPED_TRACE.

olp-cpp-sdk-core/tests/client/ApiLookupClientImplTest.cpp
```diff
@@ -393,6 +393,69 @@ TEST_F(ApiLookupClientImplTest, LookupApiCustomProvider) {
   testing::Mock::VerifyAndClearExpectations(cache_.get());
 }
 
+TEST_F(ApiLookupClientImplTest, LookupApiCustomCatalogProvider) {
```
andescu, line 396: Please find a smaller name. ApiLookupClient only has two public methods and both are called `LookupApi` so I dont think you need to especially mention that.

olp-cpp-sdk-core/tests/client/ApiLookupClientImplTest.cpp
```diff
@@ -714,4 +777,79 @@ TEST_F(ApiLookupClientImplTest, LookupApiCustomProviderAsync) {
   testing::Mock::VerifyAndClearExpectations(cache_.get());
 }
 
+TEST_F(ApiLookupClientImplTest, LookupApiCustomCatalogProviderAsync) {
```
andescu, line 780: Same here.

kzvieriev:
> Please fix your commit message:
> 
> * Why are we adding this?
> * What is the purpose of this change?
> * What exactly did you add?
> 
> Additionally unit tests alone are not enough! Integration tests with VersionedLayerClient please!

added some details to the commit message and integration tests :-)

olp-cpp-sdk-core/include/olp/core/client/ApiLookupClient.h
```diff
@@ -78,6 +78,8 @@ class CORE_API ApiLookupClient final {
    * which data should be fetched.
    * @param callback The function callback used to receive the
    * `LookupApiResponse` instance.
+   * @note If catalog endpoint provider specified and has URL for the catalog,
+   * the method is sync, meaning callback will be called in the same thread.
```
HalynaDumych, line 82:
```suggestion
   *
   * @note If the catalog endpoint provider is specified and has the catalog URL,
   * the method is synchronized, and the callback is called in the same thread.
```

olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
```diff
@@ -248,6 +249,20 @@ struct CORE_API ApiLookupSettings {
...
+   *
+   * Catalogs which already have an static URL or can be accessed through a
+   * proxy service can input the URL provider here and it will be taken by the
+   * ApiLookupClient and returned directly to the caller without any request to
+   * the Lookup API service.
+   *
+   * @note Return empty string in case of an invalid or unknown catalog.
+   * @note This call should be synchronous without any tasks scheduled on the
+   * TaskScheduler as this might result in a dead-lock.
+   */
```
HalynaDumych, line 263:
```suggestion
  /**
   * @brief The type alias of the catalog endpoint provider function.
   *
   * Catalogs that have a static URL or can be accessed through
   * a proxy service can input the URL provider here. This URL provider is taken by
   * the `ApiLookupClient` and returned directly to the caller without any requests to
   * the API Lookup Service.
   *
   * @note This call should be synchronous without any tasks scheduled in
   * `TaskScheduler` as it might result in a dead-lock.
   *
   * @return An empty string if the catalog is invalid or unknown.
   */
```

olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
```diff
@@ -259,6 +274,18 @@ struct CORE_API ApiLookupSettings {
    */
   LookupEndpointProvider lookup_endpoint_provider =
       DefaultLookupEndpointProvider();
+
+  /**
+   * @brief The provider of endpoint for API requests.
```
HalynaDumych, line 279:
```suggestion
   * @brief The endpoint provider for API requests.
```

olp-cpp-sdk-core/include/olp/core/client/OlpClientSettings.h
```diff
@@ -259,6 +274,18 @@ struct CORE_API ApiLookupSettings {
...
+   *
+   * If some of the catalogs have fixed URLs and do not need an API Lookup
+   * beforehand you can provide the static URL via the CatalogEndpointProvider
+   * and every request will receive this URL from the ApiLookupClient without
+   * any HTTP request to the API Lookup service. This provider will be called
+   * prior to the lookup_endpoint_provider and in case the output is not empty
+   * the lookup_endpoint_provider will not be called additionally.
```
HalynaDumych, line 286:
```suggestion
   * If some of the catalogs have fixed URLs and do not need the API Lookup Service,
   * you can provide the static URL via `CatalogEndpointProvider`.
   * Every request will receive this URL from `ApiLookupClient` without
   * any HTTP requests to the API Lookup Service. `CatalogEndpointProvider` is called
   * before `lookup_endpoint_provider`, and if the output is not empty,
   * `lookup_endpoint_provider` is not called additionally.
```

olp-cpp-sdk-core/include/olp/core/client/ApiLookupClient.h
```diff
@@ -79,6 +79,12 @@ class CORE_API ApiLookupClient final {
    * @param callback The function callback used to receive the
    * `LookupApiResponse` instance.
    *
+   * @note If the catalog endpoint provider is set and will provide a static URL
+   * for this catalog, the method will not trigger any async download and
+   * instead will provide to the user the result synchronous. This means that
+   * the user needs to take special care in case the callback is called within
+   * the same context and avoid locking any mutex twice.
+   *
```
HalynaDumych, line 87:
```suggestion
   * @note If the catalog endpoint provider is set and provides a static URL
   * for this catalog, the method does not trigger any asynchronous download and
   * provides the synchronous result instead. This means that
   * the user needs to take special care in case the callback is called within
   * the same context and avoid locking any mutex twice.
   *
```

tests/integration/olp-cpp-sdk-core/ApiLookupClientTest.cpp
```diff
@@ -593,4 +589,132 @@ TEST_F(ApiLookupClientTest, LookupApiAsync) {
   }
 }
 
+TEST_F(ApiLookupClientTest, CustomCatalogProvider) {
```
andescu, line 592: An integration test does not mean copy/paste from unit tests and place it in another folder. It means you test **VersionedLayerClient** with the static URL and make sure that it works as you expect on the public API.

kzvieriev: Added VersionedLayerClient test.

Still, I would like to keep ApiLookupClient integration tests, because async `LookupApi` isn't used in dataservice::read.

andescu: If it brings any benefit so be it, but there is not much difference here between unit and integration tests. But let me not be the party pooper, so please keep them.

PR #1006, opened 2020-08-04
Add Protect/Release functional test

LiubovDidkivska:
Prefetch tiles and store in cache, so that maximum treshhold before
eviction is reached. Protect saved keys. Prefetch some other keys.
As treshold is reached an keys is protected, during prefetch previous
tile data is evicted, so in cache stays all protected keys and last
prefetched. Release protected keys and request some tile. Tile, which
was previously released was evicted.

Relates-To: OLPEDGE-2128

### DIFF
--- a/tests/functional/olp-cpp-sdk-dataservice-read/VersionedLayerClientProtectTest.cpp
+++ b/tests/functional/olp-cpp-sdk-dataservice-read/VersionedLayerClientProtectTest.cpp
@@ -1,0 +1,256 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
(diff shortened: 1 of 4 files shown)


tests/functional/olp-cpp-sdk-dataservice-read/VersionedLayerClientProtectTest.cpp
```diff
@@ -0,0 +1,263 @@
...
+  };
+
+  std::vector<olp::geo::TileKey> tiles_lover_levels;
+  std::vector<olp::geo::TileKey> tiles_upper_levels;
+  tiles_lover_levels.reserve(3840);
+  tiles_upper_levels.reserve(240);
+
+  {
+    SCOPED_TRACE("Mock prefetch tiles");
+    {
```
meravingen123, line 127: Why this scope is needed?

kzvieriev: its not a test case, why adding scoped trace?

tests/functional/olp-cpp-sdk-dataservice-read/VersionedLayerClientProtectTest.cpp
```diff
@@ -0,0 +1,259 @@
...
+  constexpr auto kQuadTreeDepth = 4;
+  constexpr auto kVersion = 44;
+
+  {
+    mock_server_client_->MockLookupResourceApiResponse(
+        mockserver::ApiDefaultResponses::GenerateResourceApisResponse(
+            kTestHrn));
+    mock_server_client_->MockGetVersionResponse(
+        mockserver::ReadDefaultResponses::GenerateVersionResponse(kVersion));
+  }
```
kzvieriev, line 99: lets group up mock server expectation setup

tests/functional/olp-cpp-sdk-dataservice-read/VersionedLayerClientProtectTest.cpp
kzvieriev: ah its already fixed

PR #1007, opened 2020-08-04
Allow using immutable cache from a read only partition

wcscmp:
It should be possible to use protected cache from a read-only partition. Level DB writes some metadata when opening existing DB. Add environment overload that does not write anything to disk but reports success.

### DIFF
--- a/olp-cpp-sdk-core/CMakeLists.txt
+++ b/olp-cpp-sdk-core/CMakeLists.txt
@@ -220,6 +220,8 @@ set(OLP_SDK_CACHE_SOURCES
     ./src/cache/ProtectedKeyList.h
     ./src/cache/InMemoryCache.cpp
     ./src/cache/InMemoryCache.h
+    ./src/cache/ReadOnlyEnv.cpp
+    ./src/cache/ReadOnlyEnv.h
 )
 
 set(OLP_SDK_CLIENT_SOURCES
@@ -356,7 +358,7 @@ add_library(${PROJECT_NAME} ${OLP_SDK_CORE_SOURCES} ${OLP_SDK_CORE_HEADERS})
 
 
 if (OLP_SDK_DISABLE_DEBUG_LOGGING)
-    target_compile_definitions(${PROJECT_NAME} 
+    target_compile_definitions(${PROJECT_NAME}
         PUBLIC LOGGING_DISABLE_DEBUG_LEVEL)
 endif()
 
--- a/olp-cpp-sdk-core/src/cache/DiskCache.h
+++ b/olp-cpp-sdk-core/src/cache/DiskCache.h
@@ -45,7 +45,7 @@ class DB;
 
 namespace olp {
 namespace cache {
-class DiskCacheSizeLimitEnv;
+class SizeCountingEnv;
 
 /**
  * @brief The OpenResult enum defines the result of the open() operation.
@@ -148,7 +148,7 @@ class DiskCache {
   std::string disk_cache_path_;
   std::unique_ptr<leveldb::DB> database_;
   std::unique_ptr<const leveldb::FilterPolicy> filter_policy_;
-  std::unique_ptr<DiskCacheSizeLimitEnv> environment_;
+  std::unique_ptr<SizeCountingEnv> environment_;
   std::unique_ptr<LevelDBLogger> leveldb_logger_;
   uint64_t max_size_{kSizeMax};
   bool check_crc_{false};
--- a/olp-cpp-sdk-core/src/cache/ReadOnlyEnv.cpp
+++ b/olp-cpp-sdk-core/src/cache/ReadOnlyEnv.cpp
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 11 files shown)


olp-cpp-sdk-core/src/cache/DiskCache.h
```diff
@@ -148,7 +149,8 @@ class DiskCache {
   std::string disk_cache_path_;
   std::unique_ptr<leveldb::DB> database_;
   std::unique_ptr<const leveldb::FilterPolicy> filter_policy_;
-  std::unique_ptr<DiskCacheSizeLimitEnv> environment_;
+  std::unique_ptr<DiskCacheSizeLimitEnv> disk_cache_size_limit_environment_;
```
andescu, line 152: If it is not possible to use `DiskCacheSizeLimitEnv` as base class for `ReadOnlyEnv` then please keep the analogy in sync and name this either `default_environment` or `read_write_environment`.

olp-cpp-sdk-core/src/cache/ReadOnlyEnv.h
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include <cstdint>
+
+#include <leveldb/env.h>
+
+namespace olp {
+namespace cache {
+
+class ReadOnlyEnv : public leveldb::EnvWrapper {
```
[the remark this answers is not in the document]
wcscmp: Done

olp-cpp-sdk-core/src/cache/ReadOnlyEnv.cpp
```diff
@@ -0,0 +1,66 @@
...
+ * License-Filename: LICENSE
+ */
+
+#include "ReadOnlyEnv.h"
+
+namespace olp {
+namespace cache {
+namespace {
+
+class MockWritableFile : public leveldb::WritableFile {
```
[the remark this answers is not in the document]
wcscmp: Done

olp-cpp-sdk-core/src/cache/DiskCache.h
wcscmp: Done

wcscmp:
> Sorry missed adding: please provide unit/integration tests in the same commit.

Done

olp-cpp-sdk-core/CMakeLists.txt
```diff
@@ -399,6 +401,7 @@ target_link_libraries(${PROJECT_NAME}
         ${CMAKE_THREAD_LIBS_INIT}
     PRIVATE
         leveldb::leveldb
+        stdc++fs
```
andescu, line 404: We are C++11 only. Use boost::filesystem if it is needed.

olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
```diff
@@ -197,8 +203,9 @@ TEST_F(DefaultCacheImplTest, LruCachePut) {
 
     cache.Open();
     cache.Clear();
-    cache.Put(key, data_string, [=]() { return data_string; },
-              (std::numeric_limits<time_t>::max)());
+    cache.Put(
+        key, data_string, [=]() { return data_string; },
+        (std::numeric_limits<time_t>::max)());
```
andescu, line 208: Please make sure you are using clang-format v6 so that we do not have this reformatting issues.

olp-cpp-sdk-core/include/olp/core/utils/Dir.h
```diff
@@ -105,6 +158,17 @@ class CORE_API Dir {
...
+
+  /**
+   * @brief Change permisions of the directory and all it's content.
+   *
+   * @param path Path to the directory.
+   * @param permisions_mask PermisionsMask mask
+   *
+   * @return \c true if any file with the given path exists, \c false otherwise.
+   */
+  static bool ChangePermisions(const std::string& path,
```
andescu, line 170: If it is only needed for testing then I would prefer to have an anonymous function added in testing and not in public API.

olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
wcscmp: I have tried clang-format-5 but it does not accept some options, so I'm using clang-format-10
What clang-format should I use?

olp-cpp-sdk-core/CMakeLists.txt
wcscmp: boost filesystem is also not available sadly

olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
andescu: We are all using `clang-format version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)`

olp-cpp-sdk-core/CMakeLists.txt
andescu: Then please use Linux native function in test file only and add wrap the test in #if _linux_.

olp-cpp-sdk-core/include/olp/core/utils/Dir.h
wcscmp: Done

PR #1009, opened 2020-08-05
Make DefaultRetryCondition retry on 429 and 5xx.

kzvieriev:
DefaultRetryCondition retries only on INTERNAL_SERVER_ERROR and
SERVICE_UNAVAILABLE errors. According to current DP implementation we
should make retries on TOO_MANY_REQUESTS and all 5xx responses.

Extending DefaultRetryCondition retry list and appropriate test. Also
fixed value of INSUFFICIENT_STORAGE.

Resolves: OLPEDGE-2209


olp-cpp-sdk-core/src/client/OlpClientSettings.cpp
```diff
@@ -32,8 +32,21 @@ unsigned int DefaultBackdownPolicy(unsigned int milliseconds) {
...
+    case http::HttpStatusCode::GATEWAY_TIMEOUT:
+    case http::HttpStatusCode::VERSION_NOT_SUPPORTED:
+    case http::HttpStatusCode::VARIANT_ALSO_NEGOTIATES:
+    case http::HttpStatusCode::INSUFFICIENT_STORAGE:
+    case http::HttpStatusCode::LOOP_DETECTED:
+    case http::HttpStatusCode::BANDWIDTH_LIMIT_EXCEEDED:
+    case http::HttpStatusCode::NOT_EXTENDED:
+    case http::HttpStatusCode::NETWORK_AUTHENTICATION_REQUIRED:
+    case http::HttpStatusCode::NETWORK_READ_TIMEOUT:
+    case http::HttpStatusCode::NETWORK_CONNECT_TIMEOUT:
```
andescu, line 49: I would change this into:
```cpp
if ((response.status >= INTERNAL_SERVER_ERROR && response.status <= NETWORK_CONNECT_TIMEOUT)
    || response.status == TOO_MANY_REQUESTS) return true;
return false;
```
There are to many status codes between 500 and 599 to list here.

PR #1010, opened 2020-08-05
Fix reading protected list from cache

LiubovDidkivska:
pubsetbuf is dependent on the library implementation,
so setting buffer for istream changed.
Move Protect implementation from VersionedLayerClient
to ProtectDependencyResolver.
Add tests for cache to check that Protect/Release works
only with mutable cache.

Relates-To: OLPEDGE-2034

### DIFF
--- a/olp-cpp-sdk-core/src/cache/ProtectedKeyList.cpp
+++ b/olp-cpp-sdk-core/src/cache/ProtectedKeyList.cpp
@@ -29,6 +29,12 @@
 
 namespace {
 constexpr auto kLogTag = "ProtectedKeyList";
+
+class ReadBuffer : public std::basic_streambuf<char> {
+ public:
+  ReadBuffer(char* p, size_t l) { setg(p, p, p + l); }
+};
+
 }  // namespace
 
 namespace olp {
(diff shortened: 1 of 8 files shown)


olp-cpp-sdk-core/src/cache/ProtectedKeyList.cpp
```diff
@@ -41,10 +47,9 @@ bool ProtectedKeyList::Deserialize(KeyValueCache::ValueTypePtr value) {
   if (!value) {
     return false;
   }
-  std::stringbuf string_buf;
-  string_buf.pubsetbuf(reinterpret_cast<char*>(value->data()), value->size());
 
-  std::istream stream(&string_buf);
+  ReadBuffer buf(reinterpret_cast<char*>(value->data()), value->size());
+  std::istream stream(&buf);
```
kzvieriev, line 52:
```suggestion
  std::istringstream stream({value->begin(), value->end()});
```
I think, `istringstream` fits good here and no need for cast and additional class

LiubovDidkivska: maybe better do not copy protected list

olp-cpp-sdk-dataservice-read/src/ProtectDependencyResolver.h
```diff
@@ -0,0 +1,71 @@
...
+namespace dataservice {
+namespace read {
+
+/// Find quad trees for protected tiles. Add quad tree to list for protection.
+/// Save downloaded quads for future searches for tiles to protect, if quad tree
+/// not found look it to the cache.
+class ProtectDependencyResolver {
+ public:
+  ProtectDependencyResolver(const client::HRN& catalog,
+                            const std::string& layer_id, const int64_t& version,
```
meravingen123, line 44: You can pass version without reference

PR #1020, opened 2020-08-12
Destruction network test.

kzvieriev:
Adding fv network test. Goal of the test is to check callback is
called if network destroyed. Also added possibility to set response
delay to be sure the requests are not finished before the destruction.

Resolves: OLPEDGE-2061

### DIFF
--- a/tests/functional/network/DestructionTest.cpp
+++ b/tests/functional/network/DestructionTest.cpp
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


tests/functional/network/DestructionTest.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <olp/core/http/NetworkSettings.h>
+
+#include "NetworkTestBase.h"
+#include "ReadDefaultResponses.h"
+
+namespace {
+using NetworkRequest = olp::http::NetworkRequest;
+using NetworkResponse = olp::http::NetworkResponse;
+
+class DestructionTest : public NetworkTestBase {};
```
andescu, line 34: If you do not modify or extend, use alias, e.g. `using DestructionTest = NetworkTestBase;`

tests/functional/network/DestructionTest.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+TEST_F(DestructionTest, Callback) {
+  const std::string kUrl = "http://some-url.com";
```
andescu, line 37: I would use different URLs for each request to make sure it is not merged internally into one handle. You never know what Network impl is doing internally.

PR #1028, opened 2020-08-18
Fix cppcheck issues

meravingen123:
Rename method parameters to match declaration and definition
Remove unused variables

Resolves: OLPEDGE-1062


olp-cpp-sdk-dataservice-read/src/repositories/PartitionsRepository.h
```diff
@@ -61,7 +61,7 @@ class PartitionsRepository {
 
   PartitionsResponse GetPartitionById(const std::string& layer,
                                       boost::optional<int64_t> version,
-                                      const DataRequest& request,
+                                      const DataRequest& data_request,
```
kzvieriev, line 64: I think, `request` name was fine :-)

PR #1030, opened 2020-08-19
Extended IsCached API to support aggregated tiles

mykhailo-kuchma:
This gives functionality to check if the aggregated data is available.

Change the IsCached behavior when the version is not set.

Resolves: OLPSUP-11431

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
@@ -66,8 +66,7 @@ class VersionedLayerClientImpl {
   virtual client::CancellationToken GetData(TileRequest request,
                                             DataResponseCallback callback);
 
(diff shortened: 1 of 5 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
```diff
@@ -88,7 +88,7 @@ class VersionedLayerClientImpl {
 
   virtual bool IsCached(const std::string& partition_id) const;
 
-  virtual bool IsCached(const geo::TileKey& tile) const;
+  virtual bool IsCached(const geo::TileKey& tile, bool aggregated = false) const;
```
kzvieriev, line 91: I think, we need to add same HasAggregatedData() method to the impl.

LiubovDidkivska (approved):

kzvieriev (approved):

PR #1055, opened 2020-09-14
[POC] Priority support

mykhailo-kuchma:
This POC shows how the requests priority could be implemented for
layer clients requests.

Relates-To: OLPEDGE-1023

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/thread/TaskScheduler.h
+++ b/olp-cpp-sdk-core/include/olp/core/thread/TaskScheduler.h
@@ -23,9 +23,12 @@
 
 #include <olp/core/client/ApiResponse.h>
 #include <olp/core/client/CancellationContext.h>
+#include <olp/core/utils/WarningWorkarounds.h>
 
 namespace olp {
 namespace thread {
+
+enum Priority : uint32_t { LOW = 100, NORMAL = 500, HIGH = 1000 };
 
 /**
  * @brief An abstract interface that is used as a base for the custom thread
@@ -51,6 +54,10 @@ class CORE_API TaskScheduler {
    * pipeline.
    */
   void ScheduleTask(CallFuncType&& func) { EnqueueTask(std::move(func)); }
+
+  void ScheduleTask(CallFuncType&& func, Priority priority) {
+    EnqueueTask(std::move(func), priority);
+  }
 
   /**
    * @brief Schedules the asynchronous cancellable task.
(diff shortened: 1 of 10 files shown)


olp-cpp-sdk-core/include/olp/core/thread/TaskScheduler.h
```diff
@@ -97,6 +104,11 @@ class CORE_API TaskScheduler {
    * kept. Once this method is called, you own the task.
    */
   virtual void EnqueueTask(CallFuncType&&) = 0;
+
+  virtual void EnqueueTask(CallFuncType&& func, Priority priority) {
```
andescu, line 108: I would not use enum here, as this again limits us to only 3 values. Use uint32_t or uint16_t and provide the enum just as a helper.

olp-cpp-sdk-dataservice-read/src/Common.h
```diff
@@ -86,7 +86,27 @@ inline client::CancellationToken AddTask(
...
     context.Execute();
     pending_requests->Remove(context);
-  });
+  }, thread::NORMAL);
+
+  return context.CancelToken();
+}
+
+template <typename Function, typename Callback, typename... Args>
+inline client::CancellationToken AddTaskWithPriority(
```
andescu, line 95: Is it not possible to also name this AddTask() and forward the first one to this, e.g.

```cpp
template <typename Function, typename Callback, typename... Args>
inline client::CancellationToken AddTask(
    const std::shared_ptr<thread::TaskScheduler>& task_scheduler,
    const std::shared_ptr<client::PendingRequests>& pending_requests,
    Function task, Callback callback, Args&&... args) {
      return AddTask(task_scheduler, pending_requests, std::move(task), std::move(callback), thread::NORMAL, std::forward<Args>(args)...);
}
```

olp-cpp-sdk-dataservice-read/src/Common.h
```diff
@@ -86,7 +86,27 @@ inline client::CancellationToken AddTask(
...
+  auto context = client::TaskContext::Create(
+      std::move(task), std::move(callback), std::forward<Args>(args)...);
+  pending_requests->Insert(context);
+
+  repository::ExecuteOrSchedule(task_scheduler,
+                                [=] {
+                                  context.Execute();
+                                  pending_requests->Remove(context);
+                                },
+                                priority);
```
andescu, line 109: The formatting here seems off.

PR #1056, opened 2020-09-14
Fix bug in script,catch crashes too

ystefinko:
Add verification of exitcode is not-zero after gtest run
of functional test.

Relates-To: OLPEDGE-2297

### DIFF
--- a/scripts/linux/fv/gitlab_test_fv.sh
+++ b/scripts/linux/fv/gitlab_test_fv.sh
@@ -46,7 +46,7 @@ ulimit -c unlimited
 # Stop after 2nd retry.
 source ${FV_HOME}/olp-cpp-sdk-functional-test.variables
 set +e
-for RETRY_COUNT in 1 2
+for RETRY_COUNT in 1 2 3
 do
 	echo "This is (${RETRY_COUNT}) time run ..."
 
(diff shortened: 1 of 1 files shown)


scripts/linux/fv/gitlab_test_fv.sh
```diff
@@ -53,7 +53,7 @@ do
 	# Run functional tests
 	${FV_HOME}/gitlab-olp-cpp-sdk-functional-test.sh
 
-	if [[ $? -eq 1 ]]; then
+	if [[ $? -ne 0 ]]; then
```
andescu, line 56: I don't think this is correct.
Once a test crashes there is no reason why to repeat it multiple times, it is just a waste of CI time and developer time who needs to scroll for half an hour to find the crash in the log.
In my opinion you should add 3 checks:

```
if [[ $? -eq 1 ]]; then
  # Handle flacky test and repeat up to 3 times
elif  [[ $? -gt 1 ]]; then
  # Crash happened -> exit 1
else
  # Success -> break
```

ystefinko: Good advice, thanks!

andescu (approved):

kzvieriev (approved):

PR #1061, opened 2020-09-16
Change PublishIndex() to be sync

meravingen123:
Change write::IndexLayerClient::PublishIndex() to use sync
method approach. We need to make it similar to
dataservice::read. Making it sync will simplify the code.

Resolves: OLPEDGE-2301

### DIFF
--- a/olp-cpp-sdk-dataservice-write/src/IndexLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-write/src/IndexLayerClientImpl.cpp
@@ -27,6 +27,7 @@
 #include <olp/core/client/CancellationContext.h>
 
 #include "ApiClientLookup.h"
+#include "Common.h"
 #include "generated/BlobApi.h"
 #include "generated/ConfigApi.h"
 #include "generated/IndexApi.h"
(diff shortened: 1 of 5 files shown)


kzvieriev (approved):

olp-cpp-sdk-dataservice-write/src/IndexLayerClientImpl.cpp
```diff
@@ -153,101 +154,63 @@ client::CancellationToken IndexLayerClientImpl::PublishIndex(
...
-      };
-
-  auto insert_indexes_function = [=]() -> client::CancellationToken {
-    auto index = request.GetIndex();
-    index.SetId(data_handle);
-    return IndexApi::insertIndexes(
-        *self->apiclient_index_, index, request.GetLayerId(),
-        request.GetBillingTag(), insert_indexes_callback);
-  };
+  auto publish_task =
```
andescu, line 159: Please be aware that you also need to change L145 to L155 as both are calling the user callback sync which is not very good. All should be triggered on the TaskScheduler.

olp-cpp-sdk-dataservice-write/src/generated/IndexApi.h
```diff
@@ -95,6 +95,16 @@ class IndexApi {
       const std::string& layer_id,
       const boost::optional<std::string>& billing_tag,
       InsertIndexesCallback callback);
+
+  /**
+  * @brief Synchronous version of \c insertIndexes method.
+  */
+  static InsertIndexesResponse insertIndexes(
```
andescu, line 102:
```suggestion
  static InsertIndexesResponse InsertIndexes(
```

andescu (approved):

mykhailo-kuchma (approved):

PR #1065, opened 2020-09-17
Add priority queue class.

kzvieriev:
Priority queue is needed for requests priority feature. It will be
used by ThreadPoolTaskScheduler. We could use std::priority_queue
instead, but it's not respect order of requests with same priority.

The class has same complexity and similar API as the stl one and also
values with equal priority follow FIFO logic.

PriorityQueueExtended has API like std::queue, because it should be
used as SyncQueue container. Its conflicting with our code style. Impl
class added to split conflicting part from implementation.

Resolves: OLPEDGE-2305

### DIFF
--- a/olp-cpp-sdk-core/tests/thread/PriorityQueueExtendedTest.cpp
+++ b/olp-cpp-sdk-core/tests/thread/PriorityQueueExtendedTest.cpp
@@ -1,0 +1,180 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#include <queue>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "thread/PriorityQueueExtended.h"
+
+namespace {
+namespace thread = olp::thread;
+
+struct TestObject {
+  int value;
+  int id;
+};
+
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-core/src/thread/PriorityQueue.h
```diff
@@ -0,0 +1,224 @@
...
+      bool rc = compare_(lhs.obj, rhs.obj);  // check less
+      if (!rc) {
+        rc = !compare_(rhs.obj, lhs.obj);  // check greater
+        if (rc)                            // both objects are considered equal
+        {
+          rc = lhs.id > rhs.id;
+        }
+      }
+
+      return rc;
```
andescu, line 119: You should be able to merge this:

```cpp
return (compare_(lhs.obj, rhs.obj) || (!compare_(rhs.obj, lhs.obj) && lhs.id > rhs.id));
```

olp-cpp-sdk-core/src/thread/PriorityQueue.h
```diff
@@ -0,0 +1,224 @@
...
+  /// prorized queue comparator
+  Compare compare_;
+  /// id counter
+  std::uint32_t next_id_ = 0;
+};
+
+template <class T, class COMPARE>
+PriorityQueueExtended<T, COMPARE>::PriorityQueueExtended(
+    const COMPARE& compare) {
+  impl_ = std::make_shared<PriorityQueueExtendedImpl>(compare);
```
andescu, line 134: Initialize the impl_ in the initializer list, i.e.

```cpp
template <class T, class COMPARE>
PriorityQueueExtended<T, COMPARE>::PriorityQueueExtended(
    const COMPARE& compare) : impl_{std::make_shared<PriorityQueueExtendedImpl>(compare)} {}
```

olp-cpp-sdk-core/tests/thread/PriorityQueueExtendedTest.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "thread/PriorityQueue.h"
+
+namespace {
+namespace thread = olp::thread;
+
+TEST(PriorityQueueExtendedTest, BaseFunctionality) {
```
andescu, line 30: Please add an additional test, which verifies the order among the same priority requests is kept.
Also, add proper unit tests for all methods to make sure all work properly.

andescu (dismissed):

andescu (approved):

mykhailo-kuchma (approved):

PR #1080, opened 2020-10-07
Init get version response.

LiubovDidkivska:
In case if we do not request version
online CatalogVersionResponse variable
stays initialized bu default with
error code unknown. If getting from cache
also fails we do not exchange result,
so it should be NotFound.

Relates-To: OLPEDGE-2339


olp-cpp-sdk-dataservice-read/src/repositories/CatalogRepository.cpp
```diff
@@ -156,7 +158,7 @@ CatalogVersionResponse CatalogRepository::GetLatestVersion(
   }
 
   return version_response;
-}
+}  // namespace repository
```
kzvieriev, line 161:
```suggestion
}
```

kzvieriev (approved):

kzvieriev (dismissed):

PR #1082, opened 2020-10-08
Add responses generator to test-common lib.

LiubovDidkivska:
Integration tests should have dependency only
from public dataservice api. Move generating
urls and responses to test-common lib.

Relates-To: OLPEDGE-1901

### DIFF
--- a/tests/common/ResponseGenerator.cpp
+++ b/tests/common/ResponseGenerator.cpp
@@ -25,31 +25,19 @@
 #include <olp/dataservice/read/model/Partitions.h>
 #include <olp/dataservice/read/model/VersionResponse.h>
 #include "ApiDefaultResponses.h"
-#include "PathGenerator.h"
+#include "PlatformUrlsGenerator.h"
 #include "ReadDefaultResponses.h"
 #include "generated/model/Api.h"
 // clang-format off
 #include "generated/serializer/ApiSerializer.h"
 #include "generated/serializer/VersionResponseSerializer.h"
 #include "generated/serializer/PartitionsSerializer.h"
+#include <olp/core/generated/serializer/SerializerWrapper.h>
 #include "generated/serializer/JsonSerializer.h"
 // clang-format on
 
-namespace {
-template <class T>
-std::string serialize(std::vector<T> data) {
-  std::string str = "[";
-  for (const auto& el : data) {
-    str.append(olp::serializer::serialize(el));
-    str.append(",");
-  }
-  str[str.length() - 1] = ']';
-  return str;
-}
-}  // namespace
-
 std::string ResponseGenerator::ResourceApis(const std::string& catalog) {
-  return serialize(
+  return olp::serializer::serialize(
       mockserver::ApiDefaultResponses::GenerateResourceApisResponse(catalog));
 }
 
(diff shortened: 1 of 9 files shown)


tests/common/UrlGenerators.h
```diff
@@ -22,50 +22,26 @@
...
-    uint64_t version) {
-  std::string path = "/layers/" + layer + "/partitions?";
-  for (const auto& partition : partitions) {
-    path.append("partition=" + partition + "&");
-  }
-  path.append("version=" + std::to_string(version));
-  return path;
-}
+class PathGenerator {
+ public:
```
kzvieriev, line 29: Class and file naming is confusing. Lets name both `UrlGenerator` or `PathGenerator`.

tests/common/GenerateStringResponses.cpp
```diff
@@ -0,0 +1,65 @@
...
+  str[str.length() - 1] = ']';
+  return str;
+}
+}  // namespace
+
+std::string ResponseGenerator::ResourceApis(const std::string& catalog) {
+  auto api_response =
+      mockserver::ApiDefaultResponses::GenerateResourceApisResponse(catalog);
+  return serialize(api_response);
+}
```
kzvieriev, line 55: Why we need custom serialize here?
We have serialize for vector and for api, so should be no problem to use`olp::serializer::serialize(api_response)`

tests/common/UrlGenerators.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+std::string PathGenerator::GetQuadKey(const std::string& quadkey,
+                                      const std::string& layer,
+                                      uint64_t version, uint64_t depth) {
+  return "/layers/" + layer + "/versions/" + std::to_string(version) +
+         "/quadkeys/" + quadkey + "/depths/" + std::to_string(depth);
+}
+
+std::string PathGenerator::FullPath(const std::string& apis_response,
```
kzvieriev, line 61: maybe we can use apis model instead, so no need for converting apis to json and back.

tests/common/GenerateStringResponses.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include <olp/dataservice/read/model/Partitions.h>
+
+class ResponseGenerator {
```
kzvieriev, line 27: Same here. I think `ResponseGenerator.h` suits better :-)

tests/common/UrlGenerators.cpp
LiubovDidkivska: In this case we will have dependancy in integration tests fron src. Here we want to keep all such dependancies in test-common lib.

kzvieriev (dismissed):

tests/common/ResponseGenerator.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "PlatformUrlsGenerator.h"
+#include "ReadDefaultResponses.h"
+#include "generated/model/Api.h"
+// clang-format off
+#include "generated/serializer/ApiSerializer.h"
+#include "generated/serializer/VersionResponseSerializer.h"
+#include "generated/serializer/PartitionsSerializer.h"
+#include <olp/core/generated/serializer/SerializerWrapper.h>
+#include "generated/serializer/JsonSerializer.h"
+
```
kzvieriev, line 37: // clang-format on

meravingen123 (approved):

kzvieriev (approved):

PR #1083, opened 2020-10-08
Fix the prefetch for the aggregated tiles.

mykhailo-kuchma:
Fix the prefetch API behavior for the aggregated tiles when the
aggregated tile is too far away. A new approach is to download and
store quadtrees of lower levels until the needed aggregated tile is
placed in sub quads (which means that we can find it later).

Relates-To: OLPEDGE-2312

### DIFF
--- a/tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerPrefetch.cpp
+++ b/tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerPrefetch.cpp
@@ -19,15 +19,18 @@
 
 #include "VersionedLayerTestBase.h"
 
+#include <olp/dataservice/read/PrefetchTileResult.h>
+#include <olp/dataservice/read/PrefetchTilesRequest.h>
+
 namespace {
 
 namespace read = olp::dataservice::read;
 
 constexpr auto kWaitTimeout = std::chrono::seconds(3);
 
-class VersionedLayerGetAggregatedDataTest : public VersionedLayerTestBase {};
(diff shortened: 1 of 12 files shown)


tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerPrefetch.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+namespace read = olp::dataservice::read;
+
+constexpr auto kWaitTimeout = std::chrono::seconds(3);
+
+class VersionedLayerPrefetch : public VersionedLayerTestBase {};
+
+TEST_F(VersionedLayerPrefetch, AggregatedPrefetch) {
+  const auto layer = "testlayer";
+  const auto ver = 7;
```
andescu, line 35: No abbrebiations.

andescu (approved):

tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerTestBase.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+#pragma once
+
+#include <gmock/gmock.h>
+
+#include <olp/dataservice/read/VersionedLayerClient.h>
+
+#include <mocks/NetworkMock.h>
+
```
LiubovDidkivska, line 27: Why do we need lines here?

LiubovDidkivska (approved):

PR #1087, opened 2020-10-12
Fix published partition Id for streamed partitions of size bigger than 20MB

matthiaskunter:
The published partiton Id must be equal to the trace Id of the respective request.

Resolves: OLPSUP-12034


andescu (commented):
Please sign-off the commit message and do not exceed 50 chars in the summary line and 72 chars for the detailed text.

andescu (approved):

mykhailo-kuchma (approved):

PR #1088, opened 2020-10-12
Do not fail prefetch if query fails

LiubovDidkivska:
Change query to return error only if all query
fail. Add test to check if all possible tiles
prefetched, even if some queries fail.
Propagate query error to item's responses.
If some query fails, related items should
fail with the same error. Other items
should fail as before with error NotFound.

Relates-To: OLPEDGE-2343


olp-cpp-sdk-dataservice-read/src/QueryMetadataJob.h
```diff
@@ -34,6 +35,16 @@
 namespace olp {
 namespace dataservice {
 namespace read {
+using EmptyResponse = Response<client::ApiNoResult>;
+
+template <typename QueryType, typename QueryResponseType>
+struct QueryResponse {
```
andescu, line 41: I have the feeling that this is getting out of hand with the templates.

olp-cpp-sdk-dataservice-read/src/QueryMetadataJob.h
```diff
@@ -34,6 +35,16 @@
...
+  QueryResponse() = default;
+  explicit QueryResponse(QueryType root) : root_(std::move(root)) {}
+  QueryType root_;
+  EmptyResponse response_;
+  typename QueryResponseType::ResultType list_of_tiles_;
```
andescu, line 46: Structures should not use `_` in the member name as all are public. See https://google.github.io/styleguide/cppguide.html#Variable_Names

andescu (commented):
I would prefer we first remove some of the templates before we add more functionality we will remove later.

ystefinko:
Gitlab Internal OCM verification passed !

ystefinko:
Gitlab Internal OCM verification failed !

ystefinko:
Gitlab Internal OCM verification failed !

PR #1096, opened 2020-11-02
Add HTTP Status log to IOS

meravingen123:
Add missing HTTP status code to IOS build
on task completion

Resolves: OLPEDGE-2340


olp-cpp-sdk-core/src/http/ios/OLPHttpTask.mm
```diff
@@ -141,12 +141,18 @@ - (BOOL)cancel {
 #pragma mark - Response handlers
 
 - (void)didCompleteWithError:(NSError*)error {
+  int status = 0;
+  if (self.responseData)
+  {
+      status = self.responseData.status;
+  }
```
andescu, line 148: Why in case of an error? Will this also print 200/3xx status codes?

andescu (approved):

mykhailo-kuchma (approved):

PR #1098, opened 2020-11-04
Fixed shared_mutex for work with -fno-exceptions

andrey-kashcheev:
Exceptions could be disabled on the target platform.
When they are disabled shared_mutex will use std::abort()

Relates-To: OAM-756

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/porting/shared_mutex.h
+++ b/olp-cpp-sdk-core/include/olp/core/porting/shared_mutex.h
@@ -32,6 +32,19 @@
 #define HAVE_PTHREAD_RWLOCK
 #endif
 
+#ifndef THROW_OR_ABORT
+# if __cpp_exceptions
+#  define THROW_OR_ABORT(exception) (throw (exception))
+# else
+#  include <cstdlib>
+#  define THROW_OR_ABORT(exception) \
+    do {                            \
+      (void)exception;              \
+      std::abort();                 \
+    } while (0)
+# endif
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/include/olp/core/porting/shared_mutex.h
```diff
@@ -32,6 +32,15 @@
 #define HAVE_PTHREAD_RWLOCK
 #endif
 
+#ifndef THROW_OR_ABORT
+# if __cpp_exceptions
+#  define THROW_OR_ABORT(exception) (throw (exception))
+# else
+#  include <cstdlib>
+#  define THROW_OR_ABORT(exception) (std::abort())
```
andescu, line 40: This will result in compilation errors or warnings for unused vars. Please add cast to void for the exception class.

```cpp
(void)exception;
std::abort();
```

andrey-kashcheev: How this can cause any errors or warnings? I don't understand. AFAIK preprocessor just substitutes throw_or_abort(...) with std::abort() without any object creation or something else. We still on the parse stage.

andescu: The exception object is created and it is a class but not used anywhere if you only use std::abort().

andescu (approved):

kzvieriev (approved):

PR #1099, opened 2020-11-05
DefaultCache Open()/Close() individual caches.

kzvieriev:
Adding new Open/Close API for DefaultCache. Now it's possible to
disable protected or mutable cache, modify and enable it again. This
can be useful if for some reason we want to update individual cache
while DefaultCache is still in use.

Changed DiskCache behavior when try to open not existing folder for
read-only. Now it's creating folder and an empty leveldb DB.

Removed CreadteDir() function, since it's not used anymore and also
we have alternative util for that purpose(olp::utils::Dir::create()).

Resolves: OAM-731, OLPSUP-12312

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -29,6 +29,7 @@
 #include "olp/core/porting/make_unique.h"
 
 namespace {
+using CacheType = olp::cache::DefaultCache::CacheType;
 
 constexpr auto kLogTag = "DefaultCache";
 constexpr auto kExpirySuffix = "::expiry";
@@ -95,6 +96,18 @@ leveldb::CompressionType GetCompression(
              : leveldb::kSnappyCompression;
 }
 
+olp::cache::StorageSettings CreateStorageSettings(
+    const olp::cache::CacheSettings& settings) {
+  olp::cache::StorageSettings storage_settings;
+  storage_settings.max_disk_storage = settings.max_disk_storage;
+  storage_settings.max_chunk_size = settings.max_chunk_size;
+  storage_settings.enforce_immediate_flush = settings.enforce_immediate_flush;
+  storage_settings.max_file_size = settings.max_file_size;
+  storage_settings.compression = GetCompression(settings.compression);
+
+  return storage_settings;
+}
+
 int64_t GetElapsedTime(std::chrono::steady_clock::time_point start) {
   return std::chrono::duration_cast<std::chrono::milliseconds>(
              std::chrono::steady_clock::now() - start)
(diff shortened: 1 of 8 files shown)


andescu (commented):
UpdateRegions feature? What is this? :)
Please adapt the commit message to only mention features that are relevant for this repository!

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -124,6 +124,47 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::Open() {
   return SetupStorage();
 }
 
+DefaultCache::StorageOpenResult DefaultCacheImpl::Open(
+    DefaultCache::CacheType type) {
+  std::lock_guard<std::mutex> lock(cache_lock_);
+  if (!is_open_) {
+    return DefaultCache::OpenDiskPathFailure;
```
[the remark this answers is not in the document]
kzvieriev: Added `NotReady` error code.

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -445,5 +409,32 @@ bool DiskCache::RemoveKeysWithPrefix(const std::string& prefix,
   return result.IsSuccessful();
 }
 
+void DiskCache::InitializeDB(const StorageSettings& settings,
+                             const std::string& path) const {
+  std::unique_ptr<leveldb::DB> database;
+  leveldb::DB* db = nullptr;
+
+  auto open_options = CreateOpenOptions(settings, false);
+  leveldb::DB::Open(open_options, path, &db);
```
andescu, line 418: Why did you drop the status? Even if the first attempt is erroneous and we attempt to repair it doesn't mean that the second try will succeed. I don't think we should drop the status.

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -445,5 +409,32 @@ bool DiskCache::RemoveKeysWithPrefix(const std::string& prefix,
...
+}
+
+leveldb::Options DiskCache::CreateOpenOptions(const StorageSettings& settings,
+                                              bool is_read_only) const {
+  leveldb::Options options;
+  options.compression = settings.compression;
+  options.info_log = leveldb_logger_.get();
+  options.write_buffer_size = settings.max_chunk_size;
+  options.filter_policy = leveldb::NewBloomFilterPolicy(10);
+  options.create_if_missing = !is_read_only;
```
andescu, line 429: We should also create the db if it is not there.

kzvieriev: We are using `ReadOnlyEnv` for protected cache. It can't create files.

So only if cache folder was created, we also create db manually with `InitializeDB()`.

andescu (approved):

mykhailo-kuchma (approved):

PR #1104, opened 2020-11-16
Fixed typo in the CacheSettings

andrey-kashcheev:


andescu (commented):
Ticket missing. You will not be able to merge it without a ticket.
We may need to patch this in with another PR.

PR #1106, opened 2020-11-18
Marked DefaultCache-specific methods as virtual

andrey-kashcheev:
This is necessary for being able to fully mock this class.
Also fixed typo in the 'disk_path_protected' description

Relates-To: OAM-739


andescu (commented):
I don't think this is necessary. We usually only proceed this way for private classes and not for external classes.

HalynaDumych (approved):
Changes in the public documentation look good to me.

PR #1111, opened 2020-11-20
Implements merging of same requests in OlpClient.

andescu:
This commit implements merging of same URL requests
for the async OlpClient::CallApi() method by using
the previously created PendingUrlRequest(s) classes.
The merging is only performed in case the request
does not have any body as this could alter the
response.
Additionaly this commit adds support for HEAD and
PATCH requests, already supported by the Network
layer.
OlpClientTests adapted and extended accordingly
with additional test cases. All other tests using
mocked networked changed to returning the callback
asynchronously.

Relates-To: OLPEDGE-1805

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/client/OlpClient.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/OlpClient.h
@@ -33,7 +33,8 @@ namespace olp {
 namespace client {
 /**
  * @brief Executes HTTP requests by using the base url and the provided
- * parameters and body.
+ * parameters and body. This class will handle retries based on the
+ * RetrySettings and will merge all similar URL requests into one.
  */
 class CORE_API OlpClient {
  public:
@@ -42,6 +43,7 @@ class CORE_API OlpClient {
   using RequestBodyType = std::shared_ptr<std::vector<std::uint8_t>>;
 
   OlpClient();
+  OlpClient(const OlpClientSettings& settings, std::string base_url);
   virtual ~OlpClient();
 
   // Movable and copyable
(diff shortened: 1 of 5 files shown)


olp-cpp-sdk-core/include/olp/core/client/OlpClient.h
```diff
@@ -67,13 +71,19 @@ class CORE_API OlpClient {
...
    */
   ParametersType& GetMutableDefaultHeaders();
 
   /**
    * @brief Sets the client settings.
    *
+   * @note Handle with care and do not change while requests are ongoing.
+   * Ideally the settings would not change during the lifecycle of this
+   * instance.
+   *
```
mykhailo-kuchma, line 86: Maybe we can deprecate this?

mykhailo-kuchma (dismissed):

olp-cpp-sdk-core/include/olp/core/client/OlpClient.h
andescu: Seems reasonable. Will do it in a separate PR.

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -401,37 +466,72 @@ CancellationToken OlpClient::OlpClientImpl::CallApi(
...
+  // different body.
+  bool merge = (!post_body || post_body->empty());
+  OLP_SDK_LOG_DEBUG_F(kLogTag, "CallApi: url='%s', merge='%s'", url.c_str(),
+                      merge ? "true" : "false");
+
+  if (merge) {
+    request_ptr = (*pending_requests)[url];
+
+    // Add callback and prepare CancellationToken
+    auto call_id = request_ptr->Append(callback);
```
meravingen123, line 496: What if PendingRequests::OnRequestCompleted will be called in between? 
merge will be True and GetRetryCallback(true) will be called. Then it will try to execute PendingRequests::OnRequestCompleted with request ID that is already not in the map which will possibly cause that request will not be finished properly.

meravingen123 (approved):

mykhailo-kuchma (approved):

PR #1119, opened 2020-12-04
Add checks for integers overflow

LiubovDidkivska:
Add check to increment
size of qeries for prefech only
if it is not reached max value.

Add chack that size for QuadTreeIndex
blob do not overflow.

Relates-To: OLPEDGE-2396


andescu (commented):
This will not work, we already tried it. Once you use the vars to calculate a temp var you already produce the overflow. I will mitigate the issues in Veracode.

PR #1125, opened 2020-12-15
Make ThreadedTaskScheduler to move tasks.

kzvieriev:
For correct template deduction of rvalue references std::forward should
be used. Otherwise, scheduler task will be copied. Also added simple
test to check the issue.

Resolves: OLPEDGE-2418

### DIFF
--- a/olp-cpp-sdk-core/src/thread/PriorityQueueExtended.h
+++ b/olp-cpp-sdk-core/src/thread/PriorityQueueExtended.h
@@ -89,7 +89,8 @@ class PriorityQueueExtended<T, COMPARE>::PriorityQueueExtendedImpl {
     DistinguishableObject(std::uint32_t id, Args... args)
         : id(id), obj(args...) {}
     DistinguishableObject(std::uint32_t id, const T& obj) : id(id), obj(obj) {}
-    DistinguishableObject(std::uint32_t id, T&& obj) : id(id), obj(obj) {}
+    DistinguishableObject(std::uint32_t id, T&& obj)
+        : id(id), obj(std::move(obj)) {}
 
     std::uint32_t id;
     T obj;
@@ -137,7 +138,7 @@ void PriorityQueueExtended<T, COMPARE>::push(const T& value) {
 
 template <class T, class COMPARE>
 void PriorityQueueExtended<T, COMPARE>::push(T&& value) {
-  impl_->Push(value);
+  impl_->Push(std::move(value));
 }
 
 template <class T, class COMPARE>
@@ -176,7 +177,7 @@ void PriorityQueueExtended<T, COMPARE>::PriorityQueueExtendedImpl::Push(
 template <class T, class COMPARE>
 void PriorityQueueExtended<T, COMPARE>::PriorityQueueExtendedImpl::Push(
     T&& value) {
-  container_.push_back(DistinguishableObject(GetNextId(), value));
+  container_.push_back(DistinguishableObject(GetNextId(), std::move(value)));
   std::push_heap(container_.begin(), container_.end(), compare_);
 }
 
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-core/src/thread/ThreadPoolTaskScheduler.cpp
```diff
@@ -82,9 +82,7 @@ class ThreadPoolTaskScheduler::QueueImpl {
 
   bool Pull(ElementType& element) { return sync_queue_.Pull(element); }
 
-  void Push(ElementType&& element) {
-    sync_queue_.Push(std::forward<ElementType>(element));
-  }
+  void Push(ElementType&& element) { sync_queue_.Push(std::move(element)); }
```
mykhailo-kuchma, line 85: std::forward ?

mykhailo-kuchma (approved):

olp-cpp-sdk-core/src/thread/ThreadPoolTaskScheduler.cpp
andrey-kashcheev: For perfect forwarding we need a templated method (here we have a templated class).
E.g. `template<typename U> void Push(...){...}`

kzvieriev: But `ThreadPoolTaskScheduler::QueueImpl` is not a templated class also, so `std::forward` is just a `std::move` here.

olp-cpp-sdk-core/src/thread/PriorityQueueExtended.h
```diff
@@ -89,7 +89,8 @@ class PriorityQueueExtended<T, COMPARE>::PriorityQueueExtendedImpl {
     DistinguishableObject(std::uint32_t id, Args... args)
         : id(id), obj(args...) {}
     DistinguishableObject(std::uint32_t id, const T& obj) : id(id), obj(obj) {}
-    DistinguishableObject(std::uint32_t id, T&& obj) : id(id), obj(obj) {}
+    DistinguishableObject(std::uint32_t id, T&& obj)
+        : id(id), obj(std::forward<T>(obj)) {}
```
andescu, line 93: As far as I know, this will not work as T obj is not a deducible type but a predefined type trough the  PriorityQueueExtended deduction, thus making it a rvalue by default. So theoretically the move constructor should be called when initializing the obj member with an rvalue.  So maybe the issue is somewhere else.

olp-cpp-sdk-core/src/thread/PriorityQueueExtended.h
```diff
@@ -137,7 +138,7 @@ void PriorityQueueExtended<T, COMPARE>::push(const T& value) {
 
 template <class T, class COMPARE>
 void PriorityQueueExtended<T, COMPARE>::push(T&& value) {
-  impl_->Push(value);
+  impl_->Push(std::forward<T>(value));
```
andescu, line 141: std::forward only works on forwarding references and T is not a forwarding reference as it is specified at class instantiation and not on method call. So here it will have no effect at all, see https://en.cppreference.com/w/cpp/language/reference#Forwarding_references:

```cpp
template<class T> struct A {
    template<class U>
    A(T&& x, U&& y, int* p); // x is not a forwarding reference: T is not a
                             // type template parameter of the constructor,
                             // but y is a forwarding reference
};
```

kzvieriev: good catch :-)
`T&&` here is an rvalue reference type, but `value` is still variable with name so its an lvalue and must be moved. Thus, `std::forward` is same as `std::move`.
replaced it with `std::move` , accordingly.

andrey-kashcheev (approved):

mykhailo-kuchma (approved):

PR #1129, opened 2020-12-16
Add method to get current size and set max size of the mutable cache

mykolamalik:
Using Size() method users will be able to check current size of the data stored
in the mutable cache. Note that the returned size is approximate and
may not match cache folder size. It is the size of real data and
does not include levelDB overhead.

Using SetMaxSize() method users will be able to change maximum alowed size
of the mutable cache on a fly. If size increases, no side actions happen.
If size decreases, the data that doesn't fit into new size (taking into account
data eviction thresholds) will be exicted from the cache. If a lot of data
must be evicted, it will be splitted to chunks that fits into
CacheSettings::max_eviction_portion.

Note that implementation is done into DefaultCacheImpl and not exposed to
DefaultCache public API. New APIs will be exposed in a separate commit when
Size() for the protected cache will be implemented.

Relates-To: OLPEDGE-2392

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -21,6 +21,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <cmath>
 #include <memory>
 #include <string>
 #include <utility>
@@ -38,6 +39,7 @@ constexpr auto kInternalKeysPrefix = "internal::";
 constexpr auto kMaxDiskSize = std::uint64_t(-1);
 constexpr auto kMinDiskUsedThreshold = 0.85f;
 constexpr auto kMaxDiskUsedThreshold = 0.9f;
+constexpr auto kEvictionPortion = 1024u * 1024u;  // 1 MB
 
 // current epoch time contains 10 digits.
 constexpr auto kExpiryValueSize = 10;
@@ -129,7 +131,8 @@ DefaultCacheImpl::DefaultCacheImpl(CacheSettings settings)
       mutable_cache_(nullptr),
       mutable_cache_lru_(nullptr),
       protected_cache_(nullptr),
-      mutable_cache_data_size_(0) {}
+      mutable_cache_data_size_(0),
+      eviction_portion_(kEvictionPortion) {}
 
 DefaultCache::StorageOpenResult DefaultCacheImpl::Open() {
   std::lock_guard<std::mutex> lock(cache_lock_);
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/include/olp/core/cache/CacheSettings.h
```diff
@@ -150,6 +150,17 @@ struct CORE_API CacheSettings {
...
+  /**
+   * @brief Sets the maximum portion of data that can be evicted from the mutable
+   * cache using single eviction call during decreasing maximum cache size.
+   * The larger this number, the more RAM will be needed for eviction. If bigger
+   * bunch of data must be evicted, eviction will be
+   * splitted to chunks with size limited to provided number.
+   *
+   * The default value is 128 MB.
+   */
+  uint64_t max_eviction_portion = 1024ull * 1024ull * 128ull;
```
andescu, line 163: Why is this necessary? This is not a logical setting as you can only know how much to set this once you know how much you need to decrease the size of the cache and in most cases, you probably do not know this as it is given by the user. But you can not also change this value dynamically. Why don't you set it internally to 1MB or so and just evict and see how much it will take?

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -550,15 +551,23 @@ uint64_t DefaultCacheImpl::MaybeEvictData(leveldb::WriteBatch& batch) {
...
+    const auto expiry_key = CreateExpiryKey(key);
+    const auto current_eviction_size =
+        key.size() + properties.size + expiry_key.size() + kExpiryValueSize;
 
-    // Remove the key's expiry
-    auto expiry_key = CreateExpiryKey(key);
-    batch.Delete(expiry_key);
-    evicted += expiry_key.size() + kExpiryValueSize;
+    if (eviction_limit && evicted + current_eviction_size > eviction_limit) {
+      OLP_SDK_LOG_INFO_F(
```
andescu, line 559: Rather debug. We only need a log at the end on how many items were evicted in total.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -575,15 +584,28 @@ uint64_t DefaultCacheImpl::MaybeEvictData(leveldb::WriteBatch& batch) {
...
-    evicted += key.size() + properties.size;
-    batch.Delete(key);
+    const auto expiry_key = CreateExpiryKey(key);
+    auto current_eviction_size = key.size() + properties.size;
+    if (IsExpiryValid(properties.expiry)) {
+      current_eviction_size += expiry_key.size() + kExpiryValueSize;
+    }
+
+    if (eviction_limit && evicted + current_eviction_size > eviction_limit) {
+      OLP_SDK_LOG_INFO_F(
```
andescu, line 594: You can imagine how these logs will flood the stdout on a 1GB eviction. Put it to debug, please.
Only one final log at the end with total numbers should suffice.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.h
```diff
@@ -67,6 +67,9 @@ class DefaultCacheImpl {
   bool Release(const DefaultCache::KeyListType& keys);
   bool IsProtected(const std::string& key) const;
 
+  uint64_t Size(DefaultCache::CacheType type) const;
+  uint64_t SetMaxSize(uint64_t new_size);
```
andescu, line 71: Both are named Size()

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -568,22 +633,46 @@ uint64_t DefaultCacheImpl::MaybeEvictData(leveldb::WriteBatch& batch) {
...
+                      "count=%d, evicted=%" PRIu64,
+                      count, evicted);
+  return {count, evicted, false};
+}
+
+DefaultCacheImpl::EvictionResult DefaultCacheImpl::EvictDataPortion(
+    leveldb::WriteBatch& batch) {
+  uint64_t evicted = 0u;
+  auto count = 0u;
+  const auto target_size = kMinDiskUsedThreshold * settings_.max_disk_storage;
```
andrey-kashcheev, line 647: Maybe we should use here MaxDiskUsedThreshold? Because I don't think that we want to remove any data which is not required for removal.

mykolamalik: It was an original logic: If the size is bigger than k**Max**DiskUsedThreshold * settings_.max_disk_storage _(condition)_, we evict until the size reaches k**Min**DiskUsedThreshold * settings_.max_disk_storage _(target)_. So I didn't change this logic in my commit

meravingen123 (approved):

mykhailo-kuchma (approved):

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -529,17 +532,84 @@ uint64_t DefaultCacheImpl::MaybeEvictData(leveldb::WriteBatch& batch) {
...
+                  ? left_to_evict
+                  : eviction_portion_;
+          eviction_batch = std::make_unique<leveldb::WriteBatch>();
+          eviction_result =
+              evict_method(this, *eviction_batch, current_eviction_target);
+
+          if (eviction_result.size >= eviction_portion_) {
+            // Apply intermediate batch
+            const auto apply_result =
+                mutable_cache_->ApplyBatch(std::move(eviction_batch));
```
kzvieriev, line 562: I am not sure about this. leveldb can be tricky :-)
AFAIK, leveldb isn't deleting data instantly after applying batch, it just marks keys as deleted. Thus, data is not available anymore, but still stored on the disk. Compaction should free not needed data.
Previously, eviction could happen only when adding new data. leveldb has internal mechanism to trigger the compaction, so cache size is keeping around `max_disk_storage` value.
Lowering cache size probably means user need some free space, so we definitely must trigger the compaction process.
The operation isn't cheap, so not sure what is better - compact whole cache each time `eviction_portion_` removed or only once in the end.
If we will do only 1 compaction after all data is removed, there is no need to split deletion into several batches, because real work done only in the end.

mykolamalik: Discussed in private. Compaction is added after every ApplyBatch.
Will be tested on a big amount of data. The algorithm may be optimized based on the results.

kzvieriev (approved):

meravingen123 (approved):

PR #1131, opened 2020-12-17
Add a check if the Prefetch operation triggered

mykhailo-kuchma:
A potential place where the task can be canceled and not trigger a
callback.

Relates-To: OLPEDGE-2423

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -281,7 +281,7 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchPartitions(
           }
         };
 
-        context.ExecuteOrCancelled([&]() {
+        bool prefetch_triggered = context.ExecuteOrCancelled([&]() {
           auto download_job =
               std::make_shared<PrefetchPartitionsHelper::DownloadJob>(
                   std::move(download), std::move(append_result),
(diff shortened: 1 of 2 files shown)


olp-cpp-sdk-core/include/olp/core/client/CancellationContext.h
```diff
@@ -50,6 +52,24 @@ class CORE_API CancellationContext {
...
+   * `CancellationContext` propagates a cancel request.
+   *
+   * Note: this method does not handle the case when context is alredy canceled
+   * and may produce unexpected results. Please prefer to use ExecuteOrCancelled
+   * with two arguments.
+   *
+   * @return True if the execution is successful; false if the context was
+   * cancelled.
+   */
+  OLP_SDK_DEPRECATED("Will be removed in 06.2021")
```
andescu, line 70: I don't think this is necessary. We might need this one day or the user needs it and then we need to add it back. It is up to the user to use what is given in a senseable way.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -281,18 +281,27 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchPartitions(
...
-          return PrefetchPartitionsHelper::Prefetch(
-              std::move(download_job), request.GetPartitionIds(),
-              std::move(query), task_sink_, kQueryPartitionsMaxSize,
-              request.GetPriority());
-        });
-
-        return EmptyResponse(client::ApiNoResult());
+        bool prefetch_triggered = true;
+
+        context.ExecuteOrCancelled(
```
andescu, line 286: Use the return value instead of adding an additional lambda for achieving the same thing.

Same for below changes.

andescu (approved):

meravingen123 (approved):

kzvieriev (approved):

PR #1134, opened 2020-12-28
Fix prefetch callback called twice after cancel.

kzvieriev:
TaskContext makes additional context.IsCancelled() check after
submitted task is executed. Thus, even if init prefetch task is
finished its still can be cancelled. Prefetch tasks chain can be
completed already, so both responses are sent to the user.
Wrapping user callback to ensure it called only once.

Resolves: OLPEDGE-2435

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -181,6 +181,16 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchPartitions(
   // shared_settings for saving some mb while prefetch.
   auto shared_settings = std::make_shared<client::OlpClientSettings>(settings_);
 
+  // User callback should be called only once. If prefetch is finished, but
+  // thread with init task is still pending, cancel will cause 2nd user callback
+  // call. Wrapping it to handle the case.
+  auto is_callback_called = std::make_shared<std::atomic<bool>>(false);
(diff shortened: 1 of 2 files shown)


mykhailo-kuchma (dismissed):

mykhailo-kuchma (approved):

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -292,16 +303,18 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchPartitions(
               request.GetPriority());
         });
 
+        std::this_thread::sleep_for(std::chrono::milliseconds(1000));
```
andrey-kashcheev, line 306: Quite a big delay. Why we need it?

mykhailo-kuchma: Please remove it, it's not needed.

mykhailo-kuchma (changes requested):

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
kzvieriev: ouch, that was tmp code to help reproduce the issue.
Removed :-)

andrey-kashcheev (approved):

mykhailo-kuchma (approved):

PR #1161, opened 2021-02-22
Reduce extra quads count for prefetch.

kzvieriev:
In case of sparse data we load additional quads, but current algorithm
looks on slice root instead of requested tiles when decides to load
extra metadata. In some cases that is not correct because for all
requested tiles bundles are not that far away.

Volatile client doesn't load extra quads, thus, only versioned client
changed.

Also added the test to cover the issue.

Resolves: OLPEDGE-2454

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -386,13 +386,6 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchTiles(
 
         const bool aggregation_enabled = request.GetDataAggregationEnabled();
 
-        auto query = [=](geo::TileKey root,
-                         client::CancellationContext inner_context) mutable {
-          return repository.GetVersionedSubQuads(root, kQuadTreeDepth, version,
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -402,6 +395,20 @@ client::CancellationToken VersionedLayerClientImpl::PrefetchTiles(
...
         };
 
+        auto query = [=](geo::TileKey root,
+                         client::CancellationContext inner_context) mutable {
+          auto response = repository.GetVersionedSubQuads(
+              root, kQuadTreeDepth, version, inner_context);
+
+          if (response.IsSuccessful() && aggregation_enabled) {
+            auto subquads = filter(response.GetResult());
+            repository.LoadAggregatedSubQuads(root, subquads, version,
```
andescu, line 405: Don't we need the network statistics for the prefetch status?

andescu (approved):

ystefinko:
Gitlab Internal OCM verification passed !

mykhailo-kuchma (approved):

PR #1166, opened 2021-02-26
Add merging of same requests in sync OlpClient:CallApi()

jmoro3:
Merges same requests in sync OlpClient:CallApi

Originally sync & async CallApi implementations were different:

-async merged same requests (added callbacks to the PendingUrlTasks);

-sync simply sent a request as is. To prevent concurrent sync
requests the workaround with the NamedMutex was used.

This commit improves the sync CallApi. Now it triggers internally
async implementation.
So same requests will add callbacks to the PendingUrlTasks for
any CallApi calls, and NamedMutex workaround does not need in
most cases (the only exception is ApiClientLookup::LookupApi).

Relates-To: OLPEDGE-2417

Signed-off-by: Iuliia Moroz <ext-iuliia.moroz@here.com>

### DIFF
--- a/olp-cpp-sdk-core/src/client/OlpClient.cpp
+++ b/olp-cpp-sdk-core/src/client/OlpClient.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2020 HERE Europe B.V.
+ * Copyright (C) 2019-2021 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -49,6 +49,7 @@ struct RequestSettings {
   std::chrono::milliseconds accumulated_wait_time{0};
   std::chrono::milliseconds current_backdown_period{0};
   const std::chrono::milliseconds max_wait_time{0};
+  olp::client::NetworkStatistics accumulated_statistics;
 };
 }  // namespace
 
@@ -88,10 +89,6 @@ HttpResponse ToHttpResponse(const http::NetworkResponse& response) {
 HttpResponse ToHttpResponse(const http::SendOutcome& outcome) {
   return {static_cast<int>(outcome.GetErrorCode()),
           http::ErrorCodeToString(outcome.GetErrorCode())};
-}
-
-bool StatusSuccess(int status) {
-  return status >= 0 && status < http::HttpStatusCode::BAD_REQUEST;
 }
 
 bool CaseInsensitiveCompare(const std::string& str1, const std::string& str2) {
@@ -137,15 +134,13 @@ void ExecuteSingleRequest(const std::shared_ptr<http::Network>& network,
         request, response_body,
         [=](const http::NetworkResponse& response) {
           auto status = response.GetStatus();
-          if (!StatusSuccess(status)) {
-            response_body->str(
-                response.GetError().empty()
-                    ? "Error occurred, please check HTTP status code"
-                    : response.GetError());
-          }
 
-          callback(response.GetRequestId(),
-                   {status, std::move(*response_body), std::move(*headers)});
+          HttpResponse http_response{status, std::move(*response_body),
+                                     std::move(*headers)};
+          http_response.SetNetworkStatistics(
+              {response.GetBytesUploaded(), response.GetBytesDownloaded()});
+
+          callback(response.GetRequestId(), std::move(http_response));
         },
         [=](std::string key, std::string value) {
           headers->emplace_back(std::move(key), std::move(value));
@@ -158,6 +153,7 @@ void ExecuteSingleRequest(const std::shared_ptr<http::Network>& network,
     }
 
     id = send_outcome.GetRequestId();
+
     return CancellationToken([=] { network->Cancel(id); });
   };
 
@@ -181,6 +177,7 @@ NetworkCallbackType GetRetryCallback(
     const NetworkRequestPtr& request) {
   return [=](const http::RequestId request_id, HttpResponse response) mutable {
     ++settings->current_try;
+    settings->accumulated_statistics += response.GetNetworkStatistics();
 
     if (CheckRetryCondition(*settings, retry_settings, response)) {
       // Response is either successull or retries count/time expired
@@ -190,8 +187,10 @@ NetworkCallbackType GetRetryCallback(
             "Wrong response received, pending_request=%" PRIu64
             ", request_id=%" PRIu64,
             pending_request->GetRequestId(), request_id);
-        return;
       }
+
+      // Set accumulated statistics for all retries
+      response.SetNetworkStatistics(settings->accumulated_statistics);
 
       if (merge) {
         const auto& url = request->GetUrl();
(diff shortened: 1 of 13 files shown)


olp-cpp-sdk-core/src/client/PendingUrlRequests.h
```diff
@@ -61,6 +65,19 @@ class PendingUrlRequest {
   bool ExecuteOrCancelled(const ExecuteFuncType& func,
                           const CancelFuncType& cancel_func = nullptr);
 
+  /// Calls the managed CancellationContext ExecuteOrCancelled so that it can
+  /// remember the func returned CancellationToken to be able to cancel any
+  /// async operation taking place in our name. And then waits for response.
+  /// Default wait time is 60 seconds.
+  bool ExecuteAndWait(
```
andescu, line 72: I am not sure if this is the right place for this. I would rather put it in OlpClient as it is more bound to the sync CallApi() method.

olp-cpp-sdk-core/src/client/PendingUrlRequests.cpp
```diff
@@ -33,6 +33,9 @@ olp::client::HttpResponse GetCancelledResponse() {
 namespace olp {
 namespace client {
 
+PendingUrlRequest::PendingUrlRequest(CancellationContext context)
```
andescu, line 36: Not sure this works. Consider this. The first request comes in and creates a PendingUrlRequest with its CancellationContext embedded. Now there is a second and a third request to the same URL that is merged inside the PendingUrlRequest. Now the first request is cancelled, which means we should only remove its callback from the callback list but because we used its CancellationContext as input the entire PendingUrlRequest is cancelled, incl. the second and third request which actually are still waiting for the download. Dont this this is good.

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -545,95 +487,67 @@ HttpResponse OlpClient::OlpClientImpl::CallApi(
...
+  if (!merge || !IsPending(request_ptr)) {
+    const auto& retry_settings = settings_.retry_settings;
+    auto network = settings_.network_request_handler;
+    auto request_settings = GetRequestSettings(retry_settings);
 
-    // do the periodical sleep and check for cancellation status in between.
-    auto duration_to_sleep =
-        std::min(backdown_period, max_wait_time - accumulated_wait_time);
-    accumulated_wait_time += duration_to_sleep;
+    ExecuteSingleRequest(
```
andescu, line 529: Is this enough? What about the retry and the backdown strategy? I don't see ExecuteSingleRequest doing it.

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -545,95 +487,67 @@ HttpResponse OlpClient::OlpClientImpl::CallApi(
...
-    // which uses headers dictionary without any duplicates.
-    // User agents entries are usually separated by a whitespace, e.g.
-    // Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:47.0) Firefox/47.0
-    if (CaseInsensitiveCompare(header.first, http::kUserAgentHeader)) {
-      custom_user_agent += header.second + std::string(" ");
-    } else {
-      network_request.WithHeader(header.first, header.second);
-    }
-  }
+  auto network_request = CreateRequest(path, method, query_params,
```
andescu, line 490: One though. Is it not the same if you call async CallApi() from the sync CallApi()?
Something like:

```cpp
HttpResponse CallApi(std::string path, std::string method,
                       ParametersType query_params,
                       ParametersType header_params, ParametersType form_params,
                       RequestBodyType post_body, std::string content_type,
                       CancellationContext context) const {
   auto callback = [&](HttpResponse http_response) {
        response_data->response = std::move(http_response);
        response_data->condition.Notify();
   };
  
   context.ExecuteOrCancelled([&](){return CallApi(path, method, ..., callback)}, 
                                                    [&](){ ... });
   
     // Wait for response_data->condition and return response.
}
```

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -545,95 +487,67 @@ HttpResponse OlpClient::OlpClientImpl::CallApi(
...
+  bool merge = (!post_body || post_body->empty());
+  OLP_SDK_LOG_DEBUG_F(kLogTag, "CallApi: url='%s', merge='%s'", url.c_str(),
+                      merge ? "true" : "false");
 
-  auto response =
-      SendRequest(network_request, settings_, retry_settings, context);
+  request_ptr = std::make_shared<PendingUrlRequest>(context);
 
-  NetworkStatistics accumulated_statistics = response.GetNetworkStatistics();
+  auto call_id = merge ? pending_requests->Append(url, request_ptr, callback)
```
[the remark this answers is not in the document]
jmoro3: The changes in PendingUrlRequests are already removed. It was in draft only.

olp-cpp-sdk-core/src/client/PendingUrlRequests.h
jmoro3: It was for draft only. Completely removed

olp-cpp-sdk-core/src/client/OlpClient.cpp
jmoro3: this is can be the most elegant solution. But wouldn't we need to add condition.Notify() into cancellation tokens created inside async CallApi?

andescu: No as it is added to the CancellationContext. If you look at CancellationContext::ExecuteOrCancelled() then you see that the CancellationToken of the sub-operation is remembered:

```cpp

File: olp-cpp-sdk-core/src/client/OlpClient.cpp [hunk 6]
```diff
@@ -545,95 +469,87 @@ HttpResponse OlpClient::OlpClientImpl::CallApi(
...
-  auto network_request = http::NetworkRequest(
-      utils::Url::Construct(GetBaseUrl(), path, query_params));
+  auto network_request = CreateRequest(path, method, query_params,
+                                       header_params, post_body, content_type);
 
-  network_request.WithVerb(GetHttpVerb(method))
-      .WithBody(std::move(post_body))
-      .WithSettings(std::move(network_settings));
+  PendingUrlRequestPtr request_ptr = nullptr;
+  auto network = settings_.network_request_handler;
```
andescu, line 476: Do we need to copy it? Can we just use a reference?

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -545,95 +469,87 @@ HttpResponse OlpClient::OlpClientImpl::CallApi(
...
+    cancellation_token = CancellationToken([=] {
+      request_ptr->Cancel(call_id);
+      response_data->condition.Notify();
+    });
+  }
 
-    if (!retry_settings.retry_condition(response)) {
-      return response;
-    }
+  context.ExecuteOrCancelled(
```
andescu, line 529: Looks a little strange, but should be ok.

olp-cpp-sdk-core/src/client/PendingUrlRequests.cpp
jmoro3: This is reverted change

olp-cpp-sdk-core/src/client/OlpClient.cpp
jmoro3: Also reverted

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -537,103 +466,49 @@ HttpResponse OlpClient::OlpClientImpl::CallApi(
...
+        return CancellationToken([=] {
+          token.Cancel();
+          response_data->condition.Notify();
+        });
+      },
+      [=]() { response_data->condition.Notify(); });
 
-    // In case we retry, accumulate the stats
-    accumulated_statistics += response.GetNetworkStatistics();
+  auto waiting_time = settings_.retry_settings.timeout *
```
andescu, line 500: The `timeout` is the total timeout for all retries, which means you should not multiple it. E.g. if the user specifies timeout = 60s, then the total expected time, incl. all retries and waits, should not exceed 60s.

olp-cpp-sdk-core/src/client/PendingUrlRequests.h
```diff
@@ -90,6 +90,13 @@ class PendingUrlRequest {
   /// Will be called when the Network response arrives.
   void OnRequestCompleted(HttpResponse response);
 
+  /// Returns request's status. It is 'true' for merged requests or for requests
+  /// that are waiting for the Network callback
+  bool IsPending() const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return http_request_id_ != kInvalidRequestId || callbacks_.size() > 1;
```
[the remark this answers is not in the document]
jmoro3: Fixed to 1u to avoid warning. But we need to check size() >= 2 here because the method is triggered right after a callback has been added.

olp-cpp-sdk-core/src/client/PendingUrlRequests.h
```diff
@@ -90,6 +90,13 @@ class PendingUrlRequest {
   /// Will be called when the Network response arrives.
   void OnRequestCompleted(HttpResponse response);
 
+  /// Returns request's status. It is 'true' for merged requests or for requests
+  /// that are waiting for the Network callback
+  bool IsPending() const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return http_request_id_ != kInvalidRequestId || callbacks_.size() > 1u;
```
andescu, line 97: Ahh, sorry. Missed that you actually check > 1. I thought you only check that there is something to wait for.

andescu (dismissed):

andescu (approved):

mykhailo-kuchma (approved):

PR #1172, opened 2021-03-15
Instruction on how to work with unsupported APIs

HalynaDumych:
- Document how to work with generic Data APIs that are currently
not supported by the Data SDK.
- Add an example on how to download the difference between two versions
of a catalog.

Relates-To: OLPEDGE-2488

### DIFF
--- a/docs/work-with-data-apis.md
+++ b/docs/work-with-data-apis.md
@@ -1,0 +1,69 @@
+# Work with unsupported Data REST APIs
+
+HERE Data SDK for C++ supports only a limited number of Data REST APIs. You can find a list of them on our page with the [overall architecture](OverallArchitecture.md).
+
+If you need to work with any Data REST APIs that are currently not supported, you can use the `olp::client::OlpClient` class from our core library for the request. It has an easy-to-use interface and a retry mechanism that uses a recommended backoff strategy for subsequent requests.
+
+The example below demonstrates how to use the `olp::client::OlpClient` class from the Data SDK to call the API for downloading the difference between two versions of a versioned catalog.
+
+**Example. Download the difference between two versions of a catalog.**
+
+1. Create the `OlpClientSettings` object.
+
+   For instructions, see [Create platform client settings](create-platform-client-settings.md).
+
+2. Configure an `OlpClient` instance in one of the following ways:
+    - Use the `olp::client::ApiLookupClient` class to perform a request and get a preconfigured `olp::client::OlpClient` instance.
+    - Use a base URL:
+        1. Get the base URL from the lookup API service.
+
+            For instructions, see the [API Lookup API Developer's Guide](https://developer.here.com/documentation/api-lookup/dev_guide/index.html)
+        
+        2. Create an `OlpClient` instance with the client settings from step 1 and the base URL.
+
+            ```cpp
+            olp::client OlpClient client(client_settings, base_url);
+            ```
+
+3. Prepare the REST API request parameters and the metadata URL.
+
+    ```cpp
+    std::multimap<std::string, std::string> header_params;
+    header_params.emplace("Accept", "application/json");
+
+    // Replace "my_layer" with the actual layer name.
+    std::string metadata_url = "/layers/my_layer/changes";
+
+    std::multimap<std::string, std::string> query_params;
+
+    // Required parameters.
+    query_params.emplace("startVersion", "0");
+    query_params.emplace("endVersion", "1");
+
+    // Optional parameters.
+    query_params.emplace("additionalFields", "dataSize,checksum");
+    query_params.emplace("billingTag", "Billing1234");
+
+    // Can be used to cancel the ongoing request. Needs to be triggered
+    // from another thread as `olp::client::OlpClient::CallApi()` is a blocking call.
+    CancellationContext context;
+    olp::client::HttpResponse response = client.CallApi(
+        metadata_url, "GET", query_params, header_params, {}, nullptr, "", context);
+    ```
+
+4. Check the response and, if needed, parse JSON.
+
+    ```cpp
+    if (response.GetStatus() != olp::http::HttpStatusCode::OK) {
+        // If the response is not OK, handle error cases.
+    }
+    ```
+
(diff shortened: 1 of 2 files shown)


docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
+# Work with unsupported REST Data APIs
+
+HERE Data SDK for C++ supports only a limited number of REST Data APIs. You can find a list of them on our page with the [overall architecture](OverallArchitecture.md).
+
+If you need to work with the REST Data APIs that are currently not supported, you can configure the Data SDK with a unique base URL.
```
andescu, line 5:
```suggestion
If you need to work with any Data REST APIs that are currently not supported, you can use the `olp::client::OlpClient` class from our core library for the request. By using it you will profit from the easy-to-use interface and the retry mechanism which will use the recommended backoff strategy for subsequent requests.
```

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+
+The example below demonstrates how to use the Data SDK to call the API for downloading the difference between two versions of a versioned catalog.
```
andescu, line 7:
```suggestion
The example below demonstrates how to use the `olp::client::OlpClient` class from the Data SDK to call the API for downloading the difference between two versions of a versioned catalog.
```

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+
+**Example. Download the difference between two versions of a catalog.**
```
andescu, line 9: Do we really need the extra heading?

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+
+1. Create the `OlpClientSettings` object.
+
+   For instructions, see [Create platform client settings](create-platform-client-settings.md).
+
+2. Get a base URL from the lookup API service.
+
+    For instructions, see the [API Lookup API Developer's Guide](https://developer.here.com/documentation/api-lookup/dev_guide/index.html).
```
andescu, line 17:
```suggestion
    For instructions, see the [API Lookup API Developer's Guide](https://developer.here.com/documentation/api-lookup/dev_guide/index.html) or use the `olp::client::ApiLookupClient` class which will perform the request and return a preconfigured and ready to use `olp::client::OlpClient` instance.
```

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+
+    You can now use this URL as the base for all requests to the REST Data API.
+
+3. Create an `OlpClient` instance with the client settings from step 1 and base URL from step 2.
+
```
andescu, line 22: Add a note maybe and say that this step can be omitted when using `olp::client::ApiLookupClient` instead.

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+    olp::client OlpClient client(client_settings, base_url);
+    ```
+
+4. Prepare the REST API request parameters and the metadata URL.
+
+    ```cpp
+    std::multimap<std::string, std::string> header_params;
+    header_params.emplace("Accept", "application/json");
+
+    std::string metadata_url = "/layers/my_layer/changes";
```
andescu, line 33:
```suggestion
    // Replace "my_layer" with the actual layer name
    std::string metadata_url = "/layers/my_layer/changes";
```

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+
+    // Required parameters.
+    query_params.emplace("startVersion", "0");
+    query_params.emplace("endVersion", "1");
+
+    // Optional parameters.
+    query_params.emplace("additionalFields", "dataSize,checksum");
+    query_params.emplace("billingTag", "Billing1234");
+
+    CancellationContext context;
```
andescu, line 45:
```suggestion
    // Can be used to cancel the ongoing request. Needs to be triggered 
    // from another thread as olp::client::OlpClient::CallApi() is a blocking call.
    CancellationContext context;
```

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+    olp::client::HttpResponse response = client.CallApi(
+        metadata_url, "GET", query_params, header_params, {}, nullptr, "", context);
+    ```
+
+5. Check the response and, if needed, parse JSON.
+
+    ```cpp
+    >if (response.GetStatus() != olp::http::HttpStatusCode::OK) {
```
andescu, line 53:
```suggestion
    if (response.GetStatus() != olp::http::HttpStatusCode::OK) {
        // Response is not OK, handle error cases.
```

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+    }
+    ```
+
+6. If an error occurs, print "Error", and then exit.
```
andescu, line 57: Remove, this is covered by the above proposed comment

docs/work-with-data-apis.md
```diff
@@ -0,0 +1,66 @@
...
+
+7. Get the response.
```
andescu, line 59: I would write: If the above check passed and the status of the received `olp::client::HttpResponse` is olp::http::HttpStatusCode::OK, i.e. 200 OK, then you can proceed with extracting the encoded JSON from the response and parsing it by using RapidJSON or any other JSON parsing library.

```cpp
    std::string response_json;
    response.GetResponse(response_json);
```

docs/work-with-data-apis.md
HalynaDumych: It's not a heading, but just a title/name of the example. I added it to visually and structurally differentiate the example from the other info. As a user, I don't always want to read the whole page when I need just the example.

docs/work-with-data-apis.md
HalynaDumych: Changed.

mykhailo-kuchma (approved):

LiubovDidkivska (approved):

PR #1192, opened 2021-04-07
Use SDK leveldb environment

mykhailo-kuchma:
When the max cache size is not set, SDK use the leveldb default env,
SDK needs to replace it with its own in order to disable mmap usage.
Fix the Limiter class constructor warning.

Relates-To: OLPSUP-13838


andescu (commented):
Please enhance commit message. Why is it needed? What is the issue? Etc.

andescu (approved):

jmoro3 (approved):

PR #1193, opened 2021-04-08
Do not remove protected keys in DefaultCache.

andescu:
When user calls DefaultCache::Remove() and
DefaultCache::RemoveKeysWithPrefix() we should not remove protected
keys. This is now implemented and tested.

Relates-To: OLPEDGE-2521

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DiskCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/DiskCache.cpp
@@ -384,34 +384,46 @@ DiskCache::OperationOutcome DiskCache::ApplyBatch(
 }
 
 bool DiskCache::RemoveKeysWithPrefix(const std::string& prefix,
-                                     uint64_t& removed_data_size) {
+                                     uint64_t& removed_data_size,
+                                     const RemoveFilterFunc& filter) {
   uint64_t data_size = 0u;
-  if (!database_) {
-    removed_data_size = 0u;
-    OLP_SDK_LOG_ERROR(kLogTag,
-                      "RemoveKeysWithPrefix: Database is uninitialized");
+  removed_data_size = 0u;
+
+  // As we remove data probably it is not wise to flood the leveldb memory cache
+  leveldb::ReadOptions opts;
+  opts.verify_checksums = check_crc_;
+  opts.fill_cache = false;
+  auto iterator = NewIterator(opts);
+  if (!iterator) {
+    OLP_SDK_LOG_WARNING(kLogTag,
+                        "RemoveKeysWithPrefix: Database is uninitialized");
     return false;
   }
 
   auto batch = std::make_unique<leveldb::WriteBatch>();
+  auto prefix_slice = leveldb::Slice(prefix);
+  const bool prefix_empty = prefix_slice.empty();
 
-  leveldb::ReadOptions opts;
-  opts.verify_checksums = check_crc_;
-  opts.fill_cache = true;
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -384,34 +384,44 @@ DiskCache::OperationOutcome DiskCache::ApplyBatch(
...
+
+  auto condition = [&](const leveldb::Iterator& it) {
+    return it.Valid() && (prefix.empty() || it.key().starts_with(prefix));
+  };
+
+  for (; condition(*iterator); iterator->Next()) {
+    auto key = iterator->key();
+
+    // Do not delete if protected
+    if (filter && filter(key.ToString())) {
```
andrey-kashcheev, line 419: I would prefer to use string view here instead of copy on each call (yes, we need to port it). Data and size are available in the Slice so should be pretty straightforward.

andescu: I don't think we need it as it is happening on remove either way and it is accepted to take long.

andescu: Sadly it is currently not possible either way as the protected list uses `std::set::lower_bound()` which in C++11 expects a string as an input so no chance to avoid copying. I tried locally to pass `leveldb::Slice` instead-but in the end it needs conversion to `std::string`. The only possibility is to refactor the ProtectedKeyList, which in my opinion can be postponed for later in case there are some real performance issues, which may bother some of our users.

mykhailo-kuchma (approved):

olp-cpp-sdk-core/src/cache/DiskCache.cpp
andrey-kashcheev: COW is no longer a go-to solution for compilers for copying strings as far as I know. And for SSO those strings are too long.
So yeah, let's see the results on the real data.

andrey-kashcheev (approved):

andrey-kashcheev (approved):

olp-cpp-sdk-core/tests/cache/DefaultCacheTest.cpp
```diff
@@ -263,40 +265,199 @@ TEST(DefaultCacheTest, MemSizeTest) {
...
+    ASSERT_TRUE(cache.Get("key3", decoder).empty());
+    ASSERT_TRUE(cache.Get("key5", decoder).empty());
+    ASSERT_TRUE(cache.Get("key7", decoder).empty());
+  }
+}
+
 TEST(DefaultCacheTest, RemoveWithPrefix) {
-  olp::cache::DefaultCache cache;
+  olp::cache::CacheSettings settings;
+  settings.max_memory_cache_size = 0;
```
kzvieriev, line 386: Why disable memory cache? It would be good to know protected keys are not removed from InMemoryCache too.

andescu: But there is no way to check if the data comes from the in-memory or from the mutable cache. Of course, it would make sense to only use in-memory and no mutable cache and vice-versa to guarantee that both ways it works. I can add another test to validate that the in-memory allone also works.

jmoro3 (dismissed):

kzvieriev (approved):

andrey-kashcheev (approved):

PR #1199, opened 2021-04-21
Add Protect/Release partition

LiubovDidkivska:
Add posibility to protect/release
partition by its id, which includes
metadata and data handle for partition.

Resolves: OAM-1024

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
@@ -409,8 +409,8 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    *
    * @param partition_id The partition ID that should be removed.
    *
-   * @note Before calling the API, specify a layer version. You can set it using
-   * the constructor or after the first online request.
+   * @note Before calling the API, specify a catalog version. You can set it
+   * using the constructor or after the first online request.
    *
    * @return True if partition data is removed successfully; false otherwise.
    */
@@ -421,8 +421,8 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    *
    * @param tile The tile key that should be removed.
    *
-   * @note Before calling the API, specify a layer version. You can set it using
-   * the constructor or after the first online request.
+   * @note Before calling the API, specify a catalog version. You can set it
+   * using the constructor or after the first online request.
    *
    * @return True if tile data is removed successfully; false otherwise.
    */
@@ -433,8 +433,8 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    *
    * @param partition_id The partition ID.
    *
-   * @note Before calling the API, specify a layer version. You can set it using
-   * the constructor or after the first online request.
+   * @note Before calling the API, specify a catalog version. You can set it
+   * using the constructor or after the first online request.
    *
    * @return True if the partition data is cached; false otherwise.
    */
@@ -447,8 +447,8 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    * @param aggregated The aggregated flag, used to specify whether the tile is
    * aggregated or not.
    *
-   * @note Before calling the API, specify a layer version. You can set it using
-   * the constructor or after the first online request.
+   * @note Before calling the API, specify a catalog version. You can set it
+   * using the constructor or after the first online request.
    *
    * @return True if the tile data is cached; false otherwise.
    */
@@ -463,14 +463,14 @@ class DATASERVICE_READ_API VersionedLayerClient final {
    * not expire. The quadtree stays protected if at least one tile key is
    * protected.
    *
+   * @note Before calling the API, specify a catalog version. You can set it
+   * using the constructor or after the first online request.
+   *
    * @note You can only protect tiles which data handles are present in the
    * cache at the time of the call.
    *
    * @note Please do not call `Protect` while the `Release` call for the same
    * catalog and layer is in progress.
-   *
-   * @note Before calling the API, specify a layer version. You can set it using
-   * the constructor or after the first online request.
    *
    * @param tiles The list of tile keys to be protected.
    *
@@ -480,6 +480,29 @@ class DATASERVICE_READ_API VersionedLayerClient final {
   bool Protect(const TileKeys& tiles);
 
   /**
+   * @brief Protect partition from eviction.
+   *
+   * Protecting partition means that its data and metadata keys are added to the
+   * protected list and stored in the cache. These keys are removed from the LRU
+   * cache, so they could not be evicted. Also, they do not expire.
+   *
+   * @note Before calling the API, specify a catalog version. You can set it
+   * using the constructor or after the first online request.
+   *
+   * @note You can only protect partitions which data handles are present in the
+   * cache at the time of the call.
+   *
+   * @note Please do not call `Protect` while the `Release` call for the same
+   * catalog and layer is in progress.
+   *
+   * @param partition_id Partition id to be protected.
+   *
+   * @return True if partition keys were successfully added to the protected
+   * list; false otherwise.
+   */
+  bool Protect(const std::string& partition_id);
+
+  /**
    * @brief Removes a list of tiles from protection.
    *
    * Releasing tile keys removes data and quadtree keys from the protected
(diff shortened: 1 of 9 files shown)


olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
```diff
@@ -500,6 +500,52 @@ class DATASERVICE_READ_API VersionedLayerClient final {
...
+   *
+   * @note Before calling the API, specify a layer version. You can set it using
+   * the constructor or after the first online request.
+   *
+   * @param partition_id Partition id to be protected.
+   *
+   * @return True if some keys were successfully added to the protected list;
+   * false otherwise.
+   */
+  bool Protect(const std::string& partition_id);
```
andescu, line 525: I would group all `Protect` and all `Release` methods and not provide them in pairs.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -790,6 +790,53 @@ bool VersionedLayerClientImpl::Release(const TileKeys& tiles) {
 
   return settings_.cache->Release(keys_to_release);
 }
+cache::KeyValueCache::KeyListType
+VersionedLayerClientImpl::GetRelatedPartitionKeys(
```
andescu, line 794: Do you really need an extra method for this? Seems pretty basic and can be integrated into `Protect`

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -790,6 +790,53 @@ bool VersionedLayerClientImpl::Release(const TileKeys& tiles) {
...
+    const std::string& partition_id) {
+  if (!settings_.cache) {
+    return {};
+  }
+
+  auto version = catalog_version_.load();
```
andescu, line 800: This is a little wrong, as the version can be currently in the request phase so it will be soon available. Probably you should use `GetVersion` instead.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -790,6 +790,53 @@ bool VersionedLayerClientImpl::Release(const TileKeys& tiles) {
...
+  if (version == kInvalidVersion) {
+    OLP_SDK_LOG_WARNING(kLogTag,
+                        "Method Protect failed, version is not initialized");
+    return {};
+  }
+
+  auto cache = settings_.cache;
```
andescu, line 807: Why copy?

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -790,6 +790,53 @@ bool VersionedLayerClientImpl::Release(const TileKeys& tiles) {
...
+
+  repository::PartitionsCacheRepository partitions_repo(catalog_, layer_id_,
+                                                        cache);
+  std::string handle;
+  if (partitions_repo.GetPartitionHandle(partition_id, version, handle)) {
+    repository::DataCacheRepository data_repo(catalog_, cache);
+
+    return cache::KeyValueCache::KeyListType{
+        partitions_repo.CreatePartitionKey(partition_id, version),
+        data_repo.CreateKey(layer_id_, handle)};
```
andescu, line 817: Maybe instead of this, we should add the protect/release functionality to the repositories instead.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
LiubovDidkivska: Should I also change it for Protect/Release for tile keys, IsCached, RemoveFromCache, etc.?

andescu: As these are all sync methods it is a little tricky. If we followed the same approach for all then we need to provide a good comment. So probably it is wiser to keep in sync with the rest and see if somebody will complain that this behaviour is indefinite. Instead please add a proper comment for this in the public API to this new APIs and to all related APIs. It is catalog version what is needed and not layer version. Also we need to specify by when it can be used, means that the version is set internally once one of the async APIs are called, e.g. `GetData`, `GetPartitions`, etc.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
LiubovDidkivska: it is just common code, which the same for protect and release

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/VersionedLayerClient.h
```diff
@@ -500,6 +500,52 @@ class DATASERVICE_READ_API VersionedLayerClient final {
...
+   * Releasing partition id removes data handle and metadata keys from the
+   * protected list. The keys are added to the LRU cache, so they could be
+   * evicted. Expiration value is restored, and keys can expire. The quadtree
+   * can be removed from the protected list if all tile keys are no longer
+   * protected.
+   *
+   * @note Please make sure that `Protect` will not be called for the same
+   * catalog and layer while the `Release` call is in progress.
+   *
+   * @note Before calling the API, specify a layer version. You can set it using
```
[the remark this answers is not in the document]
LiubovDidkivska: As we discussed below, keeping this comment.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
LiubovDidkivska: For all sync methods keeping the same behavior, that version should be set in constructor, or offline request.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -791,6 +791,42 @@ bool VersionedLayerClientImpl::Release(const TileKeys& tiles) {
...
+  }
+
+  auto version = catalog_version_.load();
+  if (version == kInvalidVersion) {
+    OLP_SDK_LOG_WARNING(kLogTag,
+                        "Method Protect failed, version is not initialized");
+    return {};
+  }
+
+  repository::PartitionsCacheRepository partitions_repo(catalog_, layer_id_,
```
andescu, line 806: We should not use acronyms.

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -791,6 +791,42 @@ bool VersionedLayerClientImpl::Release(const TileKeys& tiles) {
...
+
+bool VersionedLayerClientImpl::Release(const std::string& partition_id) {
+  if (!settings_.cache) {
+    return {};
+  }
+
+  auto version = catalog_version_.load();
+  if (version == kInvalidVersion) {
+    OLP_SDK_LOG_WARNING(kLogTag,
+                        "Method Protect failed, version is not initialized");
```
andescu, line 820:
```suggestion
                        "Method Release failed, version is not initialized");
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
```diff
@@ -327,6 +333,40 @@ bool PartitionsCacheRepository::ContainsTree(
...
+        CreateKey(catalog_, layer_id_, partition_id, version),
+        CreateKey(catalog_, layer_id_, handle)};
+  }
+
+  return {};
+}
+
+bool PartitionsCacheRepository::Protect(
+    const std::string& partition_id, const boost::optional<int64_t>& version) {
+  const auto& keys = GetRelatedPartitionKeys(partition_id, version);
```
andescu, line 352:
```suggestion
  const auto keys = GetRelatedPartitionKeys(partition_id, version);
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
```diff
@@ -327,6 +333,40 @@ bool PartitionsCacheRepository::ContainsTree(
...
+  if (keys.empty()) {
+    return false;
+  }
+
+  return cache_->Protect(keys);
+}
+
+bool PartitionsCacheRepository::Release(
+    const std::string& partition_id, const boost::optional<int64_t>& version) {
+  const auto& keys = GetRelatedPartitionKeys(partition_id, version);
```
andescu, line 362:
```suggestion
  const auto keys = GetRelatedPartitionKeys(partition_id, version);
```

olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
```diff
@@ -327,6 +333,40 @@ bool PartitionsCacheRepository::ContainsTree(
   return cache_->Contains(CreateQuadKey(key, depth, version));
 }
 
+cache::KeyValueCache::KeyListType
+PartitionsCacheRepository::GetRelatedPartitionKeys(
```
andescu, line 337: It is rather a `CreatePartitionKeys()` as you create them rather than getting them from the cache.

andescu (approved):

jmoro3 (approved):

PR #1203, opened 2021-04-30
Mark inoperable cache as corrupted

andrey-kashcheev:
Cache could become inoperable due to several reasons:
- interrupted compaction leads to files on level-0, and this leads to
excessive memory usage during read operations;
- different IO errors for protected cache couldn't be solved with RepairDb
due to read-only nature of this cache (for mutable cache there will be
an attempt to repair it);

Do not remove corrupted cache and just notify a user about it.

Relates-To: OLPSUP-14088

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/cache/DefaultCache.h
+++ b/olp-cpp-sdk-core/include/olp/core/cache/DefaultCache.h
@@ -60,9 +60,10 @@ class CORE_API DefaultCache : public KeyValueCache {
    * @brief The storage open result type.
    */
   enum StorageOpenResult {
-    Success,             /*!< The operation succeeded. */
-    OpenDiskPathFailure, /*!< The disk cache failure. */
-    NotReady             /*!< The DefaultCache is closed. */
+    Success,                 /*!< The operation succeeded. */
+    OpenDiskPathFailure,     /*!< The disk cache failure. */
+    ProtectedCacheCorrupted, /*!< The protected disk cache is corrupted. */
+    NotReady                 /*!< The DefaultCache is closed. */
   };
 
   /**
@@ -140,6 +141,10 @@ class CORE_API DefaultCache : public KeyValueCache {
    * operation in parallel for the time of the compacting operation. Be aware
    * that automatic asynchronous compacting operation is triggered internally
    * once the database size exceeds the CacheSettings::max_disk_storage size.
+   *
+   * @note After the compaction is finished the cache is checked on level-0 file
+   * presence. If there are still some files present another round of compaction
+   * is performed.
    */
   void Compact();
 
--- a/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
+++ b/olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
@@ -31,6 +31,7 @@
 
 namespace {
 using CacheType = olp::cache::DefaultCache::CacheType;
+using StorageOpenResult = olp::cache::DefaultCache::StorageOpenResult;
 
 constexpr auto kLogTag = "DefaultCache";
 constexpr auto kExpirySuffix = "::expiry";
@@ -118,6 +119,21 @@ int64_t GetElapsedTime(std::chrono::steady_clock::time_point start) {
 
 bool IsInternalKey(const std::string& key) {
   return key.find(kInternalKeysPrefix) == 0u;
+}
+
+olp::cache::DefaultCache::StorageOpenResult ToStorageOpenResult(
+    olp::cache::OpenResult input) {
+  switch (input) {
+    case olp::cache::OpenResult::Fail:
+      return StorageOpenResult::OpenDiskPathFailure;
+    case olp::cache::OpenResult::Corrupted:
+      return StorageOpenResult::ProtectedCacheCorrupted;
+    case olp::cache::OpenResult::Repaired:
+    case olp::cache::OpenResult::Success:
+      return StorageOpenResult::Success;
+  }
+
+  return {};
 }
 }  // namespace
 
@@ -817,13 +833,13 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::SetupProtectedCache() {
   auto status = protected_cache_->Open(
       settings_.disk_path_protected.get(), settings_.disk_path_protected.get(),
       protected_storage_settings, OpenOptions::ReadOnly);
-  if (status == OpenResult::Fail) {
-    OLP_SDK_LOG_ERROR_F(kLogTag, "Failed to reopen protected cache %s",
+  if (status != OpenResult::Success) {
+    OLP_SDK_LOG_ERROR_F(kLogTag, "Failed to open protected cache %s",
                         settings_.disk_path_protected.get().c_str());
 
     protected_cache_.reset();
     settings_.disk_path_protected = boost::none;
-    return DefaultCache::OpenDiskPathFailure;
+    return ToStorageOpenResult(status);
   }
 
   return DefaultCache::Success;
@@ -836,13 +852,15 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::SetupMutableCache() {
   auto status = mutable_cache_->Open(settings_.disk_path_mutable.get(),
                                      settings_.disk_path_mutable.get(),
                                      storage_settings, OpenOptions::Default);
-  if (status == OpenResult::Fail) {
+  if (status == OpenResult::Repaired) {
+    OLP_SDK_LOG_INFO(kLogTag, "Mutable cache was repaired");
+  } else if (status == OpenResult::Fail) {
     OLP_SDK_LOG_ERROR_F(kLogTag, "Failed to open the mutable cache %s",
                         settings_.disk_path_mutable.get().c_str());
 
     mutable_cache_.reset();
     settings_.disk_path_mutable = boost::none;
-    return DefaultCache::OpenDiskPathFailure;
+    return StorageOpenResult::OpenDiskPathFailure;
   }
 
   // read protected keys
--- a/olp-cpp-sdk-core/src/cache/DiskCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/DiskCache.cpp
@@ -42,6 +42,7 @@ namespace cache {
 
 namespace {
 constexpr auto kLogTag = "DiskCache";
+constexpr auto kLevelDbLostFolder = "/lost";
 
 leveldb::Slice ToLeveldbSlice(const std::string& slice) {
   return leveldb::Slice(slice);
@@ -52,7 +53,13 @@ static bool RepairCache(const std::string& data_path) {
   auto status = leveldb::RepairDB(data_path, leveldb::Options());
   if (status.ok()) {
     OLP_SDK_LOG_INFO(kLogTag, "RepairCache: repaired - " << data_path);
-    DiskCacheEnv::Env()->DeleteDir(data_path + "/lost");
+    const auto lost_folder_path = data_path + kLevelDbLostFolder;
+    if (utils::Dir::Exists(lost_folder_path)) {
+      OLP_SDK_LOG_INFO_F(
+          kLogTag, "RepairCache: some data may have been lost - deleting '%s'",
+          kLevelDbLostFolder);
+      utils::Dir::Remove(lost_folder_path);
+    }
     return true;
   }
   OLP_SDK_LOG_ERROR(kLogTag,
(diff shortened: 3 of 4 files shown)


olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -236,7 +246,13 @@ OpenResult DiskCache::Open(const std::string& data_path,
   }
 
   database_.reset(db);
-  return OpenResult::Success;
+
+  if (!CheckDbState()) {
```
andescu, line 250: This should be only done in r/o mode. The mutable cache is opened and closes in all kinds of states and compaction is working there as it has a r/w env.

olp-cpp-sdk-core/include/olp/core/cache/DefaultCache.h
```diff
@@ -62,6 +62,7 @@ class CORE_API DefaultCache : public KeyValueCache {
   enum StorageOpenResult {
     Success,             /*!< The operation succeeded. */
     OpenDiskPathFailure, /*!< The disk cache failure. */
+    CacheCorrupted,      /*!< The disk cache is corrupted. */
```
andescu, line 65: I think we should differentiate here. I would name it `ProtectedCacheCorrupted` as we are not interested in mutable cache corruption as that one can heal itself without any issues.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -119,6 +119,23 @@ int64_t GetElapsedTime(std::chrono::steady_clock::time_point start) {
 bool IsInternalKey(const std::string& key) {
   return key.find(kInternalKeysPrefix) == 0u;
 }
+
+olp::cache::DefaultCache::StorageOpenResult ToStorageOpenResult(
+    olp::cache::OpenResult src) {
+  using StorageOpenResult = olp::cache::DefaultCache::StorageOpenResult;
```
andescu, line 125: Move up to the global scope.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -119,6 +119,23 @@ int64_t GetElapsedTime(std::chrono::steady_clock::time_point start) {
 bool IsInternalKey(const std::string& key) {
   return key.find(kInternalKeysPrefix) == 0u;
 }
+
+olp::cache::DefaultCache::StorageOpenResult ToStorageOpenResult(
+    olp::cache::OpenResult src) {
```
andescu, line 124:
```suggestion
    olp::cache::OpenResult input) {
```

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -817,13 +834,13 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::SetupProtectedCache() {
   auto status = protected_cache_->Open(
       settings_.disk_path_protected.get(), settings_.disk_path_protected.get(),
       protected_storage_settings, OpenOptions::ReadOnly);
-  if (status == OpenResult::Fail) {
+  if (status == OpenResult::Fail || status == OpenResult::Corrupted) {
     OLP_SDK_LOG_ERROR_F(kLogTag, "Failed to reopen protected cache %s",
```
andescu, line 838: "Failed to reopen..."? Is this a copy/paste error? It should be "Failed to open... "

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -836,13 +853,13 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::SetupMutableCache() {
   auto status = mutable_cache_->Open(settings_.disk_path_mutable.get(),
                                      settings_.disk_path_mutable.get(),
                                      storage_settings, OpenOptions::Default);
-  if (status == OpenResult::Fail) {
+  if (status == OpenResult::Fail || status == OpenResult::Corrupted) {
```
andescu, line 856: I don't think we need this for the mutable cache. I would though add a log in case we repaired it and there is a `lost/` folder present so that users may know that some data was lost. Maybe also delete the lost folder as this is extra disk space that is used.

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -216,13 +216,23 @@ OpenResult DiskCache::Open(const std::string& data_path,
...
+  if (status.IsCorruption()) {
+    OLP_SDK_LOG_ERROR_F(kLogTag, "Cache is corrupted, status=%s",
+                        status.ToString().c_str());
+    return OpenResult::Corrupted;
+  }
+
+  bool repaired = false;
+  if (status.IsIOError()) {
+    if (is_read_only) {
+      OLP_SDK_LOG_ERROR_F(kLogTag, "Cache couldn't be opened, status=%s",
```
andescu, line 228:
```suggestion
      OLP_SDK_LOG_ERROR_F(kLogTag, "Cache couldn't be opened, status='%s'",
```

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -216,13 +216,23 @@ OpenResult DiskCache::Open(const std::string& data_path,
...
   }
 
-  // If the database is r/w and corrupted, attempt to repair & reopen
-  if ((status.IsCorruption() || status.IsIOError()) &&
-      RepairCache(versioned_data_path) == true) {
-    status = leveldb::DB::Open(open_options, versioned_data_path, &db);
-    if (status.ok()) {
-      database_.reset(db);
-      return OpenResult::Repaired;
+  if (status.IsCorruption()) {
```
andescu, line 219: In the case of r/w database, which means mutable cache, we should not just abort. The mutable cache is an active cache and can be repaired.

olp-cpp-sdk-core/src/cache/DefaultCacheImpl.cpp
```diff
@@ -817,13 +833,13 @@ DefaultCache::StorageOpenResult DefaultCacheImpl::SetupProtectedCache() {
   auto status = protected_cache_->Open(
       settings_.disk_path_protected.get(), settings_.disk_path_protected.get(),
       protected_storage_settings, OpenOptions::ReadOnly);
-  if (status == OpenResult::Fail) {
-    OLP_SDK_LOG_ERROR_F(kLogTag, "Failed to reopen protected cache %s",
+  if (status == OpenResult::Fail || status == OpenResult::Corrupted) {
```
andescu, line 836: Why not `if (status != OpenResult::Success)`

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -52,7 +52,7 @@ static bool RepairCache(const std::string& data_path) {
   auto status = leveldb::RepairDB(data_path, leveldb::Options());
   if (status.ok()) {
     OLP_SDK_LOG_INFO(kLogTag, "RepairCache: repaired - " << data_path);
-    DiskCacheEnv::Env()->DeleteDir(data_path + "/lost");
+    utils::Dir::Remove(data_path + "/lost");
```
andescu, line 55: Please move "lost/" to a constexpr.
I would not let this silently go. Instead, please log that there was a lost folder and we will delete it.
```cpp
if (utils::Dir::Exists(data_path + kLeveldbLostFolder)) {
    OLP_SDK_LOG_WARNING(kLogTag, "RepairCache: data has been lost - deleting '" << kLeveldbLostFolder << "' folder");
    utils::Dir::Remove(data_path + "/lost");
}
```

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -120,6 +120,12 @@ client::ApiError GetApiError(const leveldb::Status& status) {
   return client::ApiError(code, status.ToString());
 }
 
+int GetFilesNumberAtLevel0(leveldb::DB& db) {
```
andescu, line 123: Why not `CheckCompactionFinished` or `ContinueCompaction`, as we are not really interested in the number of L0 or L1 files but if we should keep up the compaction. Also, you can move the log here. Later if we discover that it is not enough and also L1 files could be a problem we, can easily extend this method with another check instead of adding another method with another check below.

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -153,8 +159,19 @@ void DiskCache::Compact() {
   // doing it already. We don't need two at the same time.
   if (database_ && !compacting_.exchange(true)) {
     OLP_SDK_LOG_INFO(kLogTag, "Compact: Compacting database started");
-    database_->CompactRange(nullptr, nullptr);
+
+    while (true) {
```
andescu, line 163: I would use do/while instead:

```cpp
do {
 ...
} while (ContinueCompaction(*database_));

File: olp-cpp-sdk-core/src/cache/DiskCache.cpp [hunk 13]
```diff
@@ -153,8 +159,19 @@ void DiskCache::Compact() {
...
+      database_->CompactRange(nullptr, nullptr);
+      const auto num_files_at_level0 = GetFilesNumberAtLevel0(*database_);
+      if (num_files_at_level0 == 0) {
+        break;
+      }
+      OLP_SDK_LOG_INFO_F(kLogTag, "Compact: Files on level-0, num=%d",
```
andescu, line 169: Move to method `ContinueCompaction` or however, you name it.

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -216,13 +233,17 @@ OpenResult DiskCache::Open(const std::string& data_path,
...
-  if ((status.IsCorruption() || status.IsIOError()) &&
-      RepairCache(versioned_data_path) == true) {
-    status = leveldb::DB::Open(open_options, versioned_data_path, &db);
-    if (status.ok()) {
-      database_.reset(db);
-      return OpenResult::Repaired;
+  bool repaired = false;
+  if (status.IsCorruption() || status.IsIOError()) {
+    if (is_read_only) {
+      OLP_SDK_LOG_ERROR_F(kLogTag, "Open: cache corrupted, error='%s'",
```
andescu, line 239: Maybe also add path of the cache?

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -235,8 +256,18 @@ OpenResult DiskCache::Open(const std::string& data_path,
     error_ = NoError{};
   }
 
+  if (is_read_only) {
+    const auto num_files_at_level0 = GetFilesNumberAtLevel0(*db);
```
andescu, line 260: Same here. Call ContinueCompaction() method. It is not really important for the user to know how many L0 files there are, he only needs to know that compaction is not finished and he should compact the db before handing it over to us.

olp-cpp-sdk-core/src/cache/DiskCache.cpp
andrey-kashcheev: I'm not sure about the 'data has been lost' part. This may be misleading as it's not guaranteed that any data was lost and vice versa.

andescu: Then "RepairCache: data may have been lost ...."

olp-cpp-sdk-core/src/cache/DiskCache.cpp
andrey-kashcheev: The cache path will be reported by DefaultCacheImp, should we duplicate it here?

andescu: I would not trust anybody. Every component logs for itself. It is not a issue to have duplicate data for such important logs.

olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -120,6 +127,20 @@ client::ApiError GetApiError(const leveldb::Status& status) {
...
+int CheckCompactionFinished(leveldb::DB& db) {
+  std::string property_result;
+  db.GetProperty("leveldb.num-files-at-level0", &property_result);
+
+  const auto files_at_level0 = std::stoi(property_result);
+  if (files_at_level0 == 0) {
+    return true;
+  }
+
+  OLP_SDK_LOG_INFO_F(kLogTag, "Files on level-0, num=%d", files_at_level0);
```
andescu, line 139:
```suggestion
  OLP_SDK_LOG_INFO_F(kLogTag, "CheckCompactionFinished: L0 files present, files_at_level0=%d", files_at_level0);
```

andescu (approved):

leonid-ziskel (approved):

PR #1210, opened 2021-05-24
Add cache compaction to protected cache example.

kzvieriev:
If protected cache contains level 0 files leveldb will consume to
much RAM while trying to compact in read env. To prevent such
situations level 0 files treatened as error. Thus, protected cache
example must be updated.

Relates-To: OLPSUP-14367

### DIFF
--- a/examples/ProtectedCacheExample.cpp
+++ b/examples/ProtectedCacheExample.cpp
@@ -20,7 +20,7 @@
 
 #include <olp/authentication/TokenProvider.h>
 #include <olp/core/cache/CacheSettings.h>
-#include <olp/core/cache/KeyValueCache.h>
+#include <olp/core/cache/DefaultCache.h>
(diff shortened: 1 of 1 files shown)


jmoro3 (dismissed):

examples/ProtectedCacheExample.cpp
```diff
@@ -122,6 +126,11 @@ int RunExampleReadWithCache(const AccessKey& access_key,
   olp::dataservice::read::DataResponse data_response =
     future.GetFuture().get();
 
+  // Compact mutable cache, so it can be used as protected cache
+  if (cache_settings.disk_path_mutable.is_initialized()) {
```
andescu, line 130: I don't think you need to test this, as it is already tested internally.

andescu (approved):

mykhailo-kuchma (approved):

PR #1218, opened 2021-05-28
Handling of NSURLErrorNetworkConnectionLost error

mykhailo-kuchma:
Enable the handling of NSURLErrorNetworkConnectionLost network error

Resolves: OLPEDGE-2566

### DIFF
--- a/olp-cpp-sdk-core/src/http/ios/OLPNetworkIOS.mm
+++ b/olp-cpp-sdk-core/src/http/ios/OLPNetworkIOS.mm
@@ -99,6 +99,8 @@ olp::http::ErrorCode ConvertNSURLErrorToNetworkErrorCode(NSInteger error_code) {
       return ErrorCode::OFFLINE_ERROR;
     case NSURLErrorTimedOut:
       return ErrorCode::TIMEOUT_ERROR;
+    case NSURLErrorNetworkConnectionLost:
+      return ErrorCode::IO_ERROR;
     default:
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/src/http/ios/OLPNetworkIOS.mm
```diff
@@ -98,6 +98,7 @@
     case NSURLErrorDataNotAllowed:
       return ErrorCode::OFFLINE_ERROR;
     case NSURLErrorTimedOut:
+    case NSURLErrorNetworkConnectionLost:
```
andescu, line 101: Is it not rather OFFLINE_ERROR?

mykhailo-kuchma: A client or server connection was severed in the middle of an in-progress load.

https://developer.apple.com/documentation/foundation/1508628-url_loading_system_error_codes/nsurlerrornetworkconnectionlost

andescu: Yes, but this is not a time out, rather a connection lost. Maybe add a new error code to `http::ErrorCode` enum?

mykhailo-kuchma: @andescu, I checked other implementation, the error code should be IO_ERROR

ystefinko (approved):

PR #1219, opened 2021-06-01
Add the Equal Jitter backdown strategy

mykhailo-kuchma:
To avoid requests sent after 0 (or close to 0) milliseconds, backdown
strategy changed to support the Equal Jitter, as described in
https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

Resolves: OLPEDGE-2574

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/client/BackdownStrategy.h
+++ b/olp-cpp-sdk-core/include/olp/core/client/BackdownStrategy.h
@@ -28,12 +28,13 @@ namespace olp {
 namespace client {
 
 /**
- * @brief Computes wait time for the next retry attempt via
- * the exponential backoff with the added jitter.
+ * @brief Computes wait time for the next retry attempt via the exponential
+ * backoff with the added jitter.
  *
  * This backoff strategy is based on the exponential wait-time approach.
  * For example, when the wait time exponentially grows with each retry attempt,
- * but randomization is added.
+ * but randomization is added. See
+ * https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
  *
  * The actual formula can be described in the following way:
  * @code{.unparsed}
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/include/olp/core/client/BackdownStrategy.h
```diff
@@ -64,5 +64,54 @@ struct CORE_API ExponentialBackdownStrategy {
   }
 };
 
+/**
+ * @brief Computes wait time for the next retry attempt via
+ * the exponential backoff with the added jitter.
+ *
+ * This backoff strategy is based on the Equal Jitter approach. See
+ * https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
```
andescu, line 72: Please also add this link to the above ExponentialBackdownStrategy class Doxygen.

olp-cpp-sdk-core/include/olp/core/client/BackdownStrategy.h
```diff
@@ -64,5 +64,54 @@ struct CORE_API ExponentialBackdownStrategy {
...
+ * @endcode
+ */
+struct CORE_API EqualJitterBackdownStrategy {
+ public:
+  /**
+   * @brief Creates a EqualJitterBackdownStrategy instance.
+   *
+   * @param min_cap The minimum cap (ms) used in the wait time formula.
+   */
+  explicit EqualJitterBackdownStrategy(std::chrono::milliseconds cap)
```
andescu, line 87: I would not do it in the constructor for the simple reason as the user does not know what the cap should be and he will be tempted to set the max_wait_time here which is to be considered only after all the retries are done. The cap itself is in my opinion variable and therefore more suitable to be added to the operator() as it will be adapted by the OlpClient based on the current retry and what time is left from the previous retries to the max_wait_time.

olp-cpp-sdk-core/include/olp/core/client/BackdownStrategy.h
```diff
@@ -64,5 +64,54 @@ struct CORE_API ExponentialBackdownStrategy {
...
+   *
+   * @return The timeout for the next retry attempt.
+   */
+  std::chrono::milliseconds operator()(
+      std::chrono::milliseconds initial_backdown_period, size_t retry_count) {
+    const auto exponential_wait_time =
+        initial_backdown_period.count() * (1ll << retry_count);
+
+    static thread_local std::mt19937 kGenerator(std::random_device{}());
+    const auto temp = std::min<long long>(cap_.count(), exponential_wait_time);
```
andescu, line 105:
```suggestion
    const auto temp = std::min<int64_t>(cap_.count(), exponential_wait_time);
```

olp-cpp-sdk-core/tests/client/BackdownStrategyTest.cpp
```diff
@@ -0,0 +1,39 @@
...
+ * License-Filename: LICENSE
+ */
+
+#include <gtest/gtest.h>
+
+#include <olp/core/client/OlpClientSettings.h>
+
+namespace {
+
+TEST(BackdownStrategy, Assign) {
```
andescu, line 26: This is not enough. Please test that the count is growing and it does not exceed the `cap`.

olp-cpp-sdk-core/include/olp/core/client/BackdownStrategy.h
mykhailo-kuchma: We can't change the operator() signature since it is a breaking change.

andescu (approved):

meravingen123 (approved):

PR #1220, opened 2021-06-01
Free an WinHttp request after an error.

jstille:
It seems that in NetworkWinHttp, the previous code only sets the "complete" flag on the Request when all data has been received (line 786). And only with the "completed" flag, the Request gets closed (line 962). So after an error, the request is never completed and thus blocked forever.

Trying to fix this by replacing the unused "cancelling" flag by an "error" flag and then harvesting requests in error state as well.

This has been tested manually up to now only for the case of connecting to an unused port of an existing server (where we get positive information that connection is not possible).


jmoro3:
Thank you for contributing.
Please be aware that currently bots are expecting all commits are signed off.
Here is steps what steps are needed to sign a commit off: https://github.com/heremaps/here-data-sdk-cpp/pull/1220/checks?check_run_id=2720063786

andescu (approved):

jmoro3 (approved):

PR #1228, opened 2021-06-10
Disable db repair for protected cache.

kzvieriev:
Now protected cache is opened with rw permissions by default. Repair
must be prohibited, so the user could know the data is corrupted.

Relates-To: OAM-1088

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DiskCache.h
+++ b/olp-cpp-sdk-core/src/cache/DiskCache.h
@@ -111,7 +111,8 @@ class DiskCache {
   ~DiskCache();
   OpenResult Open(const std::string& data_path,
                   const std::string& versioned_data_path,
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/src/cache/DiskCache.h
```diff
@@ -111,7 +111,8 @@ class DiskCache {
   ~DiskCache();
   OpenResult Open(const std::string& data_path,
                   const std::string& versioned_data_path,
-                  StorageSettings settings, OpenOptions options);
+                  StorageSettings settings, OpenOptions options,
+                  bool can_repair_db = true);
```
andescu, line 115:
```suggestion
                  bool repair_if_broken = true);
```

andescu (approved):

jmoro3 (approved):

PR #1232, opened 2021-07-01
Wrap the leveldb::Env to shared_ptr

mykhailo-kuchma:
Change the static EnvWrapper instance to a wrapped with a smart
pointer, so that we can guaranty that the instance will be alive while
it is needed, even when the `env` was deinitialized at exit.
By this we fix the scenario when the environment is destroyed before
DiskCache in certain use cases.

Resolves: OLPSUP-14933

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DiskCacheEnv.cpp
+++ b/olp-cpp-sdk-core/src/cache/DiskCacheEnv.cpp
@@ -196,12 +196,18 @@ class EnvWrapper : public leveldb::EnvWrapper {
 namespace olp {
 namespace cache {
 
-leveldb::Env* DiskCacheEnv::Env() {
+std::shared_ptr<leveldb::Env> DiskCacheEnv::CreateEnv() {
 #ifdef PORTING_PLATFORM_WINDOWS
-  return leveldb::Env::Default();
+  // return the normal env as a shared ptr with empty deleter
+  return std::shared_ptr<leveldb::Env>(leveldb::Env::Default(),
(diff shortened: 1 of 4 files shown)


olp-cpp-sdk-core/src/cache/DiskCacheEnv.cpp
```diff
@@ -196,12 +196,18 @@ class EnvWrapper : public leveldb::EnvWrapper {
...
 #ifdef PORTING_PLATFORM_WINDOWS
-  return leveldb::Env::Default();
+  // return the normal env as a shared ptr with empty deleter
+  return std::shared_ptr<leveldb::Env>(leveldb::Env::Default(),
+                                       [](leveldb::Env*) {});
 #else
-  static EnvWrapper env;
-  return &env;
+  // Static shared_ptr is needed to share the Limiter variable, and to avoid the
+  // scenario when the env is destoryed at exit while the disk cache instance is
```
andescu, line 206: Typo
```suggestion
  // scenario when the env is destroyed at exit while the disk cache instance is
```

andescu (approved):

andrey-kashcheev (commented):
Just curious: how is it possible that some non-static object outlives static object? Destruction will be done during library closing or after the main function. Am I missing something?

mykhailo-kuchma:
> Just curious: how is it possible that some non-static object outlives static object? Destruction will be done during library closing or after the main function. Am I missing something?

The user has another static object that has an opened cache, it is destroyed after.

jmoro3 (approved):

PR #1237, opened 2021-07-19
Add missed includes.

jmoro3:
Some files that use for_each and transform have missed <algorithm>
in includes.

Relates-To: OLPEDGE-2614

Signed-off-by: Iuliia Moroz <ext-iuliia.moroz@here.com>

### DIFF
--- a/tests/common/ResponseGenerator.cpp
+++ b/tests/common/ResponseGenerator.cpp
@@ -19,6 +19,7 @@
 
 #include "ResponseGenerator.h"
 
+#include <algorithm>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


mykhailo-kuchma (approved):

mykhailo-kuchma (approved):

tests/common/ResponseGenerator.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include "ResponseGenerator.h"
 
+#include <algorithm>
 #include <string>
 #include <vector>
```
andrey-kashcheev, line 24: It would be nice to clean excess inclusions too (e.g. '<vector>' is a quite massive one but clearly not needed here). Of course as a separate ticket.

andrey-kashcheev (approved):

tests/common/ResponseGenerator.cpp
jmoro3: Agree. But to clear obsolete includes will be a bit more tricky. Make sense to add a ticket and do it lately

PR #1243, opened 2021-09-07
Remove pessimizing move (as complained about by GCC)

rbuch703:
std::move()'ing a local variable prevents copy elision, which would be
faster than std::move() itself.


andescu:
Included in #1287. Closing this as obsolete.

PR #1249, opened 2021-09-21
Extend NamedMutex with an ability to share an ApiError among threads

mykolamalik:
It is done to prevent request repetition if the same request
in other thread have ended up with an error.
This change fixes an issue when task scheduler is overloaded
by the same requests that fails.

Relates-To: OLPSUP-15885

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
@@ -162,10 +162,25 @@ BlobApi::DataResponse DataRepository::GetBlobData(
     }
   }
 
+  // Check if other threads have faced an error.
+  const auto optional_error = mutex.GetError();
+  if (optional_error) {
+    OLP_SDK_LOG_DEBUG_F(kLogTag,
+                        "Found error in NamedMutex, aborting, hrn='%s', "
+                        "key='%s', error='%s'",
+                        catalog_.ToCatalogHRNString().c_str(),
+                        data_handle->c_str(),
+                        optional_error->GetMessage().c_str());
+    return *optional_error;
+  }
+
   auto storage_api_lookup = lookup_client_.LookupApi(
       service, "v1", static_cast<client::FetchOptions>(fetch_option), context);
 
   if (!storage_api_lookup.IsSuccessful()) {
+    // Store an error to share it with other threads.
+    mutex.SetError(storage_api_lookup.GetError());
+
     return storage_api_lookup.GetError();
   }
 
@@ -204,6 +219,9 @@ BlobApi::DataResponse DataRepository::GetBlobData(
           catalog_.ToCatalogHRNString().c_str(), data_handle->c_str());
       repository.Clear(layer, data_handle.value());
     }
+
+    // Store an error to share it with other threads.
+    mutex.SetError(storage_response.GetError());
   }
 
   return storage_response;
--- a/olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 HERE Europe B.V.
+ * Copyright (C) 2020-2021 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,11 +30,14 @@ class NamedMutexStorage::Impl {
  public:
   std::mutex& AquireLock(const std::string& resource);
   void ReleaseLock(const std::string& resource);
+  void SetError(const std::string& resource, const client::ApiError& error);
+  boost::optional<client::ApiError> GetError(const std::string& resource);
 
  private:
   struct RefCounterMutex {
     std::mutex mutex;
     uint32_t use_count{0u};
+    boost::optional<client::ApiError> optional_error;
   };
 
   std::mutex mutex_;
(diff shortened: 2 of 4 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 HERE Europe B.V.
+ * Copyright (C) 2020-2001 HERE Europe B.V.
```
andrey-kashcheev, line 2: Back to the future :)

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
```diff
@@ -162,10 +162,25 @@ BlobApi::DataResponse DataRepository::GetBlobData(
     }
   }
 
+  // Check if other threads have faced an error.
+  const auto optional_error = mutex.get_error();
+  if (optional_error) {
+    OLP_SDK_LOG_DEBUG_F(kLogTag,
+                        "This blob download ended up with an error in other "
+                        "thread, hrn='%s', key='%s', error='%s'",
```
andrey-kashcheev, line 170: Bit of a mouthful for a log message

olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
mykolamalik: Thanks! Will be updated in the next revision

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
mykolamalik: That is why this is debug log and not warning :)

olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
```diff
@@ -30,11 +30,14 @@ class NamedMutexStorage::Impl {
...
   std::mutex& AquireLock(const std::string& resource);
   void ReleaseLock(const std::string& resource);
+  void SetError(const std::string& resource, const client::ApiError& error);
+  boost::optional<client::ApiError> GetError(const std::string& resource);
 
  private:
   struct RefCounterMutex {
     std::mutex mutex;
     uint32_t use_count{0u};
+    boost::optional<client::ApiError> optional_error;
```
andescu, line 40: I do not see where this error is reset. Or we do not need to reset it?

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
```diff
@@ -162,10 +162,25 @@ BlobApi::DataResponse DataRepository::GetBlobData(
     }
   }
 
+  // Check if other threads have faced an error.
+  const auto optional_error = mutex.get_error();
```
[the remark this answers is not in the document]
mykolamalik: In this case we would need to get rid of std::unique_lock<NamedMutex> and use NamedMutex::unlock() before every return statement. We have 7 return statements in DataRepository::GetBlobData, so it would complicate the code from NamedMutex user perspective.

olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
mykolamalik: RefCounterMutex instance is removed when the last its user calls ReleaseLock. So we don't need to clear the error explicitly

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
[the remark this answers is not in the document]
andescu: Ok. Let's leave it as you proposed for now, considering that it is a "temporary" solution.

olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.h
```diff
@@ -64,6 +87,21 @@ class NamedMutex final {
...
+  void set_error(const client::ApiError& error);
+
+  /**
+   * @brief Gets the stored error for this mutex.
+   *
+   * @return The stored `ApiError` instance  or `boost::none` if no error has
+   * been stored for this mutex.
+   */
+  boost::optional<client::ApiError> get_error();
+
```
andescu, line 104: Why are these methods lower camel case? The above method lock, try_lock, and unlock are lower snake case for the unique_lock.
Please add a comment above that these methods are to simulate mutex API and make the rest upper camel case.

jmoro3 (dismissed):

olp-cpp-sdk-dataservice-read/src/repositories/DataRepository.cpp
```diff
@@ -162,10 +162,25 @@ BlobApi::DataResponse DataRepository::GetBlobData(
     }
   }
 
+  // Check if other threads have faced an error.
+  const auto optional_error = mutex.GetError();
+  if (optional_error) {
+    OLP_SDK_LOG_DEBUG_F(kLogTag,
+                        "An error is stored in NamedMutex, aborting, hrn='%s', "
```
andescu, line 169: This gives the impression that you are storing the error here.
```suggestion
                        "Found error in NamedMutex, aborting, hrn='%s', "
```

mykolamalik: Done

andescu (approved):

jmoro3 (approved):

PR #1261, opened 2021-10-12
Fix authentication functional tests

mykolamalik:
Update the expectations according to recent API changes.

Also two pessimizing-move warnings are fixed.

Relates-To: OLPEDGE-2598


mykolamalik:
Forse-pushed the same changes to retrigger freezed DCO checker

PR #1293, opened 2021-12-16
Save latest version with GetLatestVersion request

andrey-kashcheev:
When user is using fixed catalog versions it is possible that request
GetLatestVersion won't return actual version from the cache since it
hadn't been written to the cache. Use StartVersion as a new latest
version in the cache-only requests when previous version is absent or
less than the user-set version

Relates-To: OAM-1387

### DIFF
--- a/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/CatalogClient.h
+++ b/olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/CatalogClient.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2020 HERE Europe B.V.
+ * Copyright (C) 2019-2021 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


andescu (commented):
Please write this new behaviour in the CatalogClient::GetLatesVersion API, explaining also that the user should make sure to always provide an existing version in the GetStartVersion() as it will be stored as ::latestVersion in the cache.

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/CatalogClient.h
```diff
@@ -119,6 +119,11 @@ class DATASERVICE_READ_API CatalogClient final {
   /**
    * @brief Gets the catalog version asynchronously.
    *
+   * @note For CacheOnly requests please be sure to specify existing version as
+   * a `StartVersion` in the request so that version could be stored as a new
+   * latest one in case of absence already cached version or when cached version
+   * is less then a new one.
```
andescu, line 125:
```suggestion
   * @note In case you call this API with `FetchOptions::CacheOnly` and a valid version in `CatalogVersionRequest::WithStartVersion()`, i.e. >= 0, then please make sure that the provided version is a existing catalog version as it will be written for later use to the cache as latest version in the following cases:
   * - There is no latest version yet written to cache.
   * - The latest version written to cache is less then the provided version.  
```

olp-cpp-sdk-dataservice-read/include/olp/dataservice/read/CatalogClient.h
```diff
@@ -132,6 +137,11 @@ class DATASERVICE_READ_API CatalogClient final {
...
+   *
```
andescu, line 144: Same as above.

andescu (approved):

mykhailo-kuchma (approved):

PR #1302, opened 2022-01-19
Add continuation functionality

dudnyk:
Add Continuation template classes and implementation
which provide functionality to create a chain of tasks,
and execute tasks.
Add continuation unit tests.

Relates-To: OLPEDGE-2077

Signed-off-by: Yevhenii Dudnyk <ext-yevhenii.dudnyk@here.com>

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/thread/Continuation.h
+++ b/olp-cpp-sdk-core/include/olp/core/thread/Continuation.h
@@ -1,0 +1,277 @@
+/*
+ * Copyright (C) 2022 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+#pragma once
+
+#include <deque>
+#include <functional>
+#include <memory>
+
+#include <olp/core/client/ApiError.h>
+#include <olp/core/client/ApiResponse.h>
+#include <olp/core/client/TaskContext.h>
+#include <olp/core/thread/ExecutionContext.h>
+#include <olp/core/thread/TypeHelpers.h>
+
+namespace olp {
+namespace thread {
+
+class TaskScheduler;
+
+namespace internal {
+
+/// An internal implementation of `Continuation`.
+/// It provides mechanisms to create a chain of tasks, start, cancel,
+/// and finalize an execution.
+///
+/// @note This is a private implementation class for internal use only, and not
+/// bound to any API stability promises. Please do not use directly.
+class CORE_API ContinuationImpl final {
+ public:
+  /// An alias of a function which returns an error as a callback.
+  using FailedCallback = std::function<void(client::ApiError)>;
+  /// Type of return value of Continuation Task.
+  using OutResultType = std::unique_ptr<UntypedSmartPointer>;
+
+  /// Type of Continuation Type.
+  using TaskType = std::function<OutResultType(void*)>;
+  /// Default callback type, uses void* as a parameter which is later converted
+  /// to the specific type, callback doesn't have a return type.
+  using CallbackType = std::function<void(void*)>;
+  /// An internal type of tasks in Continuation.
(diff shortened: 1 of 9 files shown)


olp-cpp-sdk-core/src/thread/Continuation.cpp
```diff
@@ -0,0 +1,213 @@
...
+      // go to the next task in the queue to finilize the task
+      context_copy->tasks_.pop();
+      return false;
+    }
+
+    auto callback = [this, context_copy, &current_task](void* arg) mutable {
+      // limit lifetime of the context shared_ptr to the function scope, for
+      // extra safety in case user's async function stores callback somewhere
+      // for forever
+      auto context = std::move(context_copy);
```
andescu, line 69: Capturing something in the lambda, even if you think you move it actually you only move a copy of it. Are you sure this actually leaves the lambda captured variable empty?

olp-cpp-sdk-core/src/thread/Continuation.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    auto scheduler = task_scheduler_.lock();
+    if (scheduler) {
+      scheduler->ScheduleTask(std::move(current_task_queue_element));
+    }
+  }
+
+ private:
+  struct ProcessorInternal {
+    std::weak_ptr<olp::thread::TaskScheduler> task_scheduler_;
```
andescu, line 137: Why does it need to be weak_ptr. Can it survive the TaskContinuation/Continuation?

olp-cpp-sdk-core/src/thread/Continuation.cpp
```diff
@@ -0,0 +1,213 @@
...
+                        std::pair<ContinuationImpl::AsyncTaskType,
+                                  ContinuationImpl::TaskType>& task) {
+    auto& current_task = task;
+    if (!current_task.first) {
+      // go to the next task in the queue to finilize the task
+      context_copy->tasks_.pop();
+      return false;
+    }
+
+    auto callback = [this, context_copy, &current_task](void* arg) mutable {
```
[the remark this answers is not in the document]
dudnyk: done

olp-cpp-sdk-core/src/thread/Continuation.cpp
dudnyk: removed TaskScheduler from Processor and ProcessorInternal.

olp-cpp-sdk-core/include/olp/core/thread/Continuation.inl
```diff
@@ -0,0 +1,149 @@
...
+  if (impl_.Cancelled()) {
+    finally_callback_(client::ApiError::Cancelled());
+    finally_callback_ = nullptr;
+    return;
+  }
+
+  impl_.Run([this](void* input, bool cancelled) {
+    if (cancelled) {
+      finally_callback_(client::ApiError::Cancelled());
+    } else {
```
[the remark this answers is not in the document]
dudnyk: done in FinishQueueExecution

olp-cpp-sdk-core/include/olp/core/thread/Continuation.h
```diff
@@ -0,0 +1,277 @@
...
+  void Clear();
+
+ private:
+  std::shared_ptr<thread::TaskScheduler> task_scheduler_;
+  std::deque<ContinuationTask> tasks_;
+  ExecutionContext execution_context_;
+
+  /// a variable shows whether it is allow to do the changes to the task chain,
+  /// used to prevent using methods of `Continuation` after it was run.
+  bool continuation_change_allowed{true};
```
andescu, line 134: Or during it runs. Can the user simply call Run() on the continuation and then call Then() without you crashing?

```suggestion
  bool change_allowed_{true};
```

olp-cpp-sdk-core/src/thread/Continuation.cpp
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  void FinishQueueExecution(const std::shared_ptr<ProcessorInternal>& context) {
+    if (context->final_callback_) {
+      // the final execution context, all tasks are done or the execution
+      // is cancelled
+      context->final_callback_(context->LastOutput(), context->IsCancelled());
+      while (!context->tasks_.empty()) {
+        context->tasks_.pop();
+      }
```
andescu, line 90: Is this not the same as calling `context->tasks_.clear()`

olp-cpp-sdk-core/include/olp/core/thread/Continuation.h
dudnyk: renamed the variable.
Added a unit test which checks that Then, Finally and Run don't cause an exceptions
and aren't called if methods were already called once.

olp-cpp-sdk-core/src/thread/Continuation.cpp
dudnyk: replaced queue with deque to be able to clear a container in a simple way.

andescu (approved):

jmoro3 (approved):

PR #1305, opened 2022-01-24
Extend olp::utils::Base64Encode() for URIs.

andescu:
As described in https://datatracker.ietf.org/doc/html/rfc4648#section-5
the padding at the end for URIs can be omitted, therefore the three
available Base64Encode() methods are now extended with an extra
bool parameter which indicates that this is a URI encoding and thus
will omit to pad the output with "=" characters.

Relates-To: OLPEDGE-2680


andrey-kashcheev (commented):
But different alphabet should be used as well to eliminate `'/'` usage. And besides that our `Base64Decode` may not handle output from `Base64Encode( uri, true );` since we check length first and it should be divisible by 4 means always have padding.

andrey-kashcheev (approved):

kzvieriev (approved):

PR #1309, opened 2022-02-03
Add a new cmake option OLP_SDK_USE_LIBCRYPTO

mykhailo-kuchma:
A new option OLP_SDK_USE_LIBCRYPTO is used to conditionally enable or
disable the libcrypto dependency when needed.

Resolves: OLPEDGE-2735


olp-cpp-sdk-core/CMakeLists.txt
```diff
@@ -51,9 +51,10 @@ else()
     find_package(CURL REQUIRED)
 endif()
 
-
 set(Network_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
 
+option(OLP_SDK_USE_LIBCRYPTO "Enables the libcrypto dependency" ON)
```
andescu, line 56: Should it not be in curl.cmake instead. It does not seem relevant to all other cmake includes.

andescu (approved):

HalynaDumych (approved):

PR #1318, opened 2022-04-01
Add API for getting Partition with TileRequest

andrey-kashcheev:
A new API VersionedLayerClient::QuadTreeIndex has been introduced
to get partitions including additional fields.

PartitionsRepository::GetTile and
PartitionsRepository::GetQuadTreeIndexForTile APIs have been extended
with additional_fields argument.

QuadTreeIndex class has been extended with crc.

New tests have been added to cover source code changes.
Touched test files have been slightly refactored.

Relates-To: OAM-1496

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2021 HERE Europe B.V.
+ * Copyright (C) 2019-2022 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.cpp
```diff
@@ -151,6 +151,47 @@ client::CancellationToken VersionedLayerClientImpl::GetData(
...
+    TileRequest tile_request, PartitionsResponseCallback callback) {
+  auto data_task =
+      [=](client::CancellationContext context) mutable -> PartitionsResponse {
+    if (!tile_request.GetTileKey().IsValid()) {
+      return {{client::ErrorCode::InvalidArgument, "Tile key is invalid"}};
+    }
+
+    // FIXME: OnlineOnly since it probably possible needed quad tree could be
+    //  found in the cache without additional fields
+    const auto fetch_option = FetchOptions::OnlineOnly;
```
andescu, line 164: But then we probably do not need it as we already downloaded the bundle also.
The fallback in this case should be different, means if we have something in cache, but it does not contain all the  requested fields we download it again with the requested fields and replace it in cache.

andescu (commented):
Looks already very promising. However I would start first with making sure everything is there what we need in `QuadTreeIndex` class. Means:
- Support for the additional fields
- Get single partition metadata out as JSON
- Get full quadtree out as JSON

mykhailo-kuchma (approved):

ystefinko (approved):

PR #1323, opened 2022-04-14
Change readonly result in case of absent target

smarusik:
- Add error handling to detect if dir checked for access rights is
  absent. Thus in case when access() method returns error, and
  errno variable is set to ENOENT (no such entry), false returned
  to indicate that directory with proper access rights could be
  created

Relates-To: OAM-1168

Signed-off-by: Viacheslav Marusyk <ext-viacheslav.marusyk@here.com>


andescu (commented):
Please enhance your commit message.

andescu (approved):

mykhailo-kuchma (approved):

PR #1331, opened 2022-05-27
Implement buffer appender for log

smarusik:
- Implement cirular (limited size) container on deque base
- Implement appender for circular buffer
- Add tests

Signed-off-by: Viacheslav Marusyk <ext-viacheslav.marusyk@here.com>


smarusik:
Must be moved to com-access-manager-cpp

PR #1346, opened 2022-06-24
Do not use std::regex during static initialization

mykolamalik:
Because it leads to std::bad_cast exception on certain GCC versions,
including GCC 7.5. This is a known GCC bug.

Relates-To: OAM-1620


andescu (commented):
Please extend commit message with: compiler, compiler version, possible workarounds.

ystefinko:
Gitlab-bot - Internal OCM verification passed !

andescu (approved):

dudnyk (approved):

PR #1357, opened 2022-10-17
Fix the data race in NamedMutex class

mykhailo-kuchma:
The race appears when condition variable spins on predicate, while the notify is not protected with mutex.
Additionally set the cancellation method to notify, which enables the faster tasks cancellation.

Relates-To: OAM-1784


olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
```diff
@@ -145,7 +145,10 @@ NamedMutex::NamedMutex(NamedMutexStorage& storage, const std::string& name,
 NamedMutex::~NamedMutex() { storage_.ReleaseLock(name_); }
 
 void NamedMutex::lock() {
-  if (!context_.IsCancelled() && !try_lock()) {
+  const bool valid = context_.ExecuteOrCancelled(
+      [&]() { return client::CancellationToken([&]() { Notify(); }); });
```
Hsilgos, line 149: It looks dangerous that 'this' is captured in lambda of object witch lifetime is bigger than NamedMutex. I would either overwrite lambda when 'unlock' is near finish or i would make Notify static and of necessary data outside

Hsilgos (approved):

olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
```diff
@@ -145,10 +145,15 @@ NamedMutex::NamedMutex(NamedMutexStorage& storage, const std::string& name,
...
+
+  if (valid) {
     std::unique_lock<std::mutex> unique_lock{lock_mutex_};
     lock_condition_.wait(unique_lock,
                          [&] { return context_.IsCancelled() || try_lock(); });
+    // Reset the cancel token
+    context_.ExecuteOrCancelled([&]() { return client::CancellationToken(); });
```
andescu, line 156: Why do you need to capture anything?
```suggestion
    context_.ExecuteOrCancelled([]() { return client::CancellationToken(); });
```

mykolamalik (approved):

olp-cpp-sdk-dataservice-read/src/repositories/NamedMutex.cpp
andescu: As no side effect is expected, please consider this for the next PR

PR #1363, opened 2022-11-03
Check OLP_SDK_ENABLE_ANDROID_CURL for Android only

andrey-kashcheev:
Make sure that OLP_SDK_ENABLE_ANDROID_CURL doesn't affect other platforms, as it should be possible to turn on this option without additional checks for platform.

Relates-To: OLPSUP-20966


andrey-kashcheev:
@rustam-gamidov-here Previously, the user was required to check that the platform is actually Android before enabling OLP_SDK_ENABLE_ANDROID_CURL. This is not very obvious and incorrect in my opinion and we have the first issue from that.
And now, the user shouldn't check anything and just enable this flag if needed.

andescu (approved):

PR #1375, opened 2022-12-12
Open cache in r/o mode on `no space left` errors

andrey-kashcheev:
Since protected cache could be used as a static cache it is expected to be able to use it even when there is no available space on device. But when protected cache is being open in R/W mode levelDb may try to create some files and this failure would be treated later as a corrupted cache.
Instead, on `no space left` errors we will try to open the cache again but in read-only mode to enforce static behavior.

Relates-To: OLPSUP-21193

### DIFF
--- a/olp-cpp-sdk-core/src/cache/DiskCache.cpp
+++ b/olp-cpp-sdk-core/src/cache/DiskCache.cpp
@@ -247,6 +247,13 @@ OpenResult DiskCache::Open(const std::string& data_path,
 
   if (status.IsCorruption() || status.IsIOError()) {
     if (is_read_only || !repair_if_broken) {
+      if (status.IsIOError()) {
+        OLP_SDK_LOG_ERROR_F(
+            kLogTag, "Open: IO error, cache_path='%s', error='%s'",
+            versioned_data_path.c_str(), status.ToString().c_str());
+        return OpenResult::IOError;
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/src/cache/DiskCache.cpp
```diff
@@ -45,6 +45,13 @@ constexpr auto kLogTag = "DiskCache";
 constexpr auto kLevelDbLostFolder = "/lost";
 constexpr auto kMaxL0Files = 4;
 
+#if defined(_WIN32) && !defined(__MINGW32__)
+// FIXME: How this message looks on Windows?
```
rustam-gamidov-here, line 49: Why do we need this and even with ifdef

andrey-kashcheev: I don't know how to check the error message on Windows since I couldn't find any examples of error messages but I assume it should be different from other platforms.
The issue here is that leveldb shows a `no space left` error as an `IOError`, and to get what actual IO error you need to parse the string representation of leveldb::Status.

rustam-gamidov-here (approved):

mykhailo-kuchma (approved):

rustam-gamidov-here (approved):

PR #1377, opened 2022-12-30
Consider error in openning default disk cache as not fatal

rustam-gamidov-here:
This error causes it to fall back to a memory cache so there's no need to break the application execution as it's recommended for the fatal errors.

Relates-To: OLPSUP-21746
Signed-off-by: Rustam Gamidov <ext-rustam.gamidov@here.com>


mykolamalik (dismissed):

mykolamalik (approved):

olp-cpp-sdk-core/src/client/OlpClientSettingsFactory.cpp
```diff
@@ -52,7 +52,7 @@ OlpClientSettingsFactory::CreateDefaultCache(cache::CacheSettings settings) {
   auto cache = std::make_unique<cache::DefaultCache>(settings);
   auto error = cache->Open();
   if (error == cache::DefaultCache::StorageOpenResult::OpenDiskPathFailure) {
-    OLP_SDK_LOG_FATAL_F(
+    OLP_SDK_LOG_ERROR_F(
```
andrey-kashcheev, line 55: Update log level on line 66 as well

andrey-kashcheev (approved):

mykolamalik (approved):

PR #1395, opened 2023-04-20
Makes possible to use milliseconds for network timeouts.

Hsilgos:
Timeout less than second makes sense when network is used on backend and it must quicly react if it can't connect to another service.

The changes:

- chrono is used to keep timeout in settings
- Related code is adapted to pass chrono primitives.
- CURL implementation specifies timeout in milliseconds

Relates-To: OLPSUP-23435


olp-cpp-sdk-core/include/olp/core/http/NetworkSettings.h
mykhailo-kuchma, line 1: We can't change the existing API

mykhailo-kuchma (changes requested):

olp-cpp-sdk-core/include/olp/core/http/NetworkSettings.h
Hsilgos: Should I add additional methods? Do you have deprecation process?

mykhailo-kuchma (dismissed):

olp-cpp-sdk-core/src/http/NetworkSettings.cpp
```diff
@@ -25,10 +25,20 @@ namespace http {
...
   return connection_timeout_;
 }
 
-int NetworkSettings::GetTransferTimeout() const { return transfer_timeout_; }
+int NetworkSettings::GetTransferTimeout() const {
+  return static_cast<int>(
+    std::chrono::duration_cast<std::chrono::seconds>(GetTransferTimeoutDuration()).count());
+}
+
+std::chrono::milliseconds NetworkSettings::GetTransferTimeoutDuration() const { return transfer_timeout_; }
```
andrey-kashcheev, line 41: Have you run clang-format over this PR?

andrey-kashcheev (approved):

PR #1409, opened 2023-06-26
Change log level of detailed tracing messages to TRACE instead of DEBUG

VeaaC:
One of the main things the SDK does is issue network requests. These are issued all the time, and outputting log messages for them is more akin to tracing the debug logging. The same is true for cancelling network requests: Rendering libraries will typically cancel requests if you pan the map.

Without this change debug logs are extremely spammy and hard to analyze.

What do you think?


VeaaC:
Any feedback on this?

uhertlein:
This looks like a reasonable change to me.

I assume it's intentional that only "send" and "cancel" are dropped to TRACE, but not "message completed"?

VeaaC:
@uhertlein I guess that's because those message were the only ones spamming in my apps :) 

Which messages would you suggest to also downgrade?

uhertlein:
I didn't want to suggest to change more message to TRACE.

Keeping "Message completed" at DEBUG is good IMO, because it contains enough information to make "Send request" semi-redundant.

mykhailo-kuchma:
Have you considered changing the log level on your side (you can change it for CURL only https://github.com/heremaps/here-data-sdk-cpp/blob/59730bc47b832f111e786bdb1a6facd9a17d4760/olp-cpp-sdk-core/include/olp/core/logging/Log.h#L464)? You'll need a send request log when you need to debug when the request starts.

VeaaC:
@mykhailo-kuchma sure, but then you hide other logging information as a side-effect. So the question is what you would expect conceptually on each level.

TRACE: detailed information about what is going one, expensive to log (lots of log output)
DEBUG: Extra information for weird things going on
INFO: Information you need if everything works normally
WARNING: Something is potentially broken
ERROR: Something is definitely broken

It makes sense to think this through and then put each piece of information were it belongs.

For me this seemed to fit TRACE very well, since you have a huge amount of log output (e.g. if using a simple renderer that downloads 100+ partitions when scrolling across the map)

PR #1441, opened 2023-10-19
Adapt the code to GCC-13

mykhailo-kuchma:
cstdint header is required for std::int64_t type

Relates-To: OLPEDGE-2843

### DIFF
--- a/olp-cpp-sdk-core/include/olp/core/utils/Dir.h
+++ b/olp-cpp-sdk-core/include/olp/core/utils/Dir.h
@@ -19,6 +19,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <functional>
 #include <string>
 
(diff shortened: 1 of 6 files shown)


olp-cpp-sdk-core/include/olp/core/utils/Dir.h
```diff
@@ -20,6 +20,7 @@
 #pragma once
 
 #include <functional>
+#include <cstdint>
```
mykolamalik, line 23:
```suggestion
#include <cstdint>
#include <functional>
```

mykolamalik (changes requested):

rustam-gamidov-here (approved):

mykolamalik (approved):

PR #1444, opened 2023-10-31
Added connection and transfer timeout settings.

Smartkras:
Added connection and transfer timeout options to RetrySettings. Improved methods description. Fixed curl implementation.

Resolves: OLPSUP-23107


olp-cpp-sdk-core/include/olp/core/client/RetrySettings.h
```diff
@@ -67,9 +67,32 @@ struct CORE_API RetrySettings {
...
+   * period of time ignoring `connection_timeout` and `transfer_timeout` values.
+   */
+  int timeout = 90;
+
+  /**
+   * @brief Time allowed to wait for connection to establish (in milliseconds).
+   *
+   * The default value is 60 seconds.
+   *
+   * @note The value should be smaller than `timeout`.
```
andrey-kashcheev, line 84: And what if it is larger? It is more of a suggestion I guess

olp-cpp-sdk-core/include/olp/core/client/RetrySettings.h
```diff
@@ -67,9 +67,32 @@ struct CORE_API RetrySettings {
...
   /**
-   * @brief The connection timeout limit (in seconds).
+   * @brief Maximum time for request to complete (in seconds).
+   *
+   * The default value is 90.
+   *
+   * @note Connection or data transfer will be interrupted after specified
+   * period of time ignoring `connection_timeout` and `transfer_timeout` values.
+   */
+  int timeout = 90;
```
andrey-kashcheev, line 77: Can we easily change the default values of settings?

olp-cpp-sdk-core/include/olp/core/client/RetrySettings.h
```diff
@@ -67,9 +67,32 @@ struct CORE_API RetrySettings {
...
+
+  /**
+   * @brief Time allowed to wait without data transfer (in milliseconds).
+   *
+   * The default value is 30 seconds.
+   *
+   * @note The value should be smaller than `timeout`.
    */
-  int timeout = 60;
+  std::chrono::milliseconds transfer_timeout = std::chrono::seconds(30);
```
andrey-kashcheev, line 95: Why connection is 60 but the transfer only 30? I would expect it vice versa, as usually most time spent downloading something and not just connecting

olp-cpp-sdk-core/include/olp/core/client/RetrySettings.h
Smartkras: The consequences are described in timeout field description.

olp-cpp-sdk-core/include/olp/core/client/RetrySettings.h
Smartkras: We can left is as is (60 seconds), but then we will need to change connection value to smaller, other vice default values will violate our recomendation.

olp-cpp-sdk-core/include/olp/core/client/RetrySettings.h
Smartkras: It's a default values in NetworkSettings. Values form RetrySettings will be applied to NetworkSettings eventually.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -673,8 +673,8 @@ ErrorCode NetworkCurl::SendImplementation(
   curl_easy_setopt(handle->handle, CURLOPT_FOLLOWLOCATION, 1L);
   curl_easy_setopt(handle->handle, CURLOPT_CONNECTTIMEOUT_MS,
                    config.GetConnectionTimeoutDuration().count());
-  curl_easy_setopt(handle->handle, CURLOPT_TIMEOUT_MS,
-                   config.GetTransferTimeoutDuration().count());
+  curl_easy_setopt(handle->handle, CURLOPT_SERVER_RESPONSE_TIMEOUT,
```
rustam-gamidov-here, line 676: Are you sure about this one?
It description looks more like connection timeout, and `CURLOPT_TIMEOUT_MS` is a transfer timeout.

olp-cpp-sdk-core/include/olp/core/client/RetrySettings.h
Smartkras: Usually we spend more time waiting for connection because of DNS name resolving, opening sockets, sending acknowlidge etc. When connection is established - data transfers over already existing 'channel', so it makes more sense to set this value lower.

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
Smartkras: CURLOPT_TIMEOUT_MS sets overall time for request to finish, means it will not work as a transfer timeout. This one is actually what we need: "Causes libcurl to set a timeout period (in seconds) on the amount of time that the server is allowed to take in order to send a response message for a command before the session is considered dead." Since there is no other timeouts related to data transfer in curl.

olp-cpp-sdk-core/include/olp/core/client/RetrySettings.h
Smartkras: What would be the suggested values for connection and transfer timeouts ?

rustam-gamidov-here (approved):

mykhailo-kuchma (approved):

PR #1450, opened 2023-11-23
NetworkCurl header parsing performance improvement

uhertlein:
This change contains one minor code cleanup (using a local variable 'curl_handle' instead of repeated 'handle->handle'), a fix for a Curl API deprecation warning (CURLINFO_SIZE_DOWNLOAD is deprecated >= 7.55.0), as well as one (potential) performance improvement.

The 'HeaderFunction' is called for every line of the header.  In case no 'header_callback' is given by the user then we can skip all of the processing.  Additionally, we can avoid one local string copy for the trimmed (trailing '\r' and '\n' removed) input buffer.

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
@@ -219,12 +219,21 @@ void GetTrafficData(CURL* handle, uint64_t& upload_bytes,
     download_bytes += headers_size;
   }
 
+#if CURL_AT_LEAST_VERSION(7, 55, 0)
+  off_t length_downloaded = 0;
+  if (curl_easy_getinfo(handle, CURLINFO_SIZE_DOWNLOAD_T, &length_downloaded) ==
+          CURLE_OK &&
+      length_downloaded > 0) {
+    download_bytes += length_downloaded;
+  }
+#else
   double length_downloaded;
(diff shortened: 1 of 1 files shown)


olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -919,27 +927,35 @@ size_t NetworkCurl::HeaderFunction(char* ptr, size_t size, size_t nitems,
...
   size_t count = len;
   while ((count > 1u) &&
          ((ptr[count - 1u] == '\n') || (ptr[count - 1u] == '\r')))
     count--;
   if (count == 0u) {
     return len;
   }
-  std::string str(ptr, count);
+
+  std::string_view str(ptr, count);
```
andrey-kashcheev, line 944: This project allows only C++11. So, std::string_view could not be used.

uhertlein: Got it.

Apart from that item, are you okay with the changes?

andrey-kashcheev (commented):
You need to provide a link to the ticket (e.g. Relates-To: ... ) to pass a commit checker

mykhailo-kuchma (approved):

andrey-kashcheev (approved):

PR #1453, opened 2023-11-28
Fix minimum curl version for CURLOPT_CAINFO_BLOB

rbuch703:
This MR fixes compilation on Rocky Linux 9 (and thus likely also on
RHEL9), which uses curl-7.76 and thus - as noted on
https://curl.se/libcurl/c/CURLOPT_CAINFO_BLOB.html - does not yet have
support for CURLOPT_CAINFO_BLOB.

Relates-To: OLPSUP-27800

Signed-off-by: Robert Buchholz <robert.buchholz@here.com>

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.h
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.h
@@ -49,7 +49,9 @@
 #endif
 #endif
 
-#if CURL_AT_LEAST_VERSION(7, 71, 0)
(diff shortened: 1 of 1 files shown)


andrey-kashcheev (commented):
You need to provide a link to the ticket (e.g. Relates-To: ... ) to pass a commit checker. The correct one I guess will be OLPSUP-27800.

andrey-kashcheev (approved):

mykolamalik (approved):

PR #1464, opened 2024-01-11
Add multi partitions Protect and Release

rustam-gamidov-here:
The main goal is to decrease number of calls and number of log messages

Relates-To: OLPEDGE-2855

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
+++ b/olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2023 HERE Europe B.V.
+ * Copyright (C) 2019-2024 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


olp-cpp-sdk-dataservice-read/src/repositories/PartitionsCacheRepository.cpp
```diff
@@ -357,8 +357,17 @@ PartitionsCacheRepository::CreatePartitionKeys(
...
-  const auto keys = CreatePartitionKeys(partition_id, version);
+    const std::vector<std::string>& partition_ids,
+    const boost::optional<int64_t>& version) {
+  cache::KeyValueCache::KeyListType keys;
+  std::for_each(partition_ids.cbegin(), partition_ids.cend(),
+                [&](const std::string& partition_id) {
+                  auto partition_keys =
+                      CreatePartitionKeys(partition_id, version);
+                  std::move(partition_keys.begin(), partition_keys.end(),
+                            std::back_inserter(keys));
```
andrey-kashcheev, line 368: I'm not sure that it is worth it but could we preallocate `keys` (assume successful `GetPartitionHandle` calls)? Should be `partition_ids.size() * 2u`.
Or at least verify that this approach is on par with the previous one in terms of speed.

andrey-kashcheev (approved):

mykhailo-kuchma (approved):

PR #1469, opened 2024-01-30
Followup on xcode upgrade

rustam-gamidov-here:


rustam-gamidov-here:
closing in favor of #1468 (the original one)

PR #1470, opened 2024-02-01
Add NetworkOffline for offline environment

rustam-gamidov-here:
There are cases where device want to rely on the offline data only or even want to completely avoid network related libs like libcurl and libssl.

Relates-To: OLPEDGE-2861

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2019-2021 HERE Europe B.V.
+# Copyright (C) 2019-2024 HERE Europe B.V.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -46,6 +46,7 @@ option(OLP_SDK_ENABLE_DEFAULT_CACHE "Enable default cache implementation based o
 option(OLP_SDK_ENABLE_DEFAULT_CACHE_LMDB "Enable default cache implementation based on LMDB" OFF)
 option(OLP_SDK_ENABLE_ANDROID_CURL "Enable curl based network layer for Android" OFF)
 option(OLP_SDK_ENABLE_IOS_BACKGROUND_DOWNLOAD "Enable iOS network layer downloading in background. Under testing." OFF)
+option(OLP_SDK_ENABLE_OFFLINE_MODE "Enable offline mode. Network layer is excluded from the build and all network requests returned with error." OFF)
 
 # C++ standard version. Minimum supported version is 11.
 set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ option(OLP_SDK_ENABLE_DEFAULT_CACHE "Enable default cache implementation based o
 option(OLP_SDK_ENABLE_DEFAULT_CACHE_LMDB "Enable default cache implementation based on LMDB" OFF)
 option(OLP_SDK_ENABLE_ANDROID_CURL "Enable curl based network layer for Android" OFF)
 option(OLP_SDK_ENABLE_IOS_BACKGROUND_DOWNLOAD "Enable iOS network layer downloading in background. Under testing." OFF)
+option(OLP_SDK_ENABLE_OFFLINE_MODE "Enable offline mode. All network requests returned with error." OFF)
```
andrey-kashcheev, line 49: Include here that network-related dependencies are not needed as well.

olp-cpp-sdk-core/src/http/offline/NetworkOffline.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+NetworkOffline::~NetworkOffline() {
+  OLP_SDK_LOG_TRACE(kLogTag, "Destroyed NetworkOffline object, this=" << this);
+}
+
+SendOutcome NetworkOffline::Send(NetworkRequest request,
+                                 std::shared_ptr<std::ostream> payload,
+                                 Network::Callback callback,
+                                 Network::HeaderCallback header_callback,
+                                 Network::DataCallback data_callback) {
```
andrey-kashcheev, line 48: Mark arguments with OLP_SDK_CORE_UNUSED

scripts/ios/azure_ios_build_psv.sh
```diff
@@ -18,8 +18,9 @@
 # License-Filename: LICENSE
 
 # Getting rid of boost installed to the CI image (not clear when it appeared)
-# asciidoc and source-highlight are using it so dependencies are ignored
-brew uninstall --ignore-dependencies boost
+# asciidoc and source-highlight are using it so dependencies are ignored.
+# Sometimes there is no boost and there is no need to fail in this case.
+brew uninstall --ignore-dependencies boost || true
```
andrey-kashcheev, line 23: Is this one related to the change?

rustam-gamidov-here: Partially. Recently there was an issue because of boost installed using brew in the image so it was uninstalled #1467. It appeared that not all containers has that boost so our builds started failing from time to time because of that. It was a separate commit in this PR, extracted to another PR #1471

CMakeLists.txt
rustam-gamidov-here: Extended

mykolamalik (approved):

PR #1482, opened 2024-02-28
Debug: attempt to build zlib on CI

rustam-gamidov-here:


rustam-gamidov-here:
passed with updated zlib in the android build log so #1479 is merged

PR #1491, opened 2024-04-02
Fixed Md5LookupGetBySubject change for openssl

sameermahant:
Fixed Md5LookupGetBySubject change for OpenSSL v3.1.4

### DIFF
--- a/scripts/misc/commit_checker.sh
+++ b/scripts/misc/commit_checker.sh
@@ -35,6 +35,7 @@ num_lines=`wc -l commit.log | cut -d'c' -f1`
 for (( line=1; line<=${num_lines}; line++ ))
 do
     current_line_len=`cat commit.log | sed -n ${line}p | tr -d '\n\r' | wc -m | sed 's/\ \ \ \ \ \ /''/g' `
+    current_line=`cat commit.log | sed -n ${line}p`
     if [ $line -eq 1 ] && [ ${current_line_len} -gt 50 ] ; then
         echo ""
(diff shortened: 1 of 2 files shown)


scripts/misc/commit_checker.sh
```diff
@@ -65,16 +66,31 @@ do
...
-        echo "ERROR: ${current_line_len} chars in ${line}-th line is too long. Any line length must be less than or equal to 72 chars !"
-        echo ""
-        cat commit.log
-        echo "----------------------------------------------"
-        echo "Please read following rules:"
-        cat scripts/misc/commit_message_recom.txt
-        exit 3
+    if [[ ${current_line} == *"Signed-off-by"* ]]; then
+        echo " ${line}-th line is sign-off line."
+        if [ ${current_line_len} -gt 80 ]; then
```
rustam-gamidov-here, line 71: the whole 'if' block looks like a function that can be called twice with 80 and with 72

scripts/misc/commit_checker.sh
```diff
@@ -65,16 +66,31 @@ do
...
+            cat commit.log
+            echo "----------------------------------------------"
+            echo "Please read following rules:"
+            cat scripts/misc/commit_message_recom.txt
+            exit 3
+        fi
+        echo " ${line}-th line is ${current_line_len} chars length . OK."
+        continue
     fi
+    if [ ${current_line_len} -gt 72 ] ; then
```
rustam-gamidov-here, line 84: It's quite easy to miss this 'continue' and two 'if's adding more checks after this one.
Removing continue and using else looks more robust to me

PR #1499, opened 2024-04-19
IOTSDK-24021: Fix parsing of server DateTime

diachenko-mischa:
std::get_time does not claim support of time zones in the string parameter, and seems to be subject of locale set on the system. An example is a following result on linux:

ParseTime(Thu, 1 Jan 1970 0:00:00 GMT)=-2209075200

::strptime is also a subject to the time zones of locale, but it also claim support for parsing a time zone from the passed string. Thus, additional logic incorporated to offset parsed DataTime to the local timezone.


mykolamalik (changes requested):
Please update the names of functions and variables in order to follow the project's code style: UpperCamelCase for function names, snake_case for variables

diachenko-mischa:
> Please update the names of functions and variables in order to follow the project's code style: UpperCamelCase for function names, snake_case for variables

Fixed

mykolamalik (approved):

mykhailo-kuchma (approved):

PR #1513, opened 2024-05-17
Extension of KeyValueCache interface

andrey-kashcheev:
Added main operations like Read, Write, Remove.
The difference with the old one is that new API may return actual error instead of cryptic `false`.
DefaultCache implementation was extended as well
to cover new methods.

Relates-To: OCMAM-3, OAM-2431

### DIFF
--- a/olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
+++ b/olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
@@ -122,7 +122,7 @@ class DefaultCacheImplHelper : public cache::DefaultCacheImpl {
       return false;
     }
 
-    return disk_cache->Get(key) != boost::none;
+    return disk_cache->Get(key).IsSuccessful();
   }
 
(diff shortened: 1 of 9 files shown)


olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
```diff
@@ -122,7 +122,7 @@ class DefaultCacheImplHelper : public cache::DefaultCacheImpl {
       return false;
     }
 
-    return disk_cache->Get(key) != boost::none;
+    return disk_cache->Get(key).IsSuccessful();
```
mykhailo-kuchma, line 125: Why don't we use iterator + seek to find the key?

mykhailo-kuchma: Ok, I see it's a test file

rustam-gamidov-here (approved):
`This error is not correct` looks a bit scary for not ideal error code as for me. Not critical though

mykhailo-kuchma (approved):

PR #1526, opened 2024-06-20
Add Xcode13 job

ystefinko:
After PR#1523 we are using only Xcode14 and 15 for builds
as they are default on macos-12, macos-13, macos-14.
However, macos-12 still contains Xcode13.1 which could be used.

Relates-To: MINOR-UPDATES-CI

### DIFF
--- a/scripts/ios/azure_ios_build_psv.sh
+++ b/scripts/ios/azure_ios_build_psv.sh
@@ -26,7 +26,7 @@ if [[ ${USE_LATEST_XCODE} == 0 ]]; then
   # Due to some bug which is cmake cannot detect compiler while called
   # from cmake itself when project is compiled with XCode 12.4 we must
   # switch to old XCode as a workaround.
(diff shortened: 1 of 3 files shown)


scripts/ios/azure_ios_build_psv.sh
```diff
@@ -26,7 +26,7 @@ if [[ ${USE_LATEST_XCODE} == 0 ]]; then
   # Due to some bug which is cmake cannot detect compiler while called
   # from cmake itself when project is compiled with XCode 12.4 we must
   # switch to old XCode as a workaround.
-  sudo xcode-select -s /Applications/Xcode_11.7.app
```
rustam-gamidov-here, line 29: Not big deal but looks like this mechanism got a bit misleading naming with several macos and xcode versions.
Just a comment, not something to fix in this PR

rustam-gamidov-here (approved):

mykhailo-kuchma (approved):

PR #1527, opened 2024-06-26
Use USERPROFILE env variable to get home path on Windows

mykolamalik:
Some Windows distributions have no HOMEDRIVE or/and HOMEPATH vars set by default. They have USERPROFILE variable instead.

Relates-To: OCMAM-115

### DIFF
--- a/olp-cpp-sdk-authentication/src/AuthenticationCredentials.cpp
+++ b/olp-cpp-sdk-authentication/src/AuthenticationCredentials.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2023 HERE Europe B.V.
+ * Copyright (C) 2019-2024 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


rustam-gamidov-here (approved):

olp-cpp-sdk-authentication/src/AuthenticationCredentials.cpp
```diff
@@ -34,6 +34,10 @@ const std::string kHereTokenEndpointUrl = "here.token.endpoint.url";
 std::string GetDefaultPath() {
   const char* env;
 #ifdef _WIN32
+  if ((env = std::getenv("USERPROFILE")) != nullptr) {
+    return std::string(env).append("\\.here\\credentials.properties");
```
andrey-kashcheev, line 38: It would be nice to have a properties path as a separate string since now it has multiple usages but okay

andrey-kashcheev (approved):

olp-cpp-sdk-authentication/src/AuthenticationCredentials.cpp
mykolamalik: I thought about it but decided that for 2 usages it would not make much sense. Linux path string differs. But if we would have a const var for the Windows case, then we should have a var for the Linux case as well. So it would complicate the code more than the benefits it brings.
For 3 usages I would definitely make a var.

PR #1548, opened 2024-09-25
Always require OpenSSL

andrey-kashcheev:
OpenSSL now required, and should be always present
hence `OLP_SDK_NETWORK_HAS_OPENSSL` could be removed

Relates-To: HERESUP-1440

### DIFF
--- a/olp-cpp-sdk-core/cmake/curl.cmake
+++ b/olp-cpp-sdk-core/cmake/curl.cmake
@@ -33,20 +33,13 @@ if(CURL_FOUND AND NOT NETWORK_NO_CURL)
     endif()
 
     if(NOT OPENSSL_FOUND)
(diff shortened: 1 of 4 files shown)


mykhailo-kuchma (approved):

olp-cpp-sdk-core/cmake/curl.cmake
```diff
@@ -33,20 +33,13 @@ if(CURL_FOUND AND NOT NETWORK_NO_CURL)
     endif()
 
     if(NOT OPENSSL_FOUND)
```
rustam-gamidov-here, line 35: Just curious do we need this check to prevent extra `find_package`?

rustam-gamidov-here (approved):

PR #1563, opened 2024-11-11
Update Apple CI jobs, skip boost compile warning

ystefinko:
According to updates on images: macos-12 deprecation we should use macos-13 and Xcode 14.3 as the lowest versions in CI validation.
Also, add х86_64 macos-13 image with job separately, while other images 
are by default on ARM.
Additional compiler options are related boost internal warnings 
on newer xcodes.
Add more MacOS jobs to cover more combinations: MacOS-Xcode.

Relates-TO: DATASDK-47

### DIFF
--- a/.github/workflows/psv_pipelines.yml
+++ b/.github/workflows/psv_pipelines.yml
@@ -176,45 +176,75 @@ jobs:
       run: scripts/android/build.sh
       shell: bash
 
-  psv-macos-12-xcode-14-build:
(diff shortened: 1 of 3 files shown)


rustam-gamidov-here (approved):

rustam-gamidov-here:
It probably worth a comment that additional compiler options are related boost internal warnings on newer xcodes.
Could be in the script or in the PR description to be visible in git history.

andrey-kashcheev (approved):

PR #1580, opened 2025-01-16
Minor copyright and CI update

ystefinko:
- update and fix FV pipes
- updatecopyright in readme
- update CI due to recent macOS runner changes

Relates-To: MINOR

### DIFF
--- a/.github/workflows/psv_pipelines.yml
+++ b/.github/workflows/psv_pipelines.yml
@@ -186,17 +186,7 @@ jobs:
       run: scripts/macos/psv/azure_macos_build_psv.sh
       shell: bash
 
-  psv-macos-13-xcode-15-build:
-    name: PSV.MacOS13.Xcode15
-    runs-on: macos-13-xlarge
-    steps:
-    - name: Check out repository
-      uses: actions/checkout@v4
-    - name: MacOS Build Xcode15
-      run: scripts/macos/psv/azure_macos_build_psv.sh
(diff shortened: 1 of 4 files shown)


external/CMakeLists.txt
```diff
@@ -61,10 +61,10 @@ set(OLP_SDK_CPP_ZLIB_URL "https://www.zlib.net/zlib-1.3.1.tar.gz")
 set(OLP_SDK_CPP_ZLIB_HASH "SHA256=9a93b2b7dfdac77ceba5a558a580e74667dd6fede4585b91eefb60f03b72df23")
 
 set(OLP_SDK_CPP_OPENSSL_URL "https://github.com/openssl/openssl.git")
-set(OLP_SDK_CPP_OPENSSL_TAG "OpenSSL_1_1_1w")
+set(OLP_SDK_CPP_OPENSSL_TAG "openssl-3.1.4")
```
rustam-gamidov-here, line 64: Openssl v3 has much more generated files than there is in cmake wrapper for openssl v1 so it will fail the build once forced. So we should either update that cmake wrapper for v3 and say goodbye for v1 or use openssl build system

ystefinko: thanks for checking , tbd in next MRs

rustam-gamidov-here (approved):

andrey-kashcheev (approved):

PR #1584, opened 2025-02-03
Add network verbose log output to file

mykhailo-kuchma:
Add network verbose log output to a file

Fix the CURLOPT_NOSIGNAL option is not set after reset

Relates-To: OCMAM-418

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
@@ -203,22 +203,18 @@ int ConvertErrorCode(CURLcode curl_code) {
   }
 }
 
-/**
- * @brief CURL get upload/download data.
- * @param[in] handle CURL easy handle.
- * @param[out] upload_bytes uploaded bytes(headers+data).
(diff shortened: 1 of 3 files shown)


olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -307,6 +308,17 @@ NetworkCurl::NetworkCurl(NetworkInitializationSettings settings)
                         static_cast<int>(error));
   }
 
+  if (curl_log_path_) {
+    stderr_ = fopen(curl_log_path_->c_str(), "w+");
+    if ( stderr_ == nullptr ) {
+      OLP_SDK_LOG_ERROR(
```
rustam-gamidov-here, line 314: Maybe warning as it is not crucial for the functionality

rustam-gamidov-here (dismissed):

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
mykhailo-kuchma: Will fix in the next commit

PR #1589, opened 2025-02-10
Disable CURL diagnostics for old versions

mykhailo-kuchma:
Disable timings for CURL versions before 7.61

Relates-To: OCMAM-418

### DIFF
--- a/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
+++ b/olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
@@ -348,9 +348,9 @@ void SetupDns(CURL* curl_handle, const std::vector<std::string>& dns_servers) {
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


rustam-gamidov-here (approved):

olp-cpp-sdk-core/src/http/curl/NetworkCurl.cpp
```diff
@@ -348,9 +348,9 @@ void SetupDns(CURL* curl_handle, const std::vector<std::string>& dns_servers) {
 #endif
 }
 
-Diagnostics GetDiagnostics(CURL* handle) {
+void WithDiagnostics(NetworkResponse& response, CURL* handle) {
```
andrey-kashcheev, line 351: Mark as unused otherwise

andrey-kashcheev (approved):

PR #1620, opened 2025-07-09
Custom body to be sent to URL from Credentials

diachenko-mischa:
Here credentials file specifies the endpoint that
should be used to retrieve authentication token.
HNAV backend, which implements mTLS authentication also conforms to a bit different body and response schema.

Add a possibility to add a custom body for the authentication request and parsing of token from HNAV mTLS server.

Relates-To: HERESDK-7942


rustam-gamidov-here (commented):
No comments to the code but it make sense to update years in the files changed

PR #1624, opened 2025-08-05
Add clang-format check

rustam-gamidov-here:
There is no clang-format 6.0 binaries for current ubuntu versions.
Github pool has no old ubuntu images. So preparation steps are
downloading binaries from older ubuntu distros and installing them.

python-is-python2 symlinks python to python2 and covers clang-format 6.0
package requirement for python. python-is-python3 is conflicting with
the python-is-python2 thus removed

### DIFF
--- a/.github/workflows/psv_pipelines.yml
+++ b/.github/workflows/psv_pipelines.yml
@@ -273,3 +273,49 @@ jobs:
     - name: Commit checker script. Verify commit text
       run: scripts/misc/commit_checker.sh
       shell: bash
+
+  psv-formatting-checker:
+    name: PSV.Clang.Format.Checker
+    runs-on: ubuntu-22.04
+    env:
+      CLANG_FORMAT_FILE: "clang-format.diff"
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v4
+    - name: Setup environment
+      run: |
+        set +x
+        set -e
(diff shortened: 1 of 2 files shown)


.github/workflows/psv_pipelines.yml
```diff
@@ -273,3 +273,44 @@ jobs:
...
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v4
+    - name: Setup environment
+      run: |
+        set +x
+        set -e
+        sudo apt-get update
+        sudo apt-get install -y wget libedit2 libtinfo5 python2
+        alias python=python2
```
andrey-kashcheev, line 291: Maybe package `python-is-python2` is better?

.github/workflows/psv_pipelines.yml
```diff
@@ -273,3 +273,44 @@ jobs:
...
+    - name: Store formatting check results
+      uses: actions/upload-artifact@v4
+      with:
+        name: clang-format-diff
+        path: ${{ env.CLANG_FORMAT_FILE }}
+    - name: Verify check result
+      run: |
+        set +x
+        if [ -s ${CLANG_FORMAT_FILE} ] ; then
+          exit 1
```
andrey-kashcheev, line 314: Put the message here with explanation that the link to the patch is in the previous step

PR #1625, opened 2025-08-07
Use std::optional if available.

asopov-here:
So, boost::optional type is used for optional values. For C++17 and above more convenient is to use std::optional instead. The task NLAM-23 is about making this type configurable. This commit is a second part of the task: olp-cpp-sdk-authentication.

Relates-To: NLAM-23

### DIFF
--- a/olp-cpp-sdk-authentication/include/olp/authentication/AuthorizeRequest.h
+++ b/olp-cpp-sdk-authentication/include/olp/authentication/AuthorizeRequest.h
@@ -19,11 +19,11 @@
 
 #pragma once
 
-#include <boost/optional.hpp>
 #include <string>
 #include <vector>
 
 #include <olp/authentication/AuthenticationApi.h>
+#include <olp/core/porting/optional.hpp>
 
 namespace olp {
 namespace authentication {
@@ -83,7 +83,7 @@ class AUTHENTICATION_API AuthorizeRequest final {
    *
    * @return The service ID.
    */
-  inline const std::string& GetServiceId() const { return service_id_; }
+  const std::string& GetServiceId() const { return service_id_; }
 
   /**
    * @brief Sets the service ID.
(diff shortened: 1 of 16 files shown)


olp-cpp-sdk-authentication/include/olp/authentication/AuthorizeRequest.h
```diff
@@ -122,9 +122,9 @@ class AUTHENTICATION_API AuthorizeRequest final {
    *
    * @return A reference to the updated `DecisionRequest` instance.
    */
-  inline AuthorizeRequest& WithContractId(
-      boost::optional<std::string> contract_id) {
-    contract_id_ = std::move(contract_id);
+  template <class T = porting::optional<std::string>>
+  AuthorizeRequest& WithContractId(T&& contract_id) {
```
andrey-kashcheev, line 126: Does it mean that now the signature is `WithContractId<>(...)`?

asopov-here: There were two signatures: with boost::optional<std::string> and std::string&&. boost::optional has the explicit constructor so, the call .WithContractId("123") was resolvable. std::optional has the implicit one so the resolution isn't possible. To avoid such issues I suggest to provide a single generic method valid in all cases.

andrey-kashcheev: The issue here is a breaking change of the public API

asopov-here: Could you please clarify? From my perspective it extends public API, not break it by any means. But direct replacement boost::optional by std::optional will break the compatibility with the existing code - it's a reason why I propose this change.

olp-cpp-sdk-authentication/include/olp/authentication/TokenResult.h
```diff
@@ -19,7 +19,7 @@
 
 #pragma once
 
-#include <boost/optional.hpp>
+#include <olp/core/porting/optional.hpp>
```
rustam-gamidov-here, line 22: looks like it should be in separate block like in the 'authentication/Settings.h' above not sure how strict it is though

rustam-gamidov-here (approved):

PR #1626, opened 2025-08-08
Use porting::optional in olp-cpp-sdk-write instead of boost::optional.

asopov-here:
So, boost::optional type is used for optional values. For C++17 and above more convenient is to use std::optional instead. The task NLAM-23 is about making this type configurable. This commit is a second part of the task: olp-cpp-sdk-write.

Relates-To: NLAM-23

### DIFF
--- a/olp-cpp-sdk-dataservice-write/src/generated/ConfigApi.h
+++ b/olp-cpp-sdk-dataservice-write/src/generated/ConfigApi.h
@@ -22,7 +22,7 @@
 #include <memory>
 #include <string>
 
-#include <boost/optional.hpp>
+#include <olp/core/porting/optional.hpp>
 
 #include <olp/core/client/ApiError.h>
 #include <olp/core/client/ApiResponse.h>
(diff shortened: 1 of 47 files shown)


andrey-kashcheev (approved):

olp-cpp-sdk-dataservice-write/src/generated/ConfigApi.h
```diff
@@ -51,14 +51,15 @@ class ConfigApi {
    * @param billing_tag An optional free-form tag which is used for grouping
    * billing records together. If supplied, it must be between 4 - 16
    * characters, contain only alpha/numeric ASCII characters  [A-Za-z0-9].
+   * @param callback
```
rustam-gamidov-here, line 54: this one looks suspicious, i would say `callback` should be added to the next line instead of adding the new one similar to the L52 above. Not sure maybe doxygen is happy with the additional line as well

asopov-here: Yes, it was added by formatter presumably, the original code did't refer to this parameter's name at all. Fixed for consistency.

rustam-gamidov-here (approved):

PR #1631, opened 2025-08-12
Make boost::optional as default regardless of C++ version.

asopov-here:
So, boost::optional type is used for optional values. For C++17 and above more convenient is to use std::optional instead. The task NLAM-23 is about making this type configurable. This commit makes the boost::optional default even for C++17 builds. The std::optional usage shall be switched on explicitly using OLP_SDK_USE_STD_OPTIONAL=ON flag.

Relates-To: NLAM-23


rustam-gamidov-here:
Proper PR name will make it easier to read log in future

mykolamalik (approved):

PR #1634, opened 2025-09-09
Allow overriding Java version

haraldF:
More recent Java compilers can't generate for target 1.7, so allow overriding the version via OLP_SDK_JAVA_TARGET.

Relates-To: MINOR


andrey-kashcheev (approved):
Please ignore format check results

PR #1644, opened 2025-11-13
Create olp::porting::any to use either boost::any or std::any

khomenkd:
Create olp::porting::any to use either boost::any or std::any for different compiler versions

Relates-To: NLAM-157

### DIFF
--- a/olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
+++ b/olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
@@ -18,6 +18,7 @@
  */
 
 #include <chrono>
+#include <fstream>
 #include <thread>
 
 #include <gtest/gtest.h>
@@ -31,6 +32,10 @@
 #define WIN32_LEAN_AND_MEAN
 #include <Windows.h>
 #undef max
+#else
+#include <dirent.h>
+#include <sys/stat.h>
+#include <sys/types.h>
 #endif
 
 #include "Helpers.h"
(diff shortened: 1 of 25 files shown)


olp-cpp-sdk-core/tests/cache/DefaultCacheImplTest.cpp
```diff
@@ -1447,4 +1449,53 @@ std::vector<OpenTestParameters> DefaultCacheImplOpenParams() {
 INSTANTIATE_TEST_SUITE_P(, DefaultCacheImplOpenTest,
                          testing::ValuesIn(DefaultCacheImplOpenParams()));
 
+TEST_F(DefaultCacheImplTest, ProtectedCacheIOErrorFallbackToReadOnly) {
+  SCOPED_TRACE("IOError fallback to read-only for protected cache");
+
+  // Create a writable directory with a subdirectory named "LOCK" to force
+  // LevelDB to fail when trying to create its lock file, triggering IOError.
+  // This exercises the fallback branch (lines 787-789) that retries opening
+  // in read-only mode after an IOError in write mode.
```
rustam-gamidov-here: the rest of the file contains pretty simple tests without numerous comments, maybe it make sense to stick with that stile. And it make sense to ensure this test is reaching the point you are interested in.
Not critical, just looks very different, feel free to keep as it is if you believe that's how it should be

rustam-gamidov-here: That is probably the reason of windows fails.
There was no need to make the test or change it, that was just a point of improvement

PR #1693, opened 2026-04-23
Add build with std::any and std::optional

rustam-gamidov-here:
boost::any and boost::optional used by default and it make sense to check that alternative options are not broken

Relates-To: MINOR


mykolamalik (approved):

rustam-gamidov-here:
assuming it's not needed, thus closing

andrey-kashcheev (approved):

PR #1701, opened 2026-06-12
Option to embed Boost JSON

andrey-kashcheev:
Option to control whether we should use header-only Boost JSON or as a separate library.
Turn on by default.

Relates-To: DATASDK-97


khomenkd (approved):

CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ option(OLP_SDK_NO_EXCEPTION "Disable exception handling" OFF)
 option(OLP_SDK_USE_STD_OPTIONAL "Use std::optional instead of boost::optional with C++17 and above" OFF)
 option(OLP_SDK_USE_STD_ANY "Use std::any instead of boost::any with C++17 and above" OFF)
 option(OLP_SDK_BOOST_THROW_EXCEPTION_EXTERNAL "The boost::throw_exception() is defined externally" OFF)
+option(OLP_SDK_EMBED_BOOST_JSON "Use boost.json in header-only mode." ON)
```
rustam-gamidov-here, line 43: picky one. this looks like a file, not sure is it called boost.json or boost::json
no big deal

rustam-gamidov-here (approved):

CMakeLists.txt
andrey-kashcheev: From the official documentation `Boost.JSON is a portable C++ library...` so with the single dot, but the letter case is different

PR #1707, opened 2026-07-03
Add line breaks to clang-format filter

rustam-gamidov-here:
This excludes cmake files. clang-format is not able to format cmake
scripts and makes them ugly.

These line breaks are not working on BSD grep.
Alternative \b is not perfect as well as it matches c1 and c_

Relates-To: MINOR


andrey-kashcheev (approved):
I wonder why would clang-format try to format cmake files?

rustam-gamidov-here:
script is asking git for all files except deleted
then grep is filtering files we want to format. And there's check `grep '\.c\|...` that matches `name.cmake` files so this PR is adding $ that is end of line to the pattern
P.S. clang-format is formatting any file it gets, at least version 6 is

khomenkd (approved):

PR #1708, opened 2026-07-03
Use XCode 26.1.1 in MacOS15 build

rustam-gamidov-here:
Fixing 'error: iOS 26.1 Platform Not Installed.'

Alternative: xcodebuild -downloadPlatform iOS
no delay in the job duration observed and no there's no this command in the log to check if actual download has happened or is it just a setup

Relates-To: MINOR


rustam-gamidov-here:
closing as `latest` is not macos 15 and changing image is better

PR #1710, opened 2026-07-09
Debug change to trigger CI on cmake file

rustam-gamidov-here:
Check clang-formatting

Relates-To: MINOR


rustam-gamidov-here:
used to check #1707 effectiveness  before and after merge thus closing

PR #1713, opened 2026-07-13
Keep thread name after attaching it to JVM

mzimonczyk-here:
When native thread is attached to JVM, then it's name is taken from JavaVMAttachArgs. When no JavaVMAttachArgs, or no JavaVMAttachArgs::name passed, then JVM on its own decides on naming thread. Those names are not descriptive. To preserve thread name, pass the currently set thread name in JavaVMAttachArgs::name.

### DIFF
--- a/olp-cpp-sdk-core/src/http/android/utils/JNIThreadBinder.h
+++ b/olp-cpp-sdk-core/src/http/android/utils/JNIThreadBinder.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 HERE Europe B.V.
+ * Copyright (C) 2019-2026 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


andrey-kashcheev (dismissed):

rustam-gamidov-here (dismissed):

olp-cpp-sdk-core/src/http/android/utils/JNIThreadBinder.h
rustam-gamidov-here: it make sense to update the date, not sure how critical that is

mzimonczyk-here: done

rustam-gamidov-here (approved):

mykolamalik (approved):

rustam-gamidov-here:
i assume @mzimonczyk-here is not able to merge as well as to trigger the pipelines, thus pressing the button for him

PR #1720, opened 2026-08-07
Add clang-tidy job/script

ystefinko:
Add GithubActions job and script which compiles and checks clang-tidy with minimal set of tidy configs

Relates-To: OCMAM-143


ystefinko:
The job/script was copied into PR https://github.com/heremaps/here-data-sdk-cpp/pull/1723 and merged

PR #1721, opened 2026-08-13
Allow to set CAINFO blob for HTTPS proxy

diachenko-mischa:
CAINFO for the HTTPS proxy must be passed
as a specific parameter into easy_handle
instead of general CAINFO

Relates-To: HERESUP-80353


olp-cpp-sdk-core/include/olp/core/http/NetworkProxySettings.h
```diff
@@ -124,6 +124,28 @@ class CORE_API NetworkProxySettings final {
    */
   NetworkProxySettings& WithPassword(std::string password);
 
+  /**
```
rustam-gamidov-here, line 127: please update copyrights header in changed files

andrey-kashcheev (dismissed):

rustam-gamidov-here (approved):

andrey-kashcheev (approved):

PR #1722, opened 2026-08-21
Add MtlsTokenProvider for mTLS authentication

rustam-gamidov-here:
Implementing auto refreshing in place not using AutoRefreshToken and TokenEndpoint for simplicity and because mtls endpoint has very simple list of arguments and some features implemented in client authentication may be not supported at the moment.

Relates-To: DATASDK-105

### DIFF
--- a/olp-cpp-sdk-authentication/src/AuthenticationClientImpl.cpp
+++ b/olp-cpp-sdk-authentication/src/AuthenticationClientImpl.cpp
@@ -38,8 +38,10 @@
 #include "SignInUserResultImpl.h"
 #include "SignOutResultImpl.h"
 #include "SignUpResultImpl.h"
+#include "olp/core/client/OlpClientSettingsFactory.h"
 #include "olp/core/http/Network.h"
 #include "olp/core/http/NetworkConstants.h"
+#include "olp/core/http/NetworkInitializationSettings.h"
 #include "olp/core/http/NetworkResponse.h"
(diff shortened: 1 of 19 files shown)


olp-cpp-sdk-authentication/src/AuthenticationClientImpl.cpp
```diff
@@ -395,6 +398,74 @@ client::CancellationToken AuthenticationClientImpl::SignInClient(
...
+
+      const auto status = auth_response.GetStatus();
+      if (status < 0) {
+        response = GetSignInResponse<SignInResult>(
+            auth_response, context, settings_.token_endpoint_url);
+
+        if (status == static_cast<int>(http::ErrorCode::TIMEOUT_ERROR) &&
+            context.IsCancelled()) {
+          return response;
+        }
```
andrey-kashcheev: I guess this one is redundant since GetSignInResponse is already doing it

rustam-gamidov-here: If a timeout occurred, the cancellation is done through the context and in case of retry we return `Cancelled` at the retries loop beginning. This return is making the difference.
Removed this additional handling to keep this code aligned with the `SignInClient` and changed tests accordingly

khomenkd (approved):

mykolamalik (approved):

PR #1723, opened 2026-08-21
fix: address clang-tidy 17 errors

isidorowiczdev:
CI enables clang-tidy with -warnings-as-errors, and three checks
were failing the build:

- clang-analyzer-cplusplus.NewDeleteLeaks (24+ sites across
  TokenEndpointImpl, IndexLayerClientImpl, VolatileLayerClientImpl,
  VersionedLayerClientImpl, StreamLayerClientImpl): Every flagged
  call path was traced end-to-end and contains zero raw new/delete;
  ownership is shared_ptr/std::function throughout, so each site
  is suppressed with NOLINTNEXTLINE

- clang-analyzer-deadcode.DeadStores (Crypto.cpp:171): ComputeSha256
  mutated `value` via `>>=` when packing the last output byte, a
  redundant shift whose result was never read; changed to
  non-mutating `>>`.

- clang-analyzer-optin.cplusplus.VirtualCall (SignInResultImpl.cpp:95):
  the constructor called virtual IsValid() instead of reading
  is_valid_ directly

Resolves: DATASDK-104

Signed-off-by: Iwo Sidorowicz <iwo.sidorowicz@here.com>
Co-Author: Yaroslav Stefinko <ext-yaroslav.stefinko@here.com>


scripts/misc/clang-tidy-17-check.sh
```diff
@@ -0,0 +1,24 @@
...
+    -DOLP_SDK_BUILD_EXAMPLES=OFF
+
+cmake --build "${BUILD_DIR}" -- -j"$(nproc)"
+
+run-clang-tidy-17.py -p "${BUILD_DIR}" \
+    -config-file="${PWD}/.clang-tidy" \
+    "${PWD}/olp-cpp-sdk-core/.*" \
+    "${PWD}/olp-cpp-sdk-authentication/.*" \
+    "${PWD}/olp-cpp-sdk-dataservice-read/.*" \
+    "${PWD}/olp-cpp-sdk-dataservice-write/.*"
```
rustam-gamidov-here, line 41: please update copyright blocks in changed files
maybe added scripts should have it as well, check other scripts in the reposotory

rustam-gamidov-here (approved):
No issues found, Except that i would simplify MR title a bit as 'fix' and 'address' looks pretty similar and to follow previous MRs style

ystefinko (approved):

PR #1724, opened 2026-08-28
Add a new endpoint in BlobApi to support objectstore.

asopov-here:
Handle /layers/{layer_id}/keys/{key} path.

Add tests

Relates-To: DATASDK-103

### DIFF
--- a/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
+++ b/olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2024 HERE Europe B.V.
+ * Copyright (C) 2019-2026 HERE Europe B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,6 +30,7 @@
 #include <olp/core/client/PendingRequests.h>
 #include <olp/core/porting/optional.h>
 #include <olp/dataservice/read/DataRequest.h>
+#include <olp/dataservice/read/KeyDataRequest.h>
 #include <olp/dataservice/read/PartitionsRequest.h>
 #include <olp/dataservice/read/PrefetchPartitionsRequest.h>
 #include <olp/dataservice/read/PrefetchTileResult.h>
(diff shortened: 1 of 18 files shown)


tests/integration/olp-cpp-sdk-dataservice-read/VersionedLayerClientTest.cpp
```diff
@@ -38,6 +38,7 @@
 #include <olp/core/porting/warning_disable.h>
 #include <olp/core/utils/Dir.h>
 #include <olp/core/utils/Url.h>
+#include <olp/dataservice/read/PrefetchTileResult.h>
```
rustam-gamidov-here: it looks unrelated to the change, same in several other files, header are changed without obvious reason

olp-cpp-sdk-dataservice-read/src/VersionedLayerClientImpl.h
```diff
@@ -66,6 +67,12 @@ class VersionedLayerClientImpl {
   virtual client::CancellableFuture<DataResponse> GetData(
       DataRequest data_request);
 
+  virtual client::CancellationToken GetDataByKey(KeyDataRequest request,
```
rustam-gamidov-here: Not sure about name, i see others are `GetData` on the other hand it may be easier to read when names are different. Better to doublecheck if there are other opinions

olp-cpp-sdk-dataservice-read/tests/VersionedLayerClientImplTest.cpp
```diff
@@ -110,6 +110,72 @@ TEST(VersionedLayerClientTest, GetData) {
...
+        .WillOnce(ReturnHttpResponse(olp::http::NetworkResponse().WithStatus(
+                                         olp::http::HttpStatusCode::OK),
+                                     "keyData"));
+
+    auto future = client.GetDataByKey(read::KeyDataRequest().WithKey("1/1/2"))
+                      .GetFuture();
+    const auto& response = future.get();
+    ASSERT_TRUE(response.IsSuccessful());
+    ASSERT_TRUE(response.GetResult() != nullptr);
+    ASSERT_EQ(7u, response.GetResult()->size());
```
rustam-gamidov-here: it may be more valuable to check for exact data if you know it
maybe there are more cases in PR

PR #1726, opened 2026-09-21
Add support for Retry-After header

andrey-kashcheev:
Optional additional waiting time between retry
attempts. Only version with seconds is supported

Relates-To: DATASDK-107


olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -182,6 +183,30 @@ bool CaseInsensitiveCompare(const std::string& str1, const std::string& str2) {
...
+    }
+
+    if (header.second.empty()) {
+      return std::chrono::seconds::zero();
+    }
+
+    const std::chrono::seconds::rep seconds_from_header =
+        std::strtol(header.second.c_str(), nullptr, 10);
+
+    if (seconds_from_header == LONG_MAX || seconds_from_header == LONG_MIN) {
```
rustam-gamidov-here: maybe `std::numeric_limits` ? may depends on the http header description of course

rustam-gamidov-here (dismissed):

olp-cpp-sdk-core/src/client/OlpClient.cpp
```diff
@@ -286,9 +311,10 @@ NetworkCallbackType GetRetryCallback(
 
     // TODO: Do not block thread but instead implement an event queue that will
     // trigger next retry once the time expired!
-    const auto actual_wait_time =
-        std::min(settings->current_backdown_period,
-                 settings->max_wait_time - settings->accumulated_wait_time);
+    auto wait_time = settings->current_backdown_period;
+    wait_time += GetRetryAfterWaitTime(response);
+    const auto actual_wait_time = std::min(
+        wait_time, settings->max_wait_time - settings->accumulated_wait_time);
```
mykolamalik, line 319: If the RetryAfterWaitTime is a big enough number, it may be ignored. Is it really what we want to achieve?
Same question to the line 875

olp-cpp-sdk-core/src/client/OlpClient.cpp
andrey-kashcheev: Limits check is for conversion, strtol may return min or max in case of errors, it could return 0 as well but in that case no checks needed

olp-cpp-sdk-core/src/client/OlpClient.cpp
andrey-kashcheev: Updated the description of `timeout` setting

rustam-gamidov-here (approved):

mykolamalik (approved):

